PR #11, opened 2019-05-31
Update readme.md

jeetiss:
npm install don't have flag `--save` from v5 and saves dependencies just with `install` command

```
list of flags: [--save-prod|--save-dev|--save-optional] [--save-exact] [--no-save]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Getting started guide and documentation you can find in [docs](./docs) folder.
 ### es6 via npm
 
 ```bash
-npm install --save lightweight-charts
+npm install lightweight-charts
 ```
 
 ```js


timocov:
Oh, I believe there should be a `--save-prod` option instead (to explicitly show that `lightweight-charts` is deps, not dev/optional/other deps).

README.md
```diff
@@ -22,7 +22,7 @@ Getting started guide and documentation you can find in [docs](./docs) folder.
 ### es6 via npm
 
 ```bash
-npm install --save lightweight-charts
+npm install lightweight-charts
```
timocov:
```suggestion
npm install --save-prod lightweight-charts
```

jeetiss: wtf, man
nobody writes this flag in readme 💁🏻‍♂️

PR #15, opened 2019-06-01
Spell a crosshair as "crosshair", not as "cross-hair" or "cross hair"

timocov:


timocov:
CI isn't run due maintenance - https://discuss.circleci.com/t/upcoming-scheduled-maintenance-1-june-2019-17-00-utc-21-00-utc/30565. Needs trigger it after maintenance is ended.

timocov:
I just added a test case for cross hair options, so that's why graphics-tests job is failed now.

PR #43, opened 2019-06-10
Fixed bug in setVisibleRange

### LINKED ISSUE #33: setVisibleRange works incorrectly
Please check this [jsfiddle](https://jsfiddle.net/TradingView/8v02wao7/1/) it should display the chart from the beginning of 2019 (actually, from Dec, 26), but it displays it from the middle of Jan'2019.

eugene-korobko:
Bugfix: fixed setVisibleRange function

Fixes #33

**Overview of change:**

Apply new setVisibleRange in RAF like fitContent


src/model/time-scale.ts
```diff
@@ -530,6 +530,25 @@ export class TimeScale {
...
+		}
+
+		const firstPoint = ensureNotNull(points.valueAt(firstIndex)).timestamp;
+		const lastPoint = ensureNotNull(points.valueAt(lastIndex)).timestamp;
+
+		const barRange = new BarsRange(
+			ensureNotNull(points.indexOf(Math.max(firstPoint, range.from.timestamp) as UTCTimestamp, true)),
+			ensureNotNull(points.indexOf(Math.min(lastPoint, range.to.timestamp) as UTCTimestamp, true))
+		);
+		this.setVisibleRange(barRange);
```
timocov: I would like to see graphics test case for that.

src/model/time-scale.ts
```diff
@@ -512,7 +512,7 @@ export class TimeScale {
 	}
 
 	public setVisibleRange(range: BarsRange): void {
-		const length = range.lastBar() - range.firstBar();
+		const length = range.lastBar() - range.firstBar() + 1;
```
timocov:
```suggestion
		const length = range.count();
```

src/model/time-scale.ts
eugene-korobko: I've added one

PR #44, opened 2019-06-10
color.ts: code cleanup

subzey:
**Type of PR:** enhancement

**PR checklist:**

- [ ] Addresses an existing issue: fixes #🐱
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Removed buncha unused and overcompilcated code.
Result is -125 B min.gz and much more restricted module exports.

Each change has its own commit, you probably find it easier to review commit-wise

### DIFF
--- a/src/helpers/color.ts
+++ b/src/helpers/color.ts
@@ -193,7 +193,7 @@ function normalizeNumber(min: number, n: number, max: number): number {
 	);
 }
 
-function normalizeColorComponent<T extends RedComponent | GreenComponent | BlueComponent>(component: number): T {
+function normalizeRgbComponent<T extends RedComponent | GreenComponent | BlueComponent>(component: number): T {
 	return normalizeInteger(0, component, 255) as T;
 }
 
@@ -211,9 +211,9 @@ namespace RgbShortHexRepresentation {
 	export const re = /^#([0-9a-fA-F])([0-9a-fA-F])([0-9a-fA-F])$/;
 	export function parse(matches: RegExpExecArray): Rgb {
 		return [
-			normalizeColorComponent<RedComponent>(parseInt(matches[1] + matches[1], 16)),
-			normalizeColorComponent<GreenComponent>(parseInt(matches[2] + matches[2], 16)),
-			normalizeColorComponent<BlueComponent>(parseInt(matches[3] + matches[3], 16)),
+			normalizeRgbComponent<RedComponent>(parseInt(matches[1] + matches[1], 16)),
+			normalizeRgbComponent<GreenComponent>(parseInt(matches[2] + matches[2], 16)),
+			normalizeRgbComponent<BlueComponent>(parseInt(matches[3] + matches[3], 16)),
 		];
 	}
 }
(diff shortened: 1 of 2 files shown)


src/helpers/color.ts
```diff
@@ -346,11 +330,12 @@ export function parseRgb(colorString: string): Rgb {
...
-const rgbGrayscaleWeights = [0.199, 0.687, 0.114];
-export function rgbToGrayscale(rgbValue: Rgb): number {
-	return rgbGrayscaleWeights[0] * rgbValue[0] +
-		rgbGrayscaleWeights[1] * rgbValue[1] +
-		rgbGrayscaleWeights[2] * rgbValue[2];
+function rgbToGrayscale(rgbValue: Rgb): number {
+	return (
+		0.199 * rgbValue[0] +
+		0.687 * rgbValue[1] +
+		0.114 * rgbValue[2]
```
timocov: I would like to have that constants (0.199, 0.687, 0.114) with their names. You can use `const enum RgbGrayscaleWeights {}` for example, in local constants (I believe they should be inline as well).

src/helpers/color.ts
```diff
@@ -197,19 +193,11 @@ function normalizeNumber(min: number, n: number, max: number): number {
 	);
 }
 
-export function normalizeRedComponent(red: number): RedComponent {
-	return normalizeInteger(0, red, 255) as RedComponent;
+function normalizeColorComponent<T extends RedComponent | GreenComponent | BlueComponent>(component: number): T {
```
Nipheris: `normalizeRgbComponent`? Different components in different color models are normalized to other ranges. Even alpha is normalized to [0, 100].

src/helpers/color.ts
subzey: @timocov Weights `0.199`, `0.687`, `0.114` are probably a typo.

https://en.wikipedia.org/wiki/YUV#Y.E2.80.B2UV444_to_RGB888_conversion mentions `0.299`, `0.587`, `0.114`

subzey: Fixed, incl coefficients

src/helpers/color.ts
subzey: Changed

src/helpers/color.ts
subzey: Reverted coefficients change.

src/helpers/color.ts
```diff
@@ -346,11 +330,17 @@ export function parseRgb(colorString: string): Rgb {
 	}
 }
 
-const rgbGrayscaleWeights = [0.199, 0.687, 0.114];
-export function rgbToGrayscale(rgbValue: Rgb): number {
-	return rgbGrayscaleWeights[0] * rgbValue[0] +
-		rgbGrayscaleWeights[1] * rgbValue[1] +
-		rgbGrayscaleWeights[2] * rgbValue[2];
+function rgbToGrayscale(rgbValue: Rgb): number {
+	// See https://en.wikipedia.org/wiki/YUV#Y.E2.80.B2UV444_to_RGB888_conversion
```
timocov: It looks like the values from this link aren't the same as below ones. In this case I would prefer to change this comment to something like:

```
// the following values are from https://en.wikipedia.org/wiki/YUV#Y.E2.80.B2UV444_to_RGB888_conversion
// but a little bit adjusted for real use cases
```

subzey: check

subzey:
Here comes the tests. Partially based on @Nipheris' code

PR #72, opened 2019-06-17
Fixed using transparent background

### LINKED ISSUE #40: backgroundColor: 'transparent' / using transparency
When setting the layout properties to transparent, the chart doesn't seem to be working correctly.
https://jsfiddle.net/70k4yuhL/2/

eugene-korobko:
Fixes #40

**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #40
- [x] Includes tests - graphic tests transparent-background.js
- [ ] Documentation update

**Overview of change:**
If background color has alpha component, clear canvas before paiting the background


tests/e2e/graphics/test-cases/initial-options/transparent-background.js
```diff
@@ -0,0 +1,322 @@
...
+	var chart = LightweightCharts.createChart(container, {
+		layout: {
+			backgroundColor: 'rgba(80, 80, 80, 0.5)',
+		},
+	});
+
+	var series = chart.addLineSeries();
+
+	series.setData([
+		{ time: '2018-03-28', value: 154 },
```
timocov: Can we use random data as other test cases?

tests/e2e/graphics/test-cases/initial-options/transparent-background.js
```diff
@@ -0,0 +1,322 @@
...
+		{ time: '2019-06-03', value: 96.800003 },
+		{ time: '2019-06-04', value: 96.34 },
+		{ time: '2019-06-05', value: 95.94 },
+	]);
+
+	chart.timeScale().setVisibleRange({ from: 1545782400, to: 1559692800 });
+
+	return new Promise((resolve) => {
+		setTimeout(() => {
+			chart.timeScale().setVisibleRange({ from: 1538352000, to: 1545782400 });
```
timocov: It's a copy-paste?

src/gui/canvas-utils.ts
```diff
@@ -27,6 +28,10 @@ export function resizeCanvas(canvas: HTMLCanvasElement, newSize: Size): void {
 export function clearRect(ctx: CanvasRenderingContext2D, x: number, y: number, w: number, h: number, clearColor: string): void {
 	ctx.save();
 	ctx.translate(-0.5, -0.5);
+	const color = tryParseRgbaString(clearColor);
```
timocov: Will not be here a performance issue to parse the color on every paint? Maybe we can cache the result somewhere in PaneWidget/PriceAxisWidget/TimeAxisWidget? @subzey what do you think?

timocov: Or even parse there, but here take an argument with something like `forceClear` (I don't know a good name for now for that flag).

tests/e2e/graphics/test-cases/initial-options/transparent-background.js
```diff
@@ -0,0 +1,322 @@
+// eslint-disable-next-line no-unused-vars
+function runTestCase(container) {
+	var chart = LightweightCharts.createChart(container, {
+		layout: {
+			backgroundColor: 'rgba(80, 80, 80, 0.5)',
```
timocov: Can you please change background color of the webpage as well to see it on the screenshots?

tests/e2e/graphics/test-cases/initial-options/transparent-background.js
eugene-korobko: No, different range. We need it because drawing artefacts appear after scrolling

tests/e2e/graphics/test-cases/initial-options/transparent-background.js
eugene-korobko: The idea of the test is to show that there is no drawing artefacts after scrolling

timocov: Yeah, I got it, but the reason of using "transparent background" is see body's (or under-chart element's) background, and it would be good to see, that these colors are mixed.

tests/e2e/graphics/test-cases/initial-options/transparent-background.js
timocov: Ah, you'r right!

src/gui/canvas-utils.ts
eugene-korobko: 1. Parsing is a single call of regexp, I believe it is very effective
2. Method name is "clearRect". It means it "clear rect and fill it with color". Internally, it did not clear becasue of it is not necessary if color has no alpha. However, it was just an internal optimization; calling code should know nothing about this, it just calls a function to clear rect and fill it with a color

eugene-korobko: 1. Parsing is a single call of regexp, I believe it is very effective
2. Method name is "clearRect". It means it "clear rect and fill it with color". Internally, it did not clear becasue of it is not necessary if color has no alpha. However, it was just an internal optimization; calling code should know nothing about this, it just calls a function to clear rect and fill it with a color

eugene-korobko: 1. Parsing is a single call of regexp, I believe it is very effective
2. Method name is "clearRect". It means it "clear rect and fill it with color". Internally, it did not clear becasue of it is not necessary if color has no alpha. However, it was just an internal optimization; calling code should know nothing about this, it just calls a function to clear rect and fill it with a color

timocov: What if color is named color `transparent` (see original issue https://github.com/tradingview/lightweight-charts/issues/40#issue-453651199)? Will it work too? Would be nice to have test case for it.

subzey: Regexps are not so fast. :man_shrugging:. I believe, we can benefit from caching the color.

Probably, it can be done at the `color.ts` level memoizing the color param. It can end up in creating a huge `_cache` dictionary, but as for me, it's quite improbable a lot of different colors will be used.

src/gui/canvas-utils.ts
```diff
@@ -27,6 +28,10 @@ export function resizeCanvas(canvas: HTMLCanvasElement, newSize: Size): void {
...
+	if (color !== null && color[3] < 1) {
+		ctx.clearRect(x, y, w, h);
```
subzey: Clearing the previous ctx contents and fill can be done as a single operation using [`globalCompositeOperation = 'copy'`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation). This mode could even be faster becuase it doesn't blend the previous data with the new one.

Demo: https://jsfiddle.net/subzey/cxej91f2/

subzey: The added benefit is that we don't need to check the alpha channel, so we can skip `tryParseRgbaString` entirely!

subzey: @timocov @eugene-korobko

eugene-korobko: Nice idea! Will change

tests/e2e/graphics/test-cases/initial-options/transparent-background.js
eugene-korobko: done

src/gui/canvas-utils.ts
eugene-korobko: it is not actual now after using 'copy' compositing

tests/e2e/graphics/test-cases/initial-options/transparent-background.js
eugene-korobko: Well, master already draws transparent backround, the problem was in clearing chart elements after chart changes. Test should show difference between "before" and "after", the differece is artefacts, not transparent background

src/gui/canvas-utils.ts
eugene-korobko: this code has been removed

src/gui/canvas-utils.ts
eugene-korobko: done

PR #75, opened 2019-06-18
Pass ohlc data to mouse event handlers

### LINKED ISSUE #56: subscribeCrosshairMove for Candleseries
Hi,

I'm unable to retrieve open, high, low, close in subscribeCrosshairMove param object with CandleSeries. Is there a way to achieve this ?
The only thing I manage to retrieve is a single value which I assume is the close (I think) : 
`const price = param.seriesPrices.get(this.oCandleSeries);`

Thanks,

eugene-korobko:
Fixes #56

**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #56
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**

Now handlers of mousse events (subscribeCrosshairMove/subscribeClick) receive OHLC data for seris of types Bar and Candlestick.


src/model/series.ts
```diff
@@ -325,6 +325,23 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
 		return this.data().search(index, options);
 	}
 
+	public dataAt(time: TimePointIndex): BarPrice | BarPrices | null {
```
timocov: I believe return type here can be depends on current series type (as options for example).

timocov:
Could it be tested for now? Or we need to add infra to doing something with the browser?

src/gui/chart-widget.ts
```diff
@@ -423,18 +424,16 @@ export class ChartWidget implements IDestroyable {
...
 		if (time !== null) {
 			const serieses = this._model.serieses();
 			serieses.forEach((s: Series) => {
 				// TODO: replace with search left
-				const prices = s.data().valueAt(time);
+				const prices = s.dataAt(time);
 				if (prices !== null) {
-					const price = s.barFunction()(prices.value);
-					seriesPrices.set(s, price);
+					seriesPrices.set(s, prices);
```
timocov: @kirchet can you please take a look? It looks like we'll create a new object on every move event? Is it ok for GC? @eugene-korobko

kirchet: The only thing that we can do here is to do these calculations on the API layer and only if there is at least one subscriber

eugene-korobko: We could use the same object and update it inplace and make API layer making deep copy for each subscribing

src/api/chart-api.ts
```diff
@@ -266,8 +267,8 @@ export class ChartApi implements IChartApi, DataUpdatesConsumer<SeriesType> {
 	}
 
 	private _convertMouseParams(param: MouseEventParamsImpl): MouseEventParams {
-		const seriesPrices = new Map<ISeriesApi<SeriesType>, number>();
-		param.seriesPrices.forEach((price: number, series: Series) => {
+		const seriesPrices = new Map<ISeriesApi<SeriesType>, BarPrice | BarPrices>();
```
timocov: We need to update docs as well.

docs/events.md
```diff
@@ -73,13 +73,13 @@ export type MouseEventHandler = (param: MouseEventParams) => void;
...
 - `point`: (`{ x: number, y: number }`, optional) - coordinate
-- `seriesPrices`: (`Map<ISeriesApi, number>`) - series prices
+- `seriesPrices`: (`Map<ISeriesApi, number | OHLC>`) - series prices
 
 `time` is not defined if an event was fired outside of data range (e.g. right/left of all data points).
 
 `point` is not defined if an event was fired outside of the chart (for example on a mouse leave event).
 
-`seriesPrices` contains all series prices according to the event point details.
+`seriesPrices` contains all series prices according to the event point details. Key is a series, value is a number for single-value series types (line, area, histogram) or OHLC structure for candlestick and bar series.
```
ezhukovskiy: `seriesPrices` is an object with prices of all series corresponding to the event point. The object keys are series APIs, values are prices. Each price value is a number for single-value series types (line, area, histogram) or OHLC structure for candlestick and bar series.

PR #97, opened 2019-07-03
Fixed smoothed screenshot

### LINKED ISSUE #31: Add API to take a screenshot

timocov:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #31
- [ ] Includes tests
- [ ] Documentation update

### DIFF
--- a/src/gui/chart-widget.ts
+++ b/src/gui/chart-widget.ts
@@ -213,7 +213,7 @@ export class ChartWidget implements IDestroyable {
 		const targetCanvas = document.createElement('canvas');
 		resizeCanvas(targetCanvas, new Size(targetWidth, targetHeight));
 		const ctx = ensureNotNull(getContext2d(targetCanvas));
-		// we need this atually because of resizeCanvas does translate, but we do not need it
+		// we need this actually because of resizeCanvas does translate, but we do not need it
 		ctx.translate(-0.5, -0.5);
 		let targetX = 0;
 		let targetY = 0;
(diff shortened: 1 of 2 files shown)


src/gui/chart-widget.ts
```diff
@@ -213,6 +213,7 @@ export class ChartWidget implements IDestroyable {
 		const targetCanvas = document.createElement('canvas');
 		resizeCanvas(targetCanvas, new Size(targetWidth, targetHeight));
 		const ctx = ensureNotNull(getContext2d(targetCanvas));
+		ctx.translate(0.5, 0.5);
```
timocov: It looks like this translation is already did in `resizeCanvas` function. Why do we need to have it twice?

timocov: @eugene-korobko GitHub treats me as PR's author and doesn't allow me request changes/approve/reject it. So, just come here please :)

timocov:
Also, maybe we can enable hidpi in tests to handle this?

src/gui/chart-widget.ts
eugene-korobko: Actually, we do not need translating at all, so we have to revert translate from `resizeCanvas`

src/gui/chart-widget.ts
```diff
@@ -213,6 +213,8 @@ export class ChartWidget implements IDestroyable {
 		const targetCanvas = document.createElement('canvas');
 		resizeCanvas(targetCanvas, new Size(targetWidth, targetHeight));
 		const ctx = ensureNotNull(getContext2d(targetCanvas));
+		// we need this atually because of resizeCanvas does translate, but we do not need it
```
timocov:
```suggestion
		// we need this actually because of resizeCanvas does translate, but we do not need it
```

PR #99, opened 2019-07-04
Graphics tests on different devicePixelRatio

timocov:
**Type of PR:** tests enhancement

**Overview of change:**

The PR introduce a way to run graphics tests on different `devicePixelRatio`s. By default they run with `devicePixelRatio=1`, but you can override it by passing `DEVICE_PIXEL_RATIO` environment variable.

On CI we run tests for 1, 1.5 and 2 `devicePixelRatio`s.

### DIFF
--- a/tests/e2e/graphics/graphics-test-cases.ts
+++ b/tests/e2e/graphics/graphics-test-cases.ts
@@ -6,15 +6,13 @@ import * as path from 'path';
 import { expect } from 'chai';
 import {
 	after,
-	afterEach,
-	Context,
 	describe,
 	it,
 } from 'mocha';
 import { PNG } from 'pngjs';
 
 import { compareScreenshots } from './helpers/compare-screenshots';
-import { getTestCases } from './helpers/get-test-cases';
+import { getTestCases, TestCase } from './helpers/get-test-cases';
 import { Screenshoter } from './helpers/screenshoter';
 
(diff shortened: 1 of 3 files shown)


tests/e2e/graphics/graphics-test-cases.ts
```diff
@@ -29,41 +27,56 @@ function generatePageContent(standaloneBundlePath: string, testCaseCode: string)
...
 const testStandalonePathEnvKey = 'TEST_STANDALONE_PATH';
 
+let devicePixelRatio = process.env.DEVICE_PIXEL_RATIO ? parseFloat(process.env.DEVICE_PIXEL_RATIO) : 1;
+if (isNaN(devicePixelRatio)) {
+	devicePixelRatio = 1;
+}
+
+const devicePixelRatioStr = devicePixelRatio.toFixed(2);
+
 const testResultsOutDir = path.resolve(process.env.CMP_OUT_DIR || path.join(__dirname, '.gendata'));
```
kirchet: Maybe we should have different output directories for each device pixel ratio

timocov: For each `devicePixelRatio` it creates separate folder inside `testResultsOutDir` - see https://github.com/tradingview/lightweight-charts/pull/99/files#diff-5e28aa8fc458ec37f073ea79fb37f80cR94

PR #102, opened 2019-07-05
Pinch isn't prevented by long tap

### LINKED ISSUE #95: Pinch isn't prevented by long tap

**Steps/code to reproduce:**

1. Open chart (tested on mobile Safari)
1. Long tap on chart to enable tracking mode (do not release your finger)
1. Touch with a second finger and try to move them to make a pinch zoom.

**Actual behavior:**

Both pinch zoom and tracking mode are active.

**Expected behavior:**

Only one action/gesture should be active at the moment (either pinch or long tap). Even if we just prevent pinch [here](https://github.com/tradingview/lightweight-charts/blob/deb0e0d8cc2143c26aa4f1957c4f984c7efb37ae/src/gui/mouse-event-handler.ts#L456-L461), we can release one of two fingers and then press it again. It looks like we should detect only one action while any finger is on the screen and reset them when no fingers on the screen.

kirchet:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #95 
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**

Added more checks on starting pinch event processing


timocov:
Approved, but wait for release 1.0.0.

PR #118, opened 2019-07-11
LineSeries is not displayed if starting x value is out of viewport

### LINKED ISSUE #116: LineSeries is not displayed if starting x value is out of viewport
Screenshot 2019-07-10 at 22 19 36

Screenshot 2019-07-10 at 22 22 09

You'll note the first line disappears as you zoom in and it goes just outside of the viewport, the second line will do the same when you zoom further.

Fiddle with full reference data: https://jsfiddle.net/jimeney/Lgtrhj52/9/

Steps to repeat: zoom out

kirchet:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #116 
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Changed series items range to draw for line-based series, fixed visible price range calculation

### DIFF
--- a/src/model/series.ts
+++ b/src/model/series.ts
@@ -397,11 +397,11 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
 				barsMinMax.min === barsMinMax.max ?
 					new PriceRange(barsMinMax.min - 0.5, barsMinMax.max + 0.5) : // special case: range consists of the only point
 					new PriceRange(barsMinMax.min, barsMinMax.max) :
-				new PriceRange(-0.5, 0.5);
+				null;
 
 		if (this.seriesType() === 'Histogram') {
 			const base = (this._options as HistogramStyleOptions).base;
-			range = range.merge(new PriceRange(base, base));
+			range = range !== null ? range.merge(new PriceRange(base, base)) : new PriceRange(base - 0.5, base + 0.5);
 		}
 
 		return range;
--- a/src/model/time-data.ts
+++ b/src/model/time-data.ts
@@ -49,8 +49,11 @@ function upperBoundItemsCompare(time: TimePointIndex, item: TimedValue): boolean
 }
 
 export function visibleTimedValues(items: TimedValue[], range: BarsRange, extendedRange: boolean): SeriesItemsIndexesRange {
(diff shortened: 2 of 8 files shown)


src/model/time-data.ts
```diff
@@ -48,8 +48,24 @@ function upperBoundItemsCompare(time: TimePointIndex, item: TimedValue): boolean
...
-	return { from, to };
+
+	if (!extendedRange) {
+		return { from, to };
+	}
+
+	let extendedFrom = from;
+	let extendedTo = to;
+
+	if (from > 0 && items[from].time >= range.firstBar()) {
```
timocov: Maybe we can save `range.firstBar()` and `range.lastBar()` into local variables?

tests/e2e/graphics/test-cases/series/2-points-line-series.js
```diff
@@ -0,0 +1,42 @@
...
+	var data = generateData();
+	var lineData = [
+		{ time: data[0].time, value: data[0].close },
+		{ time: data[data.length - 1].time, value: data[data.length - 1].close },
+	];
+
+	mainSeries.setData(data);
+	lineSeries.setData(lineData);
+
+	chart.timeScale().scrollToPosition(-5);
```
timocov: It would be good to have the same test but with scrolling in opposite side. @eugene-korobko what do you think?

kirchet: For what reason? In this test case both line series points are outside of the viewport

timocov: > In this test case both line series points are outside of the viewport

Ah, ok then.

src/model/time-data.ts
kirchet: Done

src/model/series.ts
```diff
@@ -397,11 +397,11 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
 				barsMinMax.min === barsMinMax.max ?
 					new PriceRange(barsMinMax.min - 0.5, barsMinMax.max + 0.5) : // special case: range consists of the only point
 					new PriceRange(barsMinMax.min, barsMinMax.max) :
-				new PriceRange(-0.5, 0.5);
+				null;
```
eugene-korobko: Is this change related to the original task?

PR #121, opened 2019-07-11
Update crosshair when updating app pane views in ChartModel

### LINKED ISSUE #120: Crosshair isn't updated when timescale is changed
**Lightweight Charts Version:** 1.0.0

**Steps/code to reproduce:**

```javascript
function generateData() {
	var res = [];
	var time = new Date(Date.UTC(2018, 0, 1, 0, 0, 0, 0));
	for (var i = 0; i < 1000; ++i) {
		res.push({
			time: time.getTime() / 1000,
			value: i,
		});

		time.setUTCDate(time.getUTCDate() + 1);
	}
	return res;
}

var chart = LightweightCharts.createChart(document.getElementById('container'));

var mainSeries = chart.addLineSeries();

mainSeries.setData(generateData());

setTimeout(() => {
	chart.timeScale().setVisibleRange({
		from: (new Date(Date.UTC(2018, 0, 1, 0, 0, 0, 0))).getTime() / 1000,
		to: (new Date(Date.UTC(2018, 1, 1, 0, 0, 0, 0))).getTime() / 1000,
	});
}, 2000);
```

1. Open the page
1. Move cursor to some place
1. Wait until timeout is fired and visible range is changed

**Actual behavior:**

Crosshair views has old state (old date/time and price).

**Expected behavior:**

Crosshair views has updated state.

**Screenshots**

Actual:

image

Expected: the date, time and price should be from the current view.

timocov:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #120
- [ ] Includes tests (tests are already in the repo)
- [ ] Documentation update

### DIFF
--- a/src/model/chart-model.ts
+++ b/src/model/chart-model.ts
@@ -158,6 +158,7 @@ export class ChartModel implements IDestroyable {
 
 	public updateAllPaneViews(): void {
 		this._panes.forEach((p: Pane) => p.updateAllViews());
+		this.updateCrosshair();
 	}
 
 	public timeScale(): TimeScale {


src/model/chart-model.ts
```diff
@@ -158,6 +158,7 @@ export class ChartModel implements IDestroyable {
 
 	public updateAllPaneViews(): void {
 		this._panes.forEach((p: Pane) => p.updateAllViews());
+		this.updateCrosshair();
```
kirchet: I think we should add a corresponding test for this

timocov: There is a lot of tests cases already 🙂 As you can see, the code in the repro in the issue is pretty similar to `set-visible-range-after-time` test case. For now in `master` we have incorrect screenshots.

Also, some tests are failed)

PR #123, opened 2019-07-12
Don't extend range by 0.5 in Series (it's already done in PriceScale)

### LINKED ISSUE #122: Price Axis Scaling Bug
It appears there's an intermittent, but repeatable issue with price axis scaling.

Renders correctly
https://www.chartex.pro/chartviewerstatic-debug.html?symbol=BINANCE:ALGO/USDT&trendlines=1&supports=1&timeframe=60&w=1600&h=800&barstoshow=500

Screenshot 2019-07-12 at 00 33 03

Axis scaling fails
https://www.chartex.pro/chartviewerstatic-debug.html?symbol=OANDA:EUR/GBP&trendlines=1&supports=1&timeframe=60&w=1600&h=800&barstoshow=500

timocov:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #122
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

We don't need to extend price range by 0.5 in series because 1) it affects result price range where are other series 2) 0.5 value can be huge for current series (for example if values are about 1.0) 3) the same thing are already done in PriceScale and this is the only good place for that.

### DIFF
--- a/src/model/price-scale.ts
+++ b/src/model/price-scale.ts
@@ -857,7 +857,7 @@ export class PriceScale {
 			}
 		}
 
-		if (priceRange) {
+		if (priceRange !== null) {
 			// keep current range is new is empty
 			if (priceRange.minValue() === priceRange.maxValue()) {
 				priceRange = new PriceRange(priceRange.minValue() - 0.5, priceRange.maxValue() + 0.5);
--- a/src/model/price-scale.ts
+++ b/src/model/price-scale.ts
@@ -732,7 +732,7 @@ export class PriceScale {
 		const mainSource = this.mainSource();
 		let base = 100;
 		if (mainSource !== null) {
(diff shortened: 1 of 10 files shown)


src/model/price-scale.ts
```diff
@@ -857,7 +857,7 @@ export class PriceScale {
 			}
 		}
 
-		if (priceRange) {
+		if (priceRange !== null) {
 			// keep current range is new is empty
 			if (priceRange.minValue() === priceRange.maxValue()) {
 				priceRange = new PriceRange(priceRange.minValue() - 0.5, priceRange.maxValue() + 0.5);
```
timocov: After discussing with @eugene-korobko we decide that here should be 5 min ticks, not just 0.5.

PR #142, opened 2019-07-22
Reimplemented the canvas creation and scaling logic using fancy-canvas library

### LINKED ISSUE #141: Migrate canvas-related logic to fancy-canvas library
Current logic of canvas creation and configuration with regard to devicePixelRatio can be improved using [`fancy-canvas`](https://github.com/tradingview/fancy-canvas/) library.

We should create:
- visible canvases - bound to current devicePixelRatio (see `bindToDevicePixelRatio` function);
- offscreen canvases - using explicit coordinate space scaling.

Nipheris:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #141 
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**
- all the canvases are created through `canvas-utils` by using the explicit static scaling (for offscreen canvas elements) OR dynamic resizing with binding to devicePixelRatio (for visible canvas elements);
- devicePixelRatio changes and related automatic canvas resize are both handled by `fancy-canvas` library;
- drawImage calls now include size in arguments;
- rendering contexts are no more stored in private fields, but are requested just before performing the drawing operations.

Browser API usage, performance.


src/gui/canvas-utils.ts
```diff
@@ -14,14 +16,29 @@ export class Size {
...
 
-export function resizeCanvas(canvas: HTMLCanvasElement, newSize: Size): void {
-	canvas.width = newSize.w;
-	canvas.height = newSize.h;
+export function getPrescaledContext2D(canvas: HTMLCanvasElement): CanvasRenderingContext2D | null {
+	const pixelRatio =
+		canvas.ownerDocument &&
+		canvas.ownerDocument.defaultView &&
+		canvas.ownerDocument.defaultView.devicePixelRatio
+		|| 1;
```
timocov: Might be we can extract that logic into new function and use it in `createPreconfiguredCanvas` function?

src/gui/pane-separator.ts
```diff
@@ -80,14 +79,16 @@ export class PaneSeparator implements IDestroyable {
 		return this._rowElement;
 	}
 
+	public getSize(): Readonly<Size> {
```
timocov: Is it used anywhere outside of this class? If don't, I would make it private instead.

src/gui/pane-widget.ts
```diff
@@ -432,11 +437,11 @@ export class PaneWidget implements IDestroyable {
...
 
 		this._size = size;
-		resizeCanvas(this._canvas, size);
-		resizeCanvas(this._topCanvas, size);
 
-		// This line is here for retina canvas shim to work
-		this._ctx = ensureNotNull(getContext2d(this._canvas));
+		// apply new size to canvases
+		this._canvasBinding.canvasSize = { width: size.w, height: size.h };
+		this._topCanvasBinding.canvasSize = { width: size.w, height: size.h };
```
timocov: Is it plainly that set a property will resize the canvas? Maybe we can add methods instead of getter/setter?

src/gui/pane-widget.ts
```diff
@@ -706,4 +713,7 @@ export class PaneWidget implements IDestroyable {
 	private _model(): ChartModel {
 		return this._chart.model();
 	}
+
+	private readonly _canvasConfiguredHandler = () => this.paint(InvalidationLevel.Full);
+	private readonly _topCanvasConfiguredHandler = () => this.paint(InvalidationLevel.Cursor);
```
timocov: Why we do here force paint instead of requesting invalidation?

src/gui/price-axis-stub.ts
```diff
@@ -92,49 +98,48 @@ export class PriceAxisStub implements IDestroyable {
...
+		ctx.restore();
 	}
 
-	private _drawBackground(): void {
-		clearRect(this._ctx, 0, 0, this._size.w, this._size.h, this._options.layout.backgroundColor);
+	private _drawBackground(ctx: CanvasRenderingContext2D): void {
+		clearRect(ctx, 0, 0, this._size.w, this._size.h, this._options.layout.backgroundColor);
 	}
+
+	private readonly _canvasConfiguredHandler = () => this.paint(InvalidationLevel.Full);
```
timocov: The same about invalidation.

src/gui/time-axis-widget.ts
```diff
@@ -181,14 +186,17 @@ export class TimeAxisWidget implements MouseEventHandlers, IDestroyable {
...
+
 	public setSizes(timeAxisSize: Size, stubWidth: number): void {
 		if (!this._size || !this._size.equals(timeAxisSize)) {
 			this._size = timeAxisSize;
-			resizeCanvas(this._canvas, timeAxisSize);
-			resizeCanvas(this._topCanvas, timeAxisSize);
 
-			getContext2d(this._canvas); // Sync HiDpiCanvas metrics
-			getContext2d(this._topCanvas); // Sync HiDpiCanvas metrics
+			// apply new size to canvases
```
timocov: That's why I would use methods instead. In almost every place where we apply new canvas size via setting the property we have such comment :)

src/gui/canvas-utils.ts
```diff
@@ -14,14 +16,29 @@ export class Size {
...
+	const ctx = canvas.getContext('2d');
+	if (ctx) {
+		// scale by pixel ratio
+		ctx.setTransform(pixelRatio, 0, 0, pixelRatio, 0, 0);
+	}
+	return ctx;
+}
 
-	const ctx = getContext2d(canvas);
+export function getPretransformedContext2D(binding: CanvasCoordinateSpaceBinding): CanvasRenderingContext2D | null {
```
timocov: What do you think about return ensured value in this function instead of checking that value in every consumer?

src/gui/canvas-utils.ts
```diff
@@ -1,4 +1,6 @@
-import { getContext2d } from '../helpers/canvas-wrapper';
+import { Binding as CanvasCoordinateSpaceBinding, bindToDevicePixelRatio } from 'fancy-canvas/coordinate-space';
```
timocov: What do you think about make `fancy-canvas/coordinate-space` for now as the default module and make re-exports?

timocov:
Also, afaik, I would be good to add new test case with candles with max zoom out.

/cc @vfisenko

src/gui/pane-separator.ts
Nipheris: https://github.com/tradingview/lightweight-charts/pull/142/files#diff-f84d64383e454ee3a64791a19691f324R245

timocov: It's a little bit strange that we still export `SEPARATOR_HEIGHT` and even use it, but ok...

src/gui/pane-widget.ts
Nipheris: 7136ec73d5e55f68b827503e706a450d1ca585a4

src/gui/price-axis-stub.ts
Nipheris: 7136ec73d5e55f68b827503e706a450d1ca585a4

src/gui/time-axis-widget.ts
Nipheris: f697411d4eeb00df40d201047e06d44db3361ce6

src/gui/pane-widget.ts
Nipheris: f697411d4eeb00df40d201047e06d44db3361ce6

PR #154, opened 2019-07-30
Series-based markers

### LINKED ISSUE #24: Series-based markers
Hi! Thanks for the great work on this library.
I couldn't find in the docs any info on whether it's possible to plot a series of potentially sparse points that would be shown with some icon for each point on the chart, or maybe something like the possibility to add annotations.
The use case I'm after is being able to show on the chart specific points in time/price where something happened, such as real world events that might have affected the price, etc.
I'm not yet very familiar with the codebase of the library, but it seems to me that the simplest way to implement something like this would be to support a kind of "scatter series" that would allow you to specify an icon (image or char?) to plot each point with, which could coexist with a normal candlestick series for example, and could show a popover with details about that specific point (potentially arbitrary text) when hovered by the user.
Any way to do something like this now? If not, would it eventually be added?
Thanks again!

### LINKED ISSUE #108: Lines and Annotations
I'm trying to plot simple horizontal lines, referencing data stored externally, and via a loop.

As there are no documented drawing tools, so I'm opting to add datapoints to a line series, on demand.

Sample data:
`{xe: 1562713200, xs: 1561834800, ye: 12835.97409692, ys: 12244}`

The code is designed to add one or more line series dependent on the data returned, although in the examples with static code, I can see that the chart line does not cope with missing data for the timeframe resulting in a solid horizontal line, when the data should start and stop in the middle of the chart.

This is the code in question:
` function loadTrendLines() {

                 
                    const url = "/api/trendlines/" + symbol.replace("/", "_") + "/" + timeframe;

                    $.ajax({
                        type: 'GET',
                        url: url,
                        contentType: 'application/json',
                        dataType: 'json',
                        responseType: 'application/json',

                        success: function (jdata)
                        {
                            console.log(jdata);

                            var i;
                            for (i = 0; i < jdata.TrendLines.length; ++i) {
                                console.log(jdata.TrendLines[i]);

                                var tldata = [];

                                tldata.push({ time: jdata.TrendLines[i].xe, value: jdata.TrendLines[i].ye });
                                tldata.push({ time: jdata.TrendLines[i].xs, value: jdata.TrendLines[i].ue });

                                var lineSeries = chart.addLineSeries({
                                    lastValueVisible: false
                                });

                                lineSeries.setData(tldata);

                            }

                        }

                    });
                }`

I'm receiving the following error:

Uncaught Error: Value is undefined
    at h (lightweight-charts.standalone.production.js:7)
    at p (lightweight-charts.standalone.production.js:7)
    at t.updateSeriesData (lightweight-charts.standalone.production.js:7)
    at t.updateData (lightweight-charts.standalone.production.js:7)
    at e.t.update (lightweight-charts.standalone.production.js:7)
    at Object.success (HTMLPage1.html?symbol=BINANCE:BTC/USDT&barstoshow=100&timeframe=60&trendlines=1:252)
    at c (jquery-3.4.1.min.js:2)
    at Object.fireWith [as resolveWith] (jquery-3.4.1.min.js:2)
    at l (jquery-3.4.1.min.js:2)
    at XMLHttpRequest.<anonymous> (jquery-3.4.1.min.js:2)

If I manually define the xe, xs, ye and ue values, running the same inner loop logic, but placing this outside of the loop, it is successful, however this is not referencing any live data, merely static examples.

Can you please advise on a) whether this workaround should be viable, and b) if there are drawing tools that can be leveraged?

kirchet:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #24
- [x] Includes tests
- [x] Documentation update

**Overview of change:**

Implemented displaying series markers.


src/model/chart-model.ts
```diff
@@ -38,6 +38,16 @@ export interface HandleScaleOptions {
 	axisPressedMouseMove: boolean;
 }
 
+export interface HoveredObject {
+	id?: string;
+	externalId?: string;
```
timocov: What's difference between `id` and `externalId`? Where each of them is used?

src/model/autoscale-info.ts
```diff
@@ -0,0 +1,23 @@
...
+		this.marginBelow = marginBelow;
+		this.marginAbove = marginAbove;
+	}
+
+	public merge(info: AutoscaleInfo): AutoscaleInfo {
+		const newRange = (this.priceRange === null) ? info.priceRange : this.priceRange.merge(info.priceRange);
+		return new AutoscaleInfo(
+			newRange,
+			(this.marginBelow + info.marginBelow) as Coordinate,
+			(this.marginAbove + info.marginAbove) as Coordinate
```
timocov: Why we add them instead of finding max value?

src/model/price-scale.ts
```diff
@@ -854,9 +861,19 @@ export class PriceScale {
 				} else {
 					priceRange = priceRange.merge(ensureNotNull(sourceRange));
 				}
+
+				marginAbove = Math.max(marginAbove, autoScaleInfo.marginAbove);
+				marginBelow = Math.max(marginAbove, autoScaleInfo.marginBelow);
```
timocov: Can we use `merge` from `AutoscaleInfo` here?

src/model/series-markers.ts
```diff
@@ -0,0 +1,15 @@
...
+export interface SeriesMarkerParams {
+}
+
+export interface SeriesMarker<TimeType> {
+	time: TimeType;
+	position: SeriesMarkerPosition;
+	shape: SeriesMarkerShape;
+	color: string;
+	id: string;
+	externalId?: string;
```
timocov: Is it public field (public for the user)? What is stands for? What's the diff between id and externalId?

src/model/autoscale-info.ts
```diff
@@ -0,0 +1,23 @@
+import { Coordinate } from './coordinate';
+import { PriceRange } from './price-range';
+
+export class AutoscaleInfo {
```
timocov: Do we need to have whole class here (with all public fields without any check in ctor) instead of "structure"/object and export merge function from here?

src/views/pane/series-markers-pane-view.ts
```diff
@@ -0,0 +1,160 @@
...
+
+	public renderer(height: number, width: number, addAnchors?: boolean): IPaneRenderer {
+		if (this._invalidated) {
+			this._makeValid();
+		}
+		this._renderer.setData(this._data);
+		return this._renderer;
+	}
+
+	public autoScaleMargins(): number {
```
timocov: Why "margins" and return only one number? 🤔

src/views/pane/series-markers-pane-view.ts
```diff
@@ -0,0 +1,160 @@
...
+import { IUpdatablePaneView, UpdateType } from './iupdatable-pane-view';
+
+interface Offsets {
+	aboveBar: number;
+	belowBar: number;
+}
+
+function calcuateY(
+		// tslint:disable-next-line:max-params
+		marker: SeriesMarker<TimePointIndex>,
```
timocov: it looks like here is one tab more than needed?

src/renderers/series-markers-utils.ts
```diff
@@ -0,0 +1,46 @@
...
+
+	ensureNever(shape);
+}
+
+export function ceilToEven(x: number): number {
+	const ceiled = Math.ceil(x);
+	return (ceiled % 2 !== 0) ? ceiled - 1 : ceiled;
+}
+
+export function ceilToOdd(x: number): number {
```
timocov: Maybe we can add unit tests for helpers from this file?

timocov:
Also there is conflict, and the PR is for #24 issue, not 108 (I don't know why somebody choose 108 in the name of the branch and put this number in the commits - I believe we can rebase them after the review to cleanup commit messages).

src/renderers/ipane-renderer.ts
```diff
@@ -1,7 +1,8 @@
+import { HoveredObject } from '../model/chart-model';
 import { Coordinate } from '../model/coordinate';
 
 export interface IPaneRenderer {
-	draw(ctx: CanvasRenderingContext2D, isHovered: boolean): void;
-	drawBackground?(ctx: CanvasRenderingContext2D, isHovered: boolean): void;
-	hitTest?(x: Coordinate, y: Coordinate): boolean;
+	draw(ctx: CanvasRenderingContext2D, isHovered: boolean, objectId?: string): void;
```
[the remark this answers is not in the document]
kirchet: For example to highlight some part of what we're going to draw.

src/model/series-markers.ts
kirchet: In our case `externalId` is related to a marker's id. `id` is the internal unique value to determine what part of a rendered object is hovered. A series marker might not have id, but internal id is always existing.

src/model/autoscale-info.ts
kirchet: Because each source is rendered independently of other sources

kirchet: Anyway, this method was removed

src/model/price-scale.ts
kirchet: Since `AutoscaleInfo` became to be an interface we cannot call `merge`

src/renderers/series-markers-square.ts
```diff
@@ -0,0 +1,34 @@
+import { Coordinate } from '../model/coordinate';
+
+import { shapeSize } from './series-markers-utils';
+
+export function drawSquare(
+	ctx: CanvasRenderingContext2D,
+	centerX: Coordinate,
+	centerY: Coordinate,
+	color: string,
+	barSpacing: number
```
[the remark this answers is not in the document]
kirchet: It's not very convinient to calculate and store all the required values to render a specific shape on a pane view update stage. Instead of it regardles of a shape type we're calculating only its center point. We can store the current `barSpacing` on the same stage and somehow rename it, but I have no idea how we should call this variable.

src/model/chart-model.ts
kirchet: In our case `externalId` is related to a marker's id. `id` is the internal unique value to determine what part of a rendered object is hovered. A series marker might not have id, but internal id is always existing.

src/api/ichart-api.ts
```diff
@@ -21,6 +21,8 @@ export interface MouseEventParams {
 	time?: UTCTimestamp | BusinessDay;
 	point?: Point;
 	seriesPrices: Map<ISeriesApi<SeriesType>, BarPrice | BarPrices>;
+	hoveredSeries?: ISeriesApi<SeriesType>;
+	hoveredObject?: string;
```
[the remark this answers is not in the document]
kirchet: Fixed, renamed to `hoveredMarkerId`

src/model/series-markers.ts
```diff
@@ -0,0 +1,15 @@
+export type SeriesMarkerPosition = 'aboveBar' | 'belowBar' | 'inBar';
+
+export type SeriesMarkerShape = 'circle' | 'square' | 'arrowUp' | 'arrowDown';
+
+export interface SeriesMarkerParams {
```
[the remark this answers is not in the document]
kirchet: Removed

src/renderers/series-markers-arrow.ts
```diff
@@ -0,0 +1,51 @@
...
+	color: string,
+	barSpacing: number
+): void {
+	ctx.fillStyle = color;
+	ctx.beginPath();
+	const arrowSize = size(barSpacing, 1);
+	const halfArrowSize = (arrowSize - 1) / 2;
+	const baseSize = size(barSpacing, 0.5);
+	const halfBaseSize = (baseSize - 1) / 2;
+	ctx.beginPath();
```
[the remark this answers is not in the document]
kirchet: Fixed

src/model/autoscale-info.ts
kirchet: Replaced with interface

src/views/pane/series-markers-pane-view.ts
kirchet: Repaced with `AutoScaleMargins`

src/views/pane/series-markers-pane-view.ts
kirchet: Fixed

src/renderers/series-markers-utils.ts
kirchet: I don't think that we need tests for these functions because they are primitive and we can change some constants within it in the future

src/model/series-markers.ts
timocov: Then I believe we need to or mark it as internal (maybe rename as well?) or create another type which will extend this one (called, for example, `InternalSeriesMarker` or something). Also, in this case, we need to either make a copy of every marker or write our fields in-place with some "unique" fields so we won't affect external user.

tests/e2e/graphics/test-cases/series-markers/series-markers-update.js
```diff
@@ -0,0 +1,36 @@
...
+	var markers = [
+		{ time: data[data.length - 10].time, position: 'belowBar', color: 'red', shape: 'arrowUp' },
+		{ time: data[data.length - 20].time, position: 'belowBar', color: 'red', shape: 'arrowUp' },
+		{ time: data[data.length - 30].time, position: 'belowBar', color: 'red', shape: 'arrowUp' },
+		{ time: data[data.length - 40].time, position: 'belowBar', color: 'red', shape: 'arrowUp' },
+	];
+
+	mainSeries.setMarkers(markers);
+
+	markers.push({ time: data[data.length - 1].time, position: 'belowBar', color: 'red', shape: 'arrowUp' });
```
timocov: Did we copy of the array passed to the `setMarkers`? 🤔

tests/e2e/graphics/test-cases/series-markers/series-markers-re-aligned.js
```diff
@@ -0,0 +1,34 @@
...
+
+	line.setMarkers([
+		{ time: '2017-04-10', position: 'inBar', color: 'orange', shape: 'circle' },
+		{ time: '2017-04-16', position: 'inBar', color: 'orange', shape: 'circle' },
+		{ time: '2017-04-17', position: 'inBar', color: 'orange', shape: 'circle' },
+		{ time: '2017-04-18', position: 'inBar', color: 'orange', shape: 'circle' },
+		{ time: '2017-04-24', position: 'inBar', color: 'orange', shape: 'circle' },
+	]);
+
+	line.update({ time: '2017-04-10', value: 85.01 });
```
timocov: Is it okay that we update the previous bars here? (the last bar is 2017-04-24, but we update 2017-04-10 here).

src/renderers/ipane-renderer.ts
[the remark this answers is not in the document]
timocov: Is it used anywhere for now?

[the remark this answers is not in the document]
kirchet: No, it's not used

src/model/series-markers.ts
kirchet: Added `InternalSeriesMarker` interface, added copying passed array in `setMarkers`

tests/e2e/graphics/test-cases/series-markers/series-markers-update.js
kirchet: Fixed, now we copy the array

tests/e2e/graphics/test-cases/series-markers/series-markers-re-aligned.js
kirchet: Yes, it's added intentionally, here we check that all markers are re-aligned even those that are at the middle of the chart bars

timocov: Maybe it would be better to call `setData` instead? We don't have any remark that update method might work for the past data and if we'll use it in that way we cannot change this behavior without breaking change.

PR #167, opened 2019-08-13
The chart cannot start from the left

### LINKED ISSUE #144: The chart can't start from the left
Hi, lightweight-charts team
I am facing an issue with lightweight-charts. I want to display my chart from left edge, I used
`fixLeftEdge: true`,  but it won't make the chart start on the left.

kirchet:
**Type of PR:** 

**PR checklist:**

- [x] Addresses an existing issue: fixes #144 
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Fixed applying the fixLeftEdge option

### DIFF
--- a/src/model/time-scale.ts
+++ b/src/model/time-scale.ts
@@ -178,7 +178,6 @@ export class TimeScale {
 				// reduce  _rightOffset means move right
 				// we could move more than required - this will be fixed by _correctOffset()
 				this._rightOffset -= Math.round(delta / this._barSpacing) + 1;
-				this._visibleBarsInvalidated = true;
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


src/model/time-scale.ts
```diff
@@ -735,6 +725,23 @@ export class TimeScale {
 			this._dateTimeFormatter = new DateFormatter(dateFormat, this._localizationOptions.locale);
 		}
 	}
+
+	private _fixLeftEdge(): void {
```
timocov: What if we change the data after `_fixLeftEdge` was called, will it work as well without calling it again? I would like to see test case for that.

timocov:
Also it looks like other test cases are failed for now. Is it ok? In some cases visible price range is changed, I'm not sure that it's ok for this PR...

src/model/time-scale.ts
kirchet: Done

PR #178, opened 2019-08-19
move to typecov

krzkaczor:
**Type of PR:** codechecks upgrade

**Overview of change:**
`type-coverage-watcher` was renamed simply to `typecov`. It has various small improvements which you might like — especially now output is more consistent.

Idea: I saw some visual regression tests in your repo. You might be interested in: https://github.com/codechecks/vis-reg I will work on adding support for storybook soon.

### DIFF
--- a/codechecks.yml
+++ b/codechecks.yml
@@ -7,7 +7,7 @@ checks:
         - path: "./dist/lightweight-charts.standalone.production.js"
           maxSize: 41.5KB
 
-  - name: type-coverage-watcher
+  - name: typecov
     options:
       atLeast: 99.9
       ignoreCatch: true
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
   "devDependencies": {
     "@codechecks/build-size-watcher": "~0.1.0",
     "@codechecks/client": "~0.1.5",
-    "@codechecks/type-coverage-watcher": "~0.1.3",
     "@types/chai": "~4.2.0",
     "@types/mocha": "~5.2.7",
     "chai": "~4.2.0",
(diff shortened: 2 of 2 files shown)


timocov:
Thank you!

timocov:
> Idea: I saw some visual regression tests in your repo. You might be interested in: https://github.com/codechecks/vis-reg I will work on adding support for storybook soon.

Looks interesting... I believe not only our screenshot tests might use codechecks, but dts-changes also.

timocov:
Hm, we need to be careful with that tests. They are [a little bit unstable](https://github.com/tradingview/lightweight-charts/blob/c10ea820623d181da3845e071fed97711870cd47/tests/e2e/graphics/graphics-test-cases.ts#L76) due some reasons and I'm afraid that sometime we can store "bad" screenshots for the master and it produced failed tests for other PRs. Also right now we can run them against any bundled code (any version, any place) and it's useful when you start something like "bisect".

But [dts-changes](https://github.com/tradingview/lightweight-charts/blob/c10ea820623d181da3845e071fed97711870cd47/.circleci/config.yml#L137-L144) still is a good candidate to check it by codechecks 👍

krzkaczor:
@timocov Hmm yes, unstable tests can be a problem. Usually, the way we approach it is that we just slowly, one by one try to eliminate these problems as we see it.

By dts changes you mean changes in typings ie public API, is that right? I thought about this for a long time - a codecheck warning about breaking changes in public API would be AMAZING.

timocov:
> Usually, the way we approach it is that we just slowly, one by one try to eliminate these problems as we see it.

We've tried, but unlucky. It's complicated e2e tests with puppeteer/chromium... Sometimes it just don't move cursor properly, sometimes something else (we tried to use setTimeout for that cases 🙂).

> By dts changes you mean changes in typings ie public API, is that right? 

Yes, that's correct.

> a codecheck warning about breaking changes in public API would be AMAZING.

That's exactly why we want to control our public API and don't add to them something unpredictable and this is why this test exists.

We can thought about make this test common and control typing in common case, but due there is no only one solution to generate bundle of typings (see https://github.com/Microsoft/TypeScript/issues/4433) we need to ensure that user can use its plugin/tool to generate dts (we use `dts-bundle-generator` for that).

Issue #270, opened 2019-11-27
Refactor DataLayer/ChartApi


timocov:
It's really unclear that series is cleared in `DataLayer`, the palette filled in `DataLayer` (`_setNewPoints`) but the data set to series in `ChartApi`. I suppose to refactor it in future releases. There is several key points:

- If we clear timescale, should we clear all series as well (see `ChartApi::updateTimeScale`) ? I guess yes.
- We don't need to modify or clear existent series' data (including palette) in `DataLayer` at all, because `ChartApi` is the guy who sends updates to chart's parts (series, timescale, etc) and the only it should do it.
- In addition to previous point - I think `DataLayer` should prepare a bit more complex timescaleUpdate without modifying anything outside of its own data/props.

_Originally posted by @timocov in https://github.com/tradingview/lightweight-charts/pull/269#discussion_r351228960_

PR #330, opened 2020-02-13
Fixed bug with wrong displaying several non regular series

### LINKED ISSUE #321: Bug while rendering few datasets with not equal timescale

**First case**

I have there 3 datasets 21 timepoints each. They are not equal in time: one datasets starts from point `{"value":0,"time":"2019-11-26"}` while others starts from `{"value":0,"time":"2019-04-14"}`. But after chart rendering you can see that this datasets (orange line) starts not from 26 Nov '19 but from 10 Sep '19.

image

jsfiddle with first case: https://jsfiddle.net/pensioneer/98p0v57a/1/

**Second case**

I have modified my fiddle to be loop-based. With this approach, I expected to get more flexible configuration. I got the same unexpected behavior, but with another result. After rendering I could see that the orange line starts from 4 Apr '19. Neither from 10 Sep '19 (first case) nor from 26 Nov '19 (expected result).

image

The datasets and configuration options are the same as in the first case. Only the initialization method was changed.

jsfiddle with second case: https://jsfiddle.net/pensioneer/ac1Le3bh/9/

**Tip**
If run my jsfiddles with Lightweight Charts Version 1.0.0 (https://unpkg.com/lightweight-charts@1.0.0/dist/lightweight-charts.standalone.production.js) orange line renders as expected (from 26 Nov '19)

kirchet:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #321
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Fixed bug with wrong displaying several non-regular series

### DIFF
--- a/src/api/chart-api.ts
+++ b/src/api/chart-api.ts
@@ -236,7 +236,7 @@ export class ChartApi implements IChartApi, DataUpdatesConsumer<SeriesType> {
 		const timeScaleUpdate = update.timeScaleUpdate;
 		model.updateTimeScale(timeScaleUpdate.index, timeScaleUpdate.changes, timeScaleUpdate.marks, true);
 		timeScaleUpdate.seriesUpdates.forEach((value: SeriesUpdatePacket, key: Series) => {
-			key.updateData(value.update);
+			key.updateData(value.update, true);
 		});
 		model.updateTimeScaleBaseIndex(0 as TimePointIndex);
 	}
--- a/src/api/chart-api.ts
+++ b/src/api/chart-api.ts
@@ -236,6 +236,9 @@ export class ChartApi implements IChartApi, DataUpdatesConsumer<SeriesType> {
 		const timeScaleUpdate = update.timeScaleUpdate;
 		model.updateTimeScale(timeScaleUpdate.index, timeScaleUpdate.changes, timeScaleUpdate.marks, true);
 		timeScaleUpdate.seriesUpdates.forEach((value: SeriesUpdatePacket, key: Series) => {
+			// the latest arg `true` must be removed in https://github.com/tradingview/lightweight-charts/issues/270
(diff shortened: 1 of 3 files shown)


src/api/chart-api.ts
```diff
@@ -236,7 +236,7 @@ export class ChartApi implements IChartApi, DataUpdatesConsumer<SeriesType> {
 		const timeScaleUpdate = update.timeScaleUpdate;
 		model.updateTimeScale(timeScaleUpdate.index, timeScaleUpdate.changes, timeScaleUpdate.marks, true);
 		timeScaleUpdate.seriesUpdates.forEach((value: SeriesUpdatePacket, key: Series) => {
-			key.updateData(value.update);
+			key.updateData(value.update, true);
```
timocov: Why can't we call `clearData` method here instead of passing `true`? Also, looks like we don't need to call `clearData` for the series in [`DataLayer::setSeriesData`](https://github.com/tradingview/lightweight-charts/blob/master/src/api/data-layer.ts#L261)?

kirchet: Because `clearData` clears also pallete which was filled right before in `DataLayer`

timocov: Oh crap 💩... Looks like we should fix it in https://github.com/tradingview/lightweight-charts/issues/270#issue-529280808 😂

timocov: What do you think about something like that?

```suggestion
			// the latest arg `true` must be removed in https://github.com/tradingview/lightweight-charts/issues/270
			// here we don't need to clear palettes because they were just filled in DataLayer
			// see https://github.com/tradingview/lightweight-charts/pull/330#discussion_r379415805
			key.updateData(value.update, true);
```

kirchet: Looks good

Issue #229, opened 2019-10-10
have the time scale automatically scaled to 100% width


krtschmr:
regardless if i have only 10 datapoints or 150, there should be an option that renders all datapoints into the visible chart area.

currently we need to hope that we have correct amount of data points for the last 24 hours and then play with timeScale.barSpacing to make it fit to the whole width
![image](https://user-images.githubusercontent.com/13366714/66548794-261de700-eb6c-11e9-90f8-5d150432f2f3.png)
![image](https://user-images.githubusercontent.com/13366714/66548826-3209a900-eb6c-11e9-95ac-b0f0433d3b18.png)

it's the same data for both graphs, just different bar spacing. would be awesome to simply ignore whatever settings and just render all datapoints into the full width.

eugene-korobko:
Hello!

Please check `fitContent` function

Example is [here](https://jsfiddle.net/TradingView/4pd5j8xa/)

Documentation is [here](https://github.com/tradingview/lightweight-charts/blob/master/docs/time-scale.md)

krtschmr:
@eugene-korobko 

it seems to be slightly broken. with `fitContent` i alwayts have these nasty borders left/right which are not really borders, the charts just don't draw there.

![image](https://user-images.githubusercontent.com/13366714/66561018-ced74100-eb82-11e9-8f3b-19825008defe.png)

krtschmr:
```
  const canvas = $("#charts .chart."+coin)[0]
  var chartOptions = { height: 175, width: canvas.offsetWidth }

  var chart = LightweightCharts.createChart(canvas, chartOptions);
  const areaSeries = chart.addAreaSeries({
    topColor: 'rgba(67, 141, 255, 0.05)',
    bottomColor: '#e8edf8',
    lineColor: '#438dff',
    lineStyle: 0,
    lineWidth: 2,
    crosshairMarkerVisible: true,
    crosshairMarkerRadius: 10,
    overlay: false,
    priceLineVisible: false
  });

  chart.applyOptions({
      handleScroll: { mouseWheel: false, pressedMouseMove: false, horzTouchDrag: false, vertTouchDrag: false },
      handleScale: { axisPressedMouseMove: false, mouseWheel: false, pinch: false, },
      timeScale: { barSpacing: 1, visible     : false },
      priceScale: { position: 'none', borderVisible: false,scaleMargins: { top: 0.2, bottom: 0.05,}},
[...]

PR #345, opened 2020-02-21
Timescale fixes

timocov:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: partially addressed to #229, but not at all
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

1. Increased max available bar spacing from 50 to 100
2. Removed +-1 from visible range in `fitContent`


timocov:
Actually this PR makes possible to zoom in the chart so part of bars couldn't be visible and scrolled:

image

Will be fixed in #352.

PR #349, opened 2020-02-22
Added text and size to series' markers

### LINKED ISSUE #207: Hard to see markers
The new markers feature is working really nice, however the markers are too small and hard to see. Would be nice if we could add a label on top/bottom of the arrow to make it easier to view. Or perhaps increase the size of the markers.

{ time: ..., label: 'bought 100 shares'...}

image

scrwdrv:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #207 
- [x] Includes tests
- [x] Documentation update

**Overview of change:**
Allow text to be added above/below the markers.
screenshot

### DIFF
--- a/docs/series-basics.md
+++ b/docs/series-basics.md
@@ -313,6 +313,7 @@ An array of items is expected. Each item should contain the following fields:
 - `time` ([Time](./time.md)) - item time
 - `position` (`aboveBar` &#124; `belowBar` &#124; `inBar`) - item position
 - `shape` (`circle` &#124; `square` &#124; `arrowUp` &#124; `arrowDown`) - item marker type
+- `size` (`number` &#124; `undefined`) - size multiplier of marker, the shape is hidden when set to `0`
 - `color` (`string`) - item color
 - `id` (`string` &#124; `undefined`) - item id, will be passed to click/crosshair move handlers
 - `text` (`string` &#124; `undefined`) - item text to be shown
--- a/docs/series-basics.md
+++ b/docs/series-basics.md
@@ -313,7 +313,7 @@ An array of items is expected. Each item should contain the following fields:
 - `time` ([Time](./time.md)) - item time
 - `position` (`aboveBar` &#124; `belowBar` &#124; `inBar`) - item position
 - `shape` (`circle` &#124; `square` &#124; `arrowUp` &#124; `arrowDown`) - item marker type
-- `size` (`number` &#124; `undefined`) - size multiplier of marker, the shape is hidden when set to `0`
+- `size` (`number` &#124; `undefined`) - size multiplier of the marker, the shape is hidden when set to `0`, default value is `1`
 - `color` (`string`) - item color
 - `id` (`string` &#124; `undefined`) - item id, will be passed to click/crosshair move handlers
 - `text` (`string` &#124; `undefined`) - item text to be shown
@@ -342,6 +342,7 @@ series.setMarkers([
         shape: 'arrowUp',
         id: 'id4',
         text: 'example',
+        size: 2,
     },
 ]);
 
--- a/docs/series-basics.md
+++ b/docs/series-basics.md
@@ -340,7 +340,7 @@ series.setMarkers([
         color: 'orange',
         shape: 'arrowUp',
         id: 'id4',
-        text: 'example'
+        text: 'example',
     },
 ]);
 
--- a/docs/series-basics.md
+++ b/docs/series-basics.md
@@ -315,6 +315,7 @@ An array of items is expected. Each item should contain the following fields:
 - `shape` (`circle` &#124; `square` &#124; `arrowUp` &#124; `arrowDown`) - item marker type
 - `color` (`string`) - item color
 - `id` (`string` &#124; `undefined`) - item id, will be passed to click/crosshair move handlers
+- `text` (`string` &#124; `undefined`) - item text to be shown
 
 Example:
 
@@ -339,6 +340,7 @@ series.setMarkers([
         color: 'orange',
         shape: 'arrowUp',
         id: 'id4',
+        text: 'example'
     },
 ]);
 
--- a/src/renderers/series-markers-arrow.ts
+++ b/src/renderers/series-markers-arrow.ts
@@ -7,6 +7,7 @@ import { hitTestSquare } from './series-markers-square';
 import { shapeSize, textPosition } from './series-markers-utils';
 
 export function drawArrow(
+	// tslint:disable-next-line:max-params
 	up: boolean,
 	ctx: CanvasRenderingContext2D,
 	centerX: Coordinate,
--- a/src/renderers/series-markers-arrow.ts
+++ b/src/renderers/series-markers-arrow.ts
@@ -12,6 +12,7 @@ export function drawArrow(
 	centerY: Coordinate,
 	color: string,
 	size: number,
+	belowBar: boolean,
 	text?: string
 ): void {
 	const arrowSize = shapeSize('arrowUp', size);
@@ -28,7 +29,6 @@ export function drawArrow(
 		ctx.lineTo(centerX + halfBaseSize, centerY + halfArrowSize);
 		ctx.lineTo(centerX - halfBaseSize, centerY + halfArrowSize);
 		ctx.lineTo(centerX - halfBaseSize, centerY);
-		if (text) ctx.fillText(text, centerX + arrowSize, centerY + halfArrowSize);
 	} else {
 		ctx.moveTo(centerX - halfArrowSize, centerY);
 		ctx.lineTo(centerX, centerY + halfArrowSize);
@@ -37,8 +37,17 @@ export function drawArrow(
 		ctx.lineTo(centerX + halfBaseSize, centerY - halfArrowSize);
 		ctx.lineTo(centerX - halfBaseSize, centerY - halfArrowSize);
 		ctx.lineTo(centerX - halfBaseSize, centerY);
-		if (text) ctx.fillText(text, centerX + arrowSize, centerY);
 	}
+
+	if (text) {
+		const textWidth = ctx.measureText(text).width;
+		const textHeight = parseInt(ctx.font, 10);
+		const textLeft = centerX - (textWidth / 2);
+		const textMargin = textHeight / 2;
+		const textTop = centerY - arrowSize / 2 + (belowBar ? 1 : -1) * (arrowSize + textMargin);
+		ctx.fillText(text, textLeft, textTop + textHeight);
+	}
+
 	ctx.fill();
 }
 
--- a/src/renderers/series-markers-arrow.ts
+++ b/src/renderers/series-markers-arrow.ts
@@ -1,26 +1,22 @@
 import { ceiledOdd } from '../helpers/mathex';
 
 import { Coordinate } from '../model/coordinate';
-import { SeriesMarkerText } from '../model/series-markers';
 
 import { hitTestSquare } from './series-markers-square';
 import { shapeSize } from './series-markers-utils';
 
 export function drawArrow(
-	// tslint:disable-next-line:max-params
 	up: boolean,
 	ctx: CanvasRenderingContext2D,
 	centerX: Coordinate,
 	centerY: Coordinate,
-	color: string,
-	size: number,
-	text?: SeriesMarkerText
+	size: number
 ): void {
 	const arrowSize = shapeSize('arrowUp', size);
 	const halfArrowSize = (arrowSize - 1) / 2;
 	const baseSize = ceiledOdd(size / 2);
 	const halfBaseSize = (baseSize - 1) / 2;
-	ctx.fillStyle = color;
+
 	ctx.beginPath();
 	if (up) {
 		ctx.moveTo(centerX - halfArrowSize, centerY);
@@ -38,10 +34,6 @@ export function drawArrow(
 		ctx.lineTo(centerX + halfBaseSize, centerY - halfArrowSize);
 		ctx.lineTo(centerX - halfBaseSize, centerY - halfArrowSize);
 		ctx.lineTo(centerX - halfBaseSize, centerY);
-	}
-
-	if (text !== undefined) {
-		ctx.fillText(text.content, text.x, text.y);
 	}
 
 	ctx.fill();
--- a/src/renderers/series-markers-arrow.ts
+++ b/src/renderers/series-markers-arrow.ts
@@ -11,7 +11,8 @@ export function drawArrow(
 	centerX: Coordinate,
 	centerY: Coordinate,
 	color: string,
-	size: number
+	size: number,
+	text?: string
 ): void {
 	const arrowSize = shapeSize('arrowUp', size);
 	const halfArrowSize = (arrowSize - 1) / 2;
@@ -27,6 +28,7 @@ export function drawArrow(
 		ctx.lineTo(centerX + halfBaseSize, centerY + halfArrowSize);
 		ctx.lineTo(centerX - halfBaseSize, centerY + halfArrowSize);
 		ctx.lineTo(centerX - halfBaseSize, centerY);
+		if (text) ctx.fillText(text, centerX + arrowSize, centerY + halfArrowSize);
 	} else {
 		ctx.moveTo(centerX - halfArrowSize, centerY);
 		ctx.lineTo(centerX, centerY + halfArrowSize);
@@ -35,6 +37,7 @@ export function drawArrow(
 		ctx.lineTo(centerX + halfBaseSize, centerY - halfArrowSize);
 		ctx.lineTo(centerX - halfBaseSize, centerY - halfArrowSize);
 		ctx.lineTo(centerX - halfBaseSize, centerY);
+		if (text) ctx.fillText(text, centerX + arrowSize, centerY);
 	}
 	ctx.fill();
 }
--- a/src/renderers/series-markers-arrow.ts
+++ b/src/renderers/series-markers-arrow.ts
@@ -4,7 +4,7 @@ import { Coordinate } from '../model/coordinate';
 import { SeriesMarkerText } from '../model/series-markers';
 
 import { hitTestSquare } from './series-markers-square';
-import { shapeSize, textPosition } from './series-markers-utils';
+import { shapeSize } from './series-markers-utils';
 
 export function drawArrow(
 	// tslint:disable-next-line:max-params
@@ -41,7 +41,7 @@ export function drawArrow(
 	}
 
 	if (text !== undefined) {
-		ctx.fillText(text.content, ...textPosition(centerX, centerY, text, arrowSize));
+		ctx.fillText(text.content, text.x, text.y);
 	}
 
 	ctx.fill();
--- a/src/renderers/series-markers-arrow.ts
+++ b/src/renderers/series-markers-arrow.ts
@@ -1,9 +1,10 @@
 import { ceiledOdd } from '../helpers/mathex';
 
 import { Coordinate } from '../model/coordinate';
+import { SeriesMarkerText } from '../model/series-markers';
 
 import { hitTestSquare } from './series-markers-square';
-import { shapeSize } from './series-markers-utils';
+import { shapeSize, textPosition } from './series-markers-utils';
 
 export function drawArrow(
 	up: boolean,
@@ -12,8 +13,7 @@ export function drawArrow(
 	centerY: Coordinate,
 	color: string,
 	size: number,
-	belowBar: boolean,
-	text?: string
+	text?: SeriesMarkerText
 ): void {
 	const arrowSize = shapeSize('arrowUp', size);
 	const halfArrowSize = (arrowSize - 1) / 2;
@@ -39,13 +39,8 @@ export function drawArrow(
 		ctx.lineTo(centerX - halfBaseSize, centerY);
 	}
 
-	if (text) {
-		const textWidth = ctx.measureText(text).width;
-		const textHeight = parseInt(ctx.font, 10);
-		const textLeft = centerX - (textWidth / 2);
-		const textMargin = textHeight / 2;
-		const textTop = centerY - arrowSize / 2 + (belowBar ? 1 : -1) * (arrowSize + textMargin);
-		ctx.fillText(text, textLeft, textTop + textHeight);
+	if (text !== undefined) {
+		ctx.fillText(text.content, ...textPosition(centerX, centerY, text, arrowSize));
 	}
 
 	ctx.fill();
--- a/src/renderers/series-markers-circle.ts
+++ b/src/renderers/series-markers-circle.ts
@@ -8,6 +8,7 @@ export function drawCircle(
 	centerY: Coordinate,
 	color: string,
 	size: number,
+	belowBar: boolean,
 	text?: string
 ): void {
 	const circleSize = shapeSize('circle', size);
@@ -15,7 +16,16 @@ export function drawCircle(
 	ctx.fillStyle = color;
 	ctx.beginPath();
 	ctx.arc(centerX, centerY, halfSize, 0, 2 * Math.PI, false);
-	if (text) ctx.fillText(text, centerX + circleSize, centerY + halfSize);
+
+	if (text) {
+		const textWidth = ctx.measureText(text).width;
+		const textHeight = parseInt(ctx.font, 10);
+		const textLeft = centerX - (textWidth / 2);
+		const textMargin = textHeight / 2;
+		const textTop = centerY - circleSize / 2 + (belowBar ? 1 : -1) * (circleSize + textMargin);
+		ctx.fillText(text, textLeft, textTop + textHeight);
+	}
+
 	ctx.fill();
 }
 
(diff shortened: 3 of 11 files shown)


src/renderers/series-markers-square.ts
```diff
@@ -7,13 +7,26 @@ export function drawSquare(
...
 ): void {
 	const squareSize = shapeSize('square', size);
 	const halfSize = (squareSize - 1) / 2;
 	const left = centerX - halfSize;
 	const top = centerY - halfSize;
+
 	ctx.fillStyle = color;
+
+	if (text) {
+		const textWidth = ctx.measureText(text).width;
```
timocov: I don't think that call `CanvasRenderingContext2D.measureText` on every paint is good idea. It might be perf bottleneck. We need to use `TextWidthCache` or something like that and reduce calls of `measureText` as much as possible.

cc @eugene-korobko

src/renderers/series-markers-square.ts
```diff
@@ -7,13 +7,26 @@ export function drawSquare(
...
+	text?: string
 ): void {
 	const squareSize = shapeSize('square', size);
 	const halfSize = (squareSize - 1) / 2;
 	const left = centerX - halfSize;
 	const top = centerY - halfSize;
+
 	ctx.fillStyle = color;
+
+	if (text) {
```
timocov:
```suggestion
	if (text !== undefined && text.length !== 0) {
```

src/renderers/series-markers-square.ts
```diff
@@ -7,13 +7,26 @@ export function drawSquare(
 	centerX: Coordinate,
 	centerY: Coordinate,
 	color: string,
-	size: number
+	size: number,
+	belowBar: boolean,
```
timocov: I don't think that name `belowBar` is valuable here - it's renderer, it doesn't even know what the bar is. How it's used?

src/renderers/series-markers-renderer.ts
```diff
@@ -63,16 +65,16 @@ export class SeriesMarkersRenderer extends ScaledRenderer {
...
-			drawArrow(true, ctx, item.x, item.y, item.color, item.size);
+			drawArrow(true, ctx, item.x, item.y, item.color, item.size, item.position === 'belowBar', item.text);
 			return;
 		case 'circle':
-			drawCircle(ctx, item.x, item.y, item.color, item.size);
+			drawCircle(ctx, item.x, item.y, item.color, item.size, item.position === 'belowBar', item.text);
 			return;
 		case 'square':
-			drawSquare(ctx, item.x, item.y, item.color, item.size);
+			drawSquare(ctx, item.x, item.y, item.color, item.size, item.position === 'belowBar', item.text);
```
timocov: @kirchet can you please take a look here? Is it possible we need to pass something like `textPosition` here?

src/renderers/series-markers-renderer.ts
```diff
@@ -88,7 +90,7 @@ function hitTestItem(item: SeriesMarkerRendererDataItem, x: Coordinate, y: Coord
 		case 'circle':
 			return hitTestCircle(item.x, item.y, item.size, x, y);
```
timocov: Why hit test doesn't have `item.position === 'belowBar'` argument? I think we would like to know that we've clicked/hovered on some item's text?

src/renderers/series-markers-square.ts
timocov: The same comments for other marker's code (in other files)

docs/series-basics.md
```diff
@@ -339,6 +340,7 @@ series.setMarkers([
         color: 'orange',
         shape: 'arrowUp',
         id: 'id4',
+        text: 'example'
```
timocov:
```suggestion
        text: 'example',
```

scrwdrv:
After submitted I soon read the `CONTRIBUTING.md` and found out it needs to be approved to make a PR, sorry about that, I hope this doesn't inconvenience you guys too much.

src/renderers/series-markers-square.ts
scrwdrv: It's only for determining the text should be above or below the marker shape. Maybe it will be better practice to calculate the offset value (text width as well) beforehand instead of passing this boolean value into renderer?

src/renderers/series-markers-arrow.ts
```diff
@@ -11,7 +11,9 @@ export function drawArrow(
 	centerX: Coordinate,
 	centerY: Coordinate,
 	color: string,
-	size: number
+	size: number,
+	belowBar: boolean,
```
eugene-korobko: I believe that boolean is a poor solution here, lets create a type like `SeriesMarkerTextPosition`

src/renderers/series-markers-circle.ts
```diff
@@ -7,13 +7,25 @@ export function drawCircle(
...
+	belowBar: boolean,
+	text?: string
 ): void {
 	const circleSize = shapeSize('circle', size);
 	const halfSize = (circleSize - 1) / 2;
 	ctx.fillStyle = color;
 	ctx.beginPath();
 	ctx.arc(centerX, centerY, halfSize, 0, 2 * Math.PI, false);
+
+	if (text) {
```
eugene-korobko: It is just a copy-paste? Lets move it to the base class or helper function.

src/renderers/series-markers-circle.ts
```diff
@@ -7,13 +7,25 @@ export function drawCircle(
...
+		const textWidth = ctx.measureText(text).width;
```
eugene-korobko: `measureText ` is a heavy operation, it is inefficient to measure text on each draw draw iteration since it does not change actually.  We have `TextWidthCache` class to optimize this, I believe we should use it here, you could refer its usage in the code

scrwdrv:
@timocov @eugene-korobko Thanks for the review, I didn't know there is a existed class for caching text width, and I took a look at the code of [text-width-cache.ts](https://github.com/tradingview/lightweight-charts/blob/1b96693de077610d7c907a5e4941eba51430f039/src/model/text-width-cache.ts) but not quite understand this:
```js
const defaultReplacementRe = /[2-9]/g;
...
const re = optimizationReplacementRe || defaultReplacementRe;
const cacheString = String(text).replace(re, '0');
```

Does this mean number `2-9` has the same width with number `0` even for non-monospaced fonts?

But again, `optimizationReplacementRe` changed to `1-9` at [time-axis-view-renderer.ts](https://github.com/tradingview/lightweight-charts/blob/3d5388a43fa2113fef0558dda6c0d475d8b2f097/src/renderers/time-axis-view-renderer.ts).

I'm a little bit confusing about what regex should be used for the marker's text cache, any suggestion?

docs/series-basics.md
HighBreedTech: Hello can we make use of this with the old v1.2 or how do i get the new version downloaded?
i wanted this option for long.

timocov:
Just remembered that markers could be stacked if they are on the same bar and we need to solve it somehow (one more test case 🙂).

docs/series-basics.md
timocov: @HighBreedTech use what? This is PR for the new feature which is even don't merged to the master yet! I understand your desire to get all features just now, but be patient and don't spam issues/PRs please.

> i wanted this option for long.

If you've wanted this feature for long, why you didn't make a PR with that feature thought? I've thought this is how open-source works, isn't? 🙂

timocov:
> Does this mean number 2-9 has the same width with number 0 even for non-monospaced fonts?

@scrwdrv Yes, I hope so. It's kind of optimization to avoid cache mismatches by having different numbers but with the same count of digits in it.

> optimizationReplacementRe changed to 1-9 at time-axis-view-renderer.ts.

I have nothing to say about that 😂. @eugene-korobko maybe you have any thoughts here?)

> I'm a little bit confusing about what regex should be used for the marker's text cache, any suggestion?

I guess in your case where the text might be any custom text (even not numbers) it doesn't really matter and you can use default one. The main purpose here why we should use TextWidthCache is caching and don't measure the same value twice.

src/renderers/series-markers-square.ts
timocov: > Maybe it will be better practice to calculate the offset value (text width as well) beforehand instead of passing this boolean value into renderer?

Yeah, maybe. Is it possible to move rendering text outside of that functions? Looks a bit similar. Also, please we have to think about https://github.com/tradingview/lightweight-charts/pull/349#issuecomment-591281530, because a solution might affect the result code.

src/views/pane/series-markers-pane-view.ts
```diff
@@ -44,6 +44,15 @@ function fillSizeAndY(
 	const lowPrice = isNumber(seriesData) ? seriesData : seriesData.low;
 	const shapeSize = calculateShapeHeight(timeScale.barSpacing());
 	rendererItem.size = shapeSize as Coordinate;
+
+	if (marker.text !== undefined && marker.text.length > 0) {
+		rendererItem.text = {
+			content: marker.text,
+			offsetX: 0,
+			offsetY: marker.position === 'aboveBar' ? -0.5 : 0.5,
```
timocov: @eugene-korobko please take a look here

scrwdrv: I just can't figure out an elegant way to pre calculate the position of text before it goes to renderer, since the actual shape size value is calculated by the renderer itself if I'm not mistaken.

So before it goes to renderer, I don't know its actual size, after goes to renderer, I don't know it's supposed to be above or below.

scrwdrv: > I just can't figure out an elegant way to calculate the position of text before it goes to renderer, since the actual shape size value is calculated by the renderer itself if I'm not mistaken.

Ah... never mind, I think I found how to solve this, it's just me thinking wrong.

scrwdrv:
Using absolute position (coordinate) seems making more sense for text positioning, and it's a more uniform style with `rendererItem`.

And yet the stacking problem is still WIP, major problem is how to get font height before it goes to the renderer, is there any interface that I can access  `ctx.font` property or something like that without making a mess?

image

scrwdrv:
> > I'm a little bit confusing about what regex should be used for the marker's text cache, any suggestion?
> 
> I guess in your case where the text might be any custom text (even not numbers) it doesn't really matter and you can use default one. The main purpose here why we should use TextWidthCache is caching and don't measure the same value twice.

It is possibly better if markers' text has its own font setting with `monospace` as default, since text could be anything besides number, it will be very easy to reach the caching limit (default is 50 if I remember it right) if there is a lot of markers.

Or maybe just raise the caching limit?

timocov:
> Or maybe just raise the caching limit?

Afaik that limit is per instance, so we'll reach it only in case of having a lot of different texts inside the same instance of TextWidthCache. I'd say that we case resize it for texts count. @eugene-korobko thoughts?

timocov:
> It is possibly better if markers' text has its own font setting with monospace as default

Afaik we have `layout.fontFamily`/`layout.fontSize` options and every text should be rendered with that font params. Quite possible that we need to use it here to set the font before rendering and calculating actual height of the text.

scrwdrv:
I made the following changes:
- `LayoutOptions.fontFamily` and  `LayoutOptions.fontSize` will be used by markers' text now
-  Reset the cache of text width when font family or font size changed
- Text of markers on the same bar will stack properly

image

image

Also, horizontal overlaps could still happens to markers text, not sure if it supposed to be added to `hitTest` since there is no width limit for marker.

scrwdrv:
> How about a hit box like this? Easier to test and avoid unwanted hover event.
> 
> > image

NVM, that was a bad idea.

timocov:
Hey @scrwdrv, after internal discussion with @eugene-korobko we think to merge this PR we need to have fixed https://github.com/tradingview/lightweight-charts/pull/349#discussion_r385779368 and added graphics test cases (see https://github.com/tradingview/lightweight-charts/tree/master/tests/e2e/graphics#writing-new-test-case).

src/renderers/series-markers-renderer.ts
```diff
@@ -53,42 +70,77 @@ export class SeriesMarkersRenderer extends ScaledRenderer {
...
+		ctx.textBaseline = 'middle';
+		ctx.font = this._font;
+
 		for (let i = this._data.visibleRange.from; i < this._data.visibleRange.to; i++) {
 			const item = this._data.items[i];
+			if (item.text !== undefined) {
+				item.text.width = this._textWidthCache.measureText(ctx, item.text.content);
+				item.text.height = this._fontSize;
+				item.text.x = item.text.x - item.text.width / 2 as Coordinate;
+			}
```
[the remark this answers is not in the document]
scrwdrv: Not sure what you meant, I thought `_drawImpl` was the draw function as you mentioned?

Also, `hitTest` did reuse these coordinates without recalculating since `_drawImpl` will be called first:

```js
	public hitTest(x: Coordinate, y: Coordinate): HoveredObject | null {
		if (this._data === null || this._data.visibleRange === null) {
			return null;
		}

		for (let i = this._data.visibleRange.from; i < this._data.visibleRange.to; i++) {
			const item = this._data.items[i];
			if (hitTestItem(item, x, y)) {
				return {
					hitTestData: item.internalId,
					externalId: item.externalId,
				};
			}
		}

		return null;
	}

	protected _drawImpl(ctx: CanvasRenderingContext2D, isHovered: boolean, hitTestData?: unknown): void {
		if (this._data === null || this._data.visibleRange === null) {
			return;
		}

		ctx.textBaseline = 'middle';
		ctx.font = this._font;

		for (let i = this._data.visibleRange.from; i < this._data.visibleRange.to; i++) {
			const item = this._data.items[i];
			if (item.text !== undefined) {
				item.text.width = this._textWidthCache.measureText(ctx, item.text.content);
				item.text.height = this._fontSize;
				item.text.x = item.text.x - item.text.width / 2 as Coordinate;
			}
			drawItem(item, ctx);
		}
	}
```

scrwdrv:
@timocov I have added a test case, including markers painted in following scenarios:
- Markers stacking in the same bar
- Different shape, position & size
- Empty string as text option

If there anything needs to be added as well please let me know.

Also, do I need to pass `Build Size` test as well? I assuming failing `dts-changes` and those graphics tests is expected since this is a new feature, but not quite sure about that size test.

src/renderers/series-markers-renderer.ts
```diff
@@ -53,42 +70,77 @@ export class SeriesMarkersRenderer extends ScaledRenderer {
...
+			case 'circle':
+				drawCircle(ctx, item.x, item.y, item.size);
+				return;
+			case 'square':
+				drawSquare(ctx, item.x, item.y, item.size);
+				return;
+		}
+	} else {
+		return;
 	}
```
timocov:
```suggestion
	if (item.size === 0) {
		return;
	}

	switch (item.shape) {
		case 'arrowDown':
			drawArrow(false, ctx, item.x, item.y, item.size);
			return;
		case 'arrowUp':
			drawArrow(true, ctx, item.x, item.y, item.size);
			return;
		case 'circle':
			drawCircle(ctx, item.x, item.y, item.size);
			return;
		case 'square':
			drawSquare(ctx, item.x, item.y, item.size);
			return;
	}
```

src/renderers/series-markers-renderer.ts
```diff
@@ -53,42 +70,77 @@ export class SeriesMarkersRenderer extends ScaledRenderer {
...
+			case 'arrowUp':
+				return hitTestArrow(false, item.x, item.y, item.size, x, y);
+			case 'circle':
+				return hitTestCircle(item.x, item.y, item.size, x, y);
+			case 'square':
+				return hitTestSquare(item.x, item.y, item.size, x, y);
+		}
+	} else {
+		return false;
 	}
```
timocov:
```suggestion
	if (item.size === 0) {
		return false;
	}

	switch (item.shape) {
		case 'arrowDown':
			return hitTestArrow(true, item.x, item.y, item.size, x, y);
		case 'arrowUp':
			return hitTestArrow(false, item.x, item.y, item.size, x, y);
		case 'circle':
			return hitTestCircle(item.x, item.y, item.size, x, y);
		case 'square':
			return hitTestSquare(item.x, item.y, item.size, x, y);
	}
```

src/views/pane/series-markers-pane-view.ts
```diff
@@ -77,13 +92,20 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
...
 		this._series = series;
 		this._model = model;
 		this._data = {
 			items: [],
 			visibleRange: null,
 		};
+
+		const layout = model.options().layout;
+		this._chartModel = model;
+		this._fontSize = layout.fontSize;
```
timocov:
```suggestion
	private _fontSize: number;

	public constructor(series: Series, model: ChartModel) {
		this._series = series;
		this._model = model;
		this._data = {
			items: [],
			visibleRange: null,
		};
```

src/views/pane/series-markers-pane-view.ts
```diff
@@ -98,7 +120,13 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
 		if (this._invalidated) {
 			this._makeValid();
 		}
+
+		const layout = this._chartModel.options().layout;
+		this._fontSize = layout.fontSize;
+
+		this._renderer.setParams(layout.fontSize, layout.fontFamily);
```
timocov:
```suggestion
		const layout = this._model.options().layout;
		this._renderer.setParams(layout.fontSize, layout.fontFamily);
```

src/views/pane/series-markers-pane-view.ts
```diff
@@ -136,6 +164,7 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
 				color: marker.color,
 				internalId: marker.internalId,
 				externalId: marker.id,
+				text: undefined,
 			}));
 			this._dataInvalidated = false;
 		}
```
timocov:
```suggestion
		}

		const layoutOptions = this._model.options().layout;
```

src/views/pane/series-markers-pane-view.ts
```diff
@@ -171,11 +200,20 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
...
+					width: 0,
+					height: 0,
+				};
+			}
 			const dataAt = this._series.dataAt(marker.time);
 			if (dataAt === null) {
 				continue;
 			}
-			fillSizeAndY(rendererItem, marker, dataAt, offsets, shapeMargin, priceScale, timeScale, firstValue.value);
+			fillSizeAndY(rendererItem, marker, dataAt, offsets, this._fontSize, shapeMargin, priceScale, timeScale, firstValue.value);
```
timocov:
```suggestion
			fillSizeAndY(rendererItem, marker, dataAt, offsets, layoutOptions.fontSize, shapeMargin, priceScale, timeScale, firstValue.value);
```

src/views/pane/series-markers-pane-view.ts
```diff
@@ -42,20 +43,34 @@ function fillSizeAndY(
...
+				rendererItem.text.y = rendererItem.y + halfSize + shapeMargin / 2 + textHeight * 0.6 as Coordinate;
+			}
 			return;
 		}
 		case 'aboveBar': {
-			rendererItem.y = (priceScale.priceToCoordinate(highPrice, firstValue) - shapeSize / 2 - offsets.aboveBar) as Coordinate;
+			rendererItem.y = (priceScale.priceToCoordinate(highPrice, firstValue) - halfSize - offsets.aboveBar) as Coordinate;
+			if (rendererItem.text !== undefined) {
+				rendererItem.text.y = rendererItem.y - halfSize - textHeight * 0.6 as Coordinate;
+				offsets.aboveBar += textHeight * 1.2;
```
timocov: What 0.6 and 1.2 stands for?

docs/series-basics.md
```diff
@@ -339,6 +341,7 @@ series.setMarkers([
         color: 'orange',
         shape: 'arrowUp',
         id: 'id4',
+        text: 'example',
     },
```
timocov:
```suggestion
        text: 'example',
        size: 2,
    },
```

docs/series-basics.md
```diff
@@ -313,8 +313,10 @@ An array of items is expected. Each item should contain the following fields:
 - `time` ([Time](./time.md)) - item time
 - `position` (`aboveBar` &#124; `belowBar` &#124; `inBar`) - item position
 - `shape` (`circle` &#124; `square` &#124; `arrowUp` &#124; `arrowDown`) - item marker type
+- `size` (`number` &#124; `undefined`) - size multiplier of marker, the shape is hidden when set to `0`
```
timocov:
```suggestion
- `size` (`number` &#124; `undefined`) - size multiplier of the marker, the shape is hidden when set to `0`, default value is `1`
```

src/views/pane/series-markers-pane-view.ts
```diff
@@ -77,6 +92,8 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
 
 	private _renderer: SeriesMarkersRenderer = new SeriesMarkersRenderer();
 
+	private _fontSize: number;
+
```
timocov:
```suggestion
```

src/views/pane/series-markers-pane-view.ts
timocov: Can we leave a comment or/and even extract that constants into `const enum Constants` with proper name to avoid confusing in the future?

scrwdrv: It adds one tenth of text height as margin space, `1.2` = `1 + 0.1 * 2`.
Since I set `textBaseline` to `middle`, text height should be divided by 2 when calculating corresponding y coordinate.

src/views/pane/series-markers-pane-view.ts
```diff
@@ -23,6 +23,10 @@ import {
 
 import { IUpdatablePaneView, UpdateType } from './iupdatable-pane-view';
 
+const enum Constants {
+	textMargin = 0.1,
```
timocov:
```suggestion
	TextMargin = 0.1,
```

src/views/pane/series-markers-pane-view.ts
```diff
@@ -52,24 +56,24 @@ function fillSizeAndY(
 		case 'inBar': {
 			rendererItem.y = priceScale.priceToCoordinate(inBarPrice, firstValue);
 			if (rendererItem.text !== undefined) {
-				rendererItem.text.y = rendererItem.y + halfSize + shapeMargin / 2 + textHeight * 0.6 as Coordinate;
+				rendererItem.text.y = rendererItem.y + halfSize + shapeMargin + textHeight * (0.5 + Constants.textMargin) as Coordinate;
```
timocov:
```suggestion
				rendererItem.text.y = rendererItem.y + halfSize + shapeMargin + textHeight * (0.5 + Constants.TextMargin) as Coordinate;
```

src/views/pane/series-markers-pane-view.ts
```diff
@@ -52,24 +56,24 @@ function fillSizeAndY(
...
 			}
 			return;
 		}
 		case 'aboveBar': {
 			rendererItem.y = (priceScale.priceToCoordinate(highPrice, firstValue) - halfSize - offsets.aboveBar) as Coordinate;
 			if (rendererItem.text !== undefined) {
-				rendererItem.text.y = rendererItem.y - halfSize - textHeight * 0.6 as Coordinate;
-				offsets.aboveBar += textHeight * 1.2;
+				rendererItem.text.y = rendererItem.y - halfSize - textHeight * (0.5 + Constants.textMargin) as Coordinate;
+				offsets.aboveBar += textHeight * (1 + 2 * Constants.textMargin);
```
timocov:
```suggestion
				rendererItem.text.y = rendererItem.y - halfSize - textHeight * (0.5 + Constants.TextMargin) as Coordinate;
				offsets.aboveBar += textHeight * (1 + 2 * Constants.TextMargin);
```

src/views/pane/series-markers-pane-view.ts
```diff
@@ -52,24 +56,24 @@ function fillSizeAndY(
...
 			offsets.aboveBar += shapeSize + shapeMargin;
 			return;
 		}
 		case 'belowBar': {
 			rendererItem.y = (priceScale.priceToCoordinate(lowPrice, firstValue) + halfSize + offsets.belowBar) as Coordinate;
 			if (rendererItem.text !== undefined) {
-				rendererItem.text.y = rendererItem.y + halfSize + shapeMargin + textHeight * 0.6 as Coordinate;
-				offsets.belowBar += textHeight * 1.2;
+				rendererItem.text.y = rendererItem.y + halfSize + shapeMargin + textHeight * (0.5 + Constants.textMargin) as Coordinate;
+				offsets.belowBar += textHeight * (1 + 2 * Constants.textMargin);
```
timocov:
```suggestion
				rendererItem.text.y = rendererItem.y + halfSize + shapeMargin + textHeight * (0.5 + Constants.TextMargin) as Coordinate;
				offsets.belowBar += textHeight * (1 + 2 * Constants.TextMargin);
```

timocov:
Looks good for me. Thank you a lot @scrwdrv for your work and being patient with the code review ❤️

Issue #335, opened 2020-02-17
Visible time range should have bars count of the space from left/right


nilsenpaul:
What I tried to do, is display a 1 day chart (with only the data for one day), and then load more data if a user zooms out (or more specific data if a user zooms in, but that's not the problem here).

In order to do that, I'd need to know the timestamps for the new (after scroll) boundaries of the chart:

![Schermafbeelding 2020-02-17 om 08 55 22](https://user-images.githubusercontent.com/6794819/74634316-e42d2f00-5163-11ea-91a9-fd4082b1f260.png)

There may be a way to get that, but I didn't find it yet.

timocov:
We can't provide timestamps for spaces from left and right (because we can't approximate them), but I think we can provide offsets _in bars_ from right and left.

timocov:
Actually right bars count might be taken from https://github.com/tradingview/lightweight-charts/blob/master/docs/time-scale.md#scrollposition - if the value is positive - it's exactly offset from the right :)

timocov:
As one of possible solutions I'd suggest to add 2 fields to [`TimeRange`](https://github.com/tradingview/lightweight-charts/blob/e0e1e2c10d634745187179abf86a5288a14998e9/src/api/itime-scale-api.ts#L7-L10) interface with bars before and bars after count and fill them in every place (I can't tell you for sure the names of that properties, but it might be something like `barsCountBefore`/`barsCountAfter` or just `barsBefore`/`barsAfter`, @tradingview/lightweight-charts-ios-maintainers any thoughts?).

gkaindl:
I don't want to hijack this issue, but since this functionality would be quite crucial for me as well, I'd like to implement it and file a pull request, but as per CONTRIBUTING.md, I wanted to make sure that this is welcome (the issue doesn't have the "help wanted" tag).

Adding `barsBefore / barsAfter` sounds like a viable approach to me, or is there a better proposal? The calculations are already available and used internally, so it's really more or less a matter of exporting them into the public API.

Additionally, I think it'd also make sense to introduce a method to query the chart object for the amount of bars it can currently display at most in the viewport (based off of `width` and `barSpacing`).

My use-case would be to determine the amount of data to load up-front in order to fill the chart viewport completely when displaying it, then subscribing to the timeScale change event to dynamically load additional chunks while the user scrolls or zooms the chart.

timocov:
[...]

Issue #209, opened 2019-09-16
Omit value in candlestick chart


primozv:
**Lightweight Charts Version:** 
1.2.0 

**Steps/code to reproduce:**

To leave a value empty for a line or a histogram while you set the data with series.setData you can simply omit the value. 
ie. series.setData({ time: '2019-04-11' });

When you set data by omiting the value (high,low,close,open)  for candlestick series it fails due to value being undefined in the following code:
    function ensureDefined(value) {
        if (value === undefined) {
            throw new Error('Value is undefined');
        }
        return value;
    }

Can I do that differently? Is it even possible to omit a value for candlesticks? Is there a workaround?

timocov:
Can you please describe a little why do you need this and what do you want to achieve by doing this?

primozv:
Sometimes data is missing or is not available but we still want to assure that one bar is 1 minute (if one bar represents 1 minute) not 2 miuntes , or 5, or 10. The visualization in case of missing bars is deceiving and doesn't show time, it shows data entries.
It already works correctly for series with simple/single value per bar. Candlesticks on the other hand fail due to value being undefined.
I hope this helps.

timocov:
I got it, sounds reasonable.

sysopsnw18:
I tired sending empty data part for area series and below is the output. Is this empty data part possible in area series?
![image](https://user-images.githubusercontent.com/61038438/80703425-6345c880-8b00-11ea-99b4-b9da485cec36.png)

sysopsnw18:
> I tired sending empty data part for area series and below is the output. Is this empty data part possible in area series?
> ![image](https://user-images.githubusercontent.com/61038438/80703425-6345c880-8b00-11ea-99b4-b9da485cec36.png)

Hey any update on this? When this feature will be available?

timocov:
> I tired sending empty data part for area series and below is the output. Is this empty data part possible in area series?

@sysopsnw18 Yes, we'll handle it in the fix for this issue. cc @kirchet 

> Hey any update on this? When this feature will be available?

We don't have any ETA, but it'll be fixed in 3.0 (assigned milestone).

timocov:
Needs to be sure that https://github.com/tradingview/lightweight-charts/issues/152#issuecomment-637107459 will be covered.

PR #370, opened 2020-03-07
Replaced TimeRange with TimeRangeWithBars in getVisibleRange

gkaindl:
TimeRangeWithBars extends the TimeRange interface to include the amount of bars missing before and after the data set in order to fill the entire chart viewport.

The idea is to use these fields to determine the amount of data to load dynamically, as the user scrolls and zooms in the chart.

**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: adresses #335 
- [ ] Includes tests
- [x] Documentation update

**Overview of change:**

The pull request addresses the addition of `barsBefore/barsAfter` fields to `getVisibleRange()` in `time-scale.ts` (and also in the handler of visible time range change events) by extending the `TimeRange` interface to `TimeRangeWithBars`. This way, an API client can determine how much additional data to load dynamically when the users zoom and pans through the chart.

The discussion for this functionality (as well as my extended rationale) is in #335

### DIFF
--- a/docs/time-scale.md
+++ b/docs/time-scale.md
@@ -68,7 +68,7 @@ chart.timeScale().scrollToRealTime();
 
 ### getVisibleRange()
 
-Returns the current visible time range of the chart as an object with the first and last time points of the time range, or returns `null` if the chart has no data at all.
+Returns the current visible time range of the chart as an object with the first and last time points of the time range, or returns `null` if the chart has no data at all. The additional fields `barsBefore` and `barsAfter` denote the amount of bars outside of the viewport in the given direction. If the number is negative, it means that the given amount of bars is missing in the data set on this side in order to fill the viewport completely.
 
 ```javascript
 chart.timeScale().getVisibleRange();
--- a/docs/time-scale.md
+++ b/docs/time-scale.md
@@ -68,7 +68,7 @@ chart.timeScale().scrollToRealTime();
 
 ### getVisibleRange()
 
-Returns the current visible time range of the chart as an object with the first and last time points of the time range, or returns `null` if the chart has no data at all. The additional fields `barsBefore` and `barsAfter` denote the amount of bars outside of the viewport in the given direction. If the number is negative, it means that the given amount of bars is missing in the data set on this side in order to fill the viewport completely.
+Returns the current visible time range of the chart as an object with the first and last time points of the time range, or returns `null` if the chart has no data at all.
 
 ```javascript
 chart.timeScale().getVisibleRange();
--- a/src/api/iseries-api.ts
+++ b/src/api/iseries-api.ts
@@ -7,14 +7,15 @@ import {
 	SeriesPartialOptionsMap,
 	SeriesType,
 } from '../model/series-options';
-import { TimePoint, TimePointIndexRange } from '../model/time-data';
+import { TimePoint } from '../model/time-data';
 
 import { SeriesDataItemTypeMap, Time } from './data-consumer';
 import { IPriceLine } from './iprice-line';
+import { LogicalRange } from './itime-scale-api';
 
 export interface BarsInfo {
-	from: TimePoint;
-	to: TimePoint;
+	from?: TimePoint;
+	to?: TimePoint;
 	barsBefore: number;
 	barsAfter: number;
 }
@@ -51,16 +52,16 @@ export interface ISeriesApi<TSeriesType extends SeriesType> {
 	coordinateToPrice(coordinate: Coordinate): BarPrice | null;
 
 	/**
-	 * Retrieves information about the series' data within a given index range.
-	 * @param range - the index range to retrieve info for
-	 * @returns - the bars info for the given index range: fields `from` and `to` are
-	 *   `Time` values for the first and last bar within the range, and `barsBefore` and
+	 * Retrieves information about the series' data within a given logical range.
+	 * @param range - the logical range to retrieve info for
+	 * @returns - the bars info for the given logical range: fields `from` and `to` are
+	 *   `Logical` values for the first and last bar within the range, and `barsBefore` and
 	 *   `barsAfter` count the the available bars outside the given index range. If these
 	 *   values are negative, it means that the given range us not fully filled with bars
 	 *   on the given side, but bars are missing instead (would show up as a margin if the
 	 *   the given index range falls into the viewport).
 	 */
-	barsInIndexRange(range: TimePointIndexRange): BarsInfo | null;
+	barsInLogicalRange(range: LogicalRange): BarsInfo | null;
 
 	/**
 	 * Applies new options to the existing series
--- a/src/api/iseries-api.ts
+++ b/src/api/iseries-api.ts
@@ -7,14 +7,14 @@ import {
 	SeriesPartialOptionsMap,
 	SeriesType,
 } from '../model/series-options';
-import { Time, TimePointIndexRange } from '../model/time-data';
+import { TimePoint, TimePointIndexRange } from '../model/time-data';
 
-import { SeriesDataItemTypeMap } from './data-consumer';
+import { SeriesDataItemTypeMap, Time } from './data-consumer';
 import { IPriceLine } from './iprice-line';
 
 export interface BarsInfo {
-	from: Time;
-	to: Time;
+	from: TimePoint;
+	to: TimePoint;
 	barsBefore: number;
 	barsAfter: number;
 }
--- a/src/api/iseries-api.ts
+++ b/src/api/iseries-api.ts
@@ -7,9 +7,17 @@ import {
 	SeriesPartialOptionsMap,
 	SeriesType,
 } from '../model/series-options';
+import { Time, TimePointIndexRange } from '../model/time-data';
 
-import { SeriesDataItemTypeMap, Time } from './data-consumer';
+import { SeriesDataItemTypeMap } from './data-consumer';
 import { IPriceLine } from './iprice-line';
+
+export interface BarsInfo {
+	from: Time;
+	to: Time;
+	barsBefore: number;
+	barsAfter: number;
+}
 
 /** Interface to be implemented by the object in order to be used as a price formatter */
 export interface IPriceFormatter {
@@ -41,6 +49,18 @@ export interface ISeriesApi<TSeriesType extends SeriesType> {
 	 * @returns - price value of the coordinate on the chart
 	 */
 	coordinateToPrice(coordinate: Coordinate): BarPrice | null;
+
+	/**
+	 * Retrieves information about the series' data within a given index range.
+	 * @param range - the index range to retrieve info for
+	 * @returns - the bars info for the given index range: fields `from` and `to` are
+	 *   `Time` values for the first and last bar within the range, and `barsBefore` and
+	 *   `barsAfter` count the the available bars outside the given index range. If these
+	 *   values are negative, it means that the given range us not fully filled with bars
+	 *   on the given side, but bars are missing instead (would show up as a margin if the
+	 *   the given index range falls into the viewport).
+	 */
+	barsInIndexRange(range: TimePointIndexRange): BarsInfo | null;
 
 	/**
 	 * Applies new options to the existing series
--- a/src/api/itime-scale-api.ts
+++ b/src/api/itime-scale-api.ts
@@ -1,17 +1,21 @@
 import { DeepPartial } from '../helpers/strict-type-checks';
 
-import { TimePointIndexRange } from '../model/time-data';
 import { TimeScaleOptions } from '../model/time-scale';
 
-import { Time } from './data-consumer';
+import { Logical, Time } from './data-consumer';
 
 export interface TimeRange {
 	from: Time;
 	to: Time;
 }
 
+export interface LogicalRange {
+	from: Logical;
+	to: Logical;
+}
+
 export type TimeRangeChangeEventHandler = (timeRange: TimeRange | null) => void;
-export type TimePointIndexRangeChangeEventHandler = (indexRange: TimePointIndexRange | null) => void;
+export type LogicalRangeChangeEventHandler = (logicalRange: LogicalRange | null) => void;
 
 /** Interface to chart time scale */
 export interface ITimeScaleApi {
@@ -46,16 +50,16 @@ export interface ITimeScaleApi {
 	setVisibleRange(range: TimeRange): void;
 
 	/**
-	 * Returns the currently visible index range of data.
+	 * Returns the currently visible logical range of data.
 	 * @returns - visible range or null if the chart has no data at all
 	 */
-	getVisibleIndexRange(): TimePointIndexRange | null;
+	getVisibleLogicalRange(): LogicalRange | null;
 
 	/**
-	 * Sets visible index range of data.
-	 * @param range - target visible index range of data.
+	 * Sets visible logical range of data.
+	 * @param range - target visible logical range of data.
 	 */
-	setVisibleIndexRange(range: TimePointIndexRange): void;
+	setVisibleLogicalRange(range: LogicalRange): void;
 
 	/**
 	 * Restores default zooming and scroll position of the time scale
@@ -84,13 +88,13 @@ export interface ITimeScaleApi {
 	 * Adds a subscription to visible index range changes to receive notifications about visible indexes of the data
 	 * @param handler - handler (function) to be called when the visible indexes change
 	 */
-	subscribeVisibleIndexRangeChange(handler: TimePointIndexRangeChangeEventHandler): void;
+	subscribeVisibleLogicalRangeChange(handler: LogicalRangeChangeEventHandler): void;
 
 	/**
 	 * Removes a subscription to visible index range changes
 	 * @param handler - previously subscribed handler
 	 */
-	unsubscribeVisibleIndexRangeChange(handler: TimePointIndexRangeChangeEventHandler): void;
+	unsubscribeVisibleLogicalRangeChange(handler: LogicalRangeChangeEventHandler): void;
 
 	/**
 	 * Applies new options to the time scale.
--- a/src/api/itime-scale-api.ts
+++ b/src/api/itime-scale-api.ts
@@ -9,6 +9,8 @@ export interface TimeRange {
 	to: Time;
 }
 
+export type TimeRangeChangeEventHandler = (timeRange: TimeRange | null) => void;
+
 /** Interface to chart time scale */
 export interface ITimeScaleApi {
 	/**
@@ -51,6 +53,18 @@ export interface ITimeScaleApi {
 	 * This is a momentary operation.
 	 */
 	fitContent(): void;
+
+	/**
+	 * Adds a subscription to visible range changes to receive notification about visible range of data changes
+	 * @param handler - handler (function) to be called on changing visible data range
+	 */
+	subscribeVisibleTimeRangeChange(handler: TimeRangeChangeEventHandler): void;
+
+	/**
+	 * Removes a subscription to visible range changes
+	 * @param handler - previously subscribed handler
+	 */
+	unsubscribeVisibleTimeRangeChange(handler: TimeRangeChangeEventHandler): void;
 
 	/**
 	 * Applies new options to the time scale.
--- a/src/api/itime-scale-api.ts
+++ b/src/api/itime-scale-api.ts
@@ -1,7 +1,14 @@
 import { DeepPartial } from '../helpers/strict-type-checks';
 
-import { TimePointIndexRange, TimeRange } from '../model/time-data';
+import { TimePointIndexRange } from '../model/time-data';
 import { TimeScaleOptions } from '../model/time-scale';
+
+import { Time } from './data-consumer';
+
+export interface TimeRange {
+	from: Time;
+	to: Time;
+}
 
 export type TimeRangeChangeEventHandler = (timeRange: TimeRange | null) => void;
 export type TimePointIndexRangeChangeEventHandler = (indexRange: TimePointIndexRange | null) => void;
--- a/src/api/itime-scale-api.ts
+++ b/src/api/itime-scale-api.ts
@@ -9,11 +9,6 @@ export interface TimeRange {
 	to: Time;
 }
 
-export interface TimeRangeWithBars extends TimeRange {
-	barsBefore: number;
-	barsAfter: number;
-}
-
 /** Interface to chart time scale */
 export interface ITimeScaleApi {
 	/**
@@ -38,7 +33,7 @@ export interface ITimeScaleApi {
 	 * Returns current visible time range of the chart
 	 * @returns - visible range or null if the chart has no data at all
 	 */
-	getVisibleRange(): TimeRangeWithBars | null;
+	getVisibleRange(): TimeRange | null;
 
 	/**
 	 * Sets visible range of data
--- a/src/api/itime-scale-api.ts
+++ b/src/api/itime-scale-api.ts
@@ -1,14 +1,7 @@
 import { DeepPartial } from '../helpers/strict-type-checks';
 
-import { TimePointIndexRange } from '../model/time-data';
+import { TimePointIndexRange, TimeRange } from '../model/time-data';
 import { TimeScaleOptions } from '../model/time-scale';
-
-import { Time } from './data-consumer';
-
-export interface TimeRange {
-	from: Time;
-	to: Time;
-}
 
 export type TimeRangeChangeEventHandler = (timeRange: TimeRange | null) => void;
 export type TimePointIndexRangeChangeEventHandler = (indexRange: TimePointIndexRange | null) => void;
--- a/src/api/itime-scale-api.ts
+++ b/src/api/itime-scale-api.ts
@@ -11,6 +11,7 @@ export interface TimeRange {
 }
 
 export type TimeRangeChangeEventHandler = (timeRange: TimeRange | null) => void;
+export type TimePointIndexRangeChangeEventHandler = (indexRange: TimePointIndexRange | null) => void;
 
 /** Interface to chart time scale */
 export interface ITimeScaleApi {
@@ -78,6 +79,18 @@ export interface ITimeScaleApi {
 	 * @param handler - previously subscribed handler
 	 */
 	unsubscribeVisibleTimeRangeChange(handler: TimeRangeChangeEventHandler): void;
+
+	/**
+	 * Adds a subscription to visible index range changes to receive notifications about visible indexes of the data
+	 * @param handler - handler (function) to be called when the visible indexes change
+	 */
+	subscribeVisibleIndexRangeChange(handler: TimePointIndexRangeChangeEventHandler): void;
+
+	/**
+	 * Removes a subscription to visible index range changes
+	 * @param handler - previously subscribed handler
+	 */
+	unsubscribeVisibleIndexRangeChange(handler: TimePointIndexRangeChangeEventHandler): void;
 
 	/**
 	 * Applies new options to the time scale.
--- a/src/api/itime-scale-api.ts
+++ b/src/api/itime-scale-api.ts
@@ -1,5 +1,6 @@
 import { DeepPartial } from '../helpers/strict-type-checks';
 
+import { TimePointIndexRange } from '../model/time-data';
 import { TimeScaleOptions } from '../model/time-scale';
 
 import { Time } from './data-consumer';
@@ -42,6 +43,18 @@ export interface ITimeScaleApi {
 	 * @param range - target visible range of data
 	 */
 	setVisibleRange(range: TimeRange): void;
+
+	/**
+	 * Returns the currently visible index range of data.
+	 * @returns - visible range or null if the chart has no data at all
+	 */
+	getVisibleIndexRange(): TimePointIndexRange | null;
+
+	/**
+	 * Sets visible index range of data.
+	 * @param range - target visible index range of data.
+	 */
+	setVisibleIndexRange(range: TimePointIndexRange): void;
 
 	/**
 	 * Restores default zooming and scroll position of the time scale
--- a/src/api/itime-scale-api.ts
+++ b/src/api/itime-scale-api.ts
@@ -9,6 +9,11 @@ export interface TimeRange {
 	to: Time;
 }
 
+export interface TimeRangeWithBars extends TimeRange {
+	barsBefore: number;
+	barsAfter: number;
+}
+
 /** Interface to chart time scale */
 export interface ITimeScaleApi {
 	/**
@@ -33,7 +38,7 @@ export interface ITimeScaleApi {
 	 * Returns current visible time range of the chart
 	 * @returns - visible range or null if the chart has no data at all
 	 */
-	getVisibleRange(): TimeRange | null;
+	getVisibleRange(): TimeRangeWithBars | null;
 
 	/**
 	 * Sets visible range of data
--- a/src/api/series-api.ts
+++ b/src/api/series-api.ts
@@ -3,6 +3,7 @@ import { clone, merge } from '../helpers/strict-type-checks';
 
 import { BarPrice } from '../model/bar';
 import { Coordinate } from '../model/coordinate';
+import { PlotRowSearchMode } from '../model/plot-list';
 import { PriceLineOptions } from '../model/price-line-options';
 import { Series } from '../model/series';
 import { SeriesMarker } from '../model/series-markers';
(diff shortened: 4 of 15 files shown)


gkaindl:
After actually implementing the dynamic data loading functionality in my own app for which I needed the new fields, I realized I made a design mistake:

In the original version of the pull request, I set up the `barsBefore` and `barsAfter` fields, so that they would always be 0, as long as the viewport is completely filled on the given side, and then turn positive if an empty gap starts showing there. E.g., the semantics of the fields was something like "The amount of bars missing on this side to fill the viewport."

However, a (to my mind) better behavior is for the fields to be the amount of bars in the current data set that our outside of the viewport to the left and right edges. The number are positive, as long as there are bars left outside the viewport, but turns negative when a gap starts showing. The magnitude of the negative value is the amount of bars missing on this side of the viewport.

This change enables two behaviors that I find more useful than what was originally possible:

- Previously, I would request additional data once the bar count field in a direction would. turn positive, e.g. bars are already missing in the viewport. With the new behavior, I can already start loading additional data once only a "half a page" or so of extra bars become available in the direction the user is scrolling. If your API is fast, you can thus provide the illusion of "endless data", since the user might never even see an empty gap anymore.

- As the user scrolls very far towards one direction (and thus, the bar count in the direction away from the scroll direction increases), you can eventually start to drop data points in order to not to have too many data points in memory (provided you can re-request them later, should the user scroll back).

I'm of course open to feedback! I find this behavior more useful, and I think the idea of positive values meaning "there is this amount of data still left" and negative values meaning "this amount of data is missing" also provides for a nice mental model.

Also, I don't think I can fix the failing test in CI by myself, since I had to change the dts in order to introduce the `TimeRangeWithBars` interface.

src/api/time-scale-api.ts
```diff
@@ -55,6 +55,8 @@ export class TimeScaleApi implements ITimeScaleApi, IDestroyable {
 		return {
 			from: timePointToTime(ensureNotNull(points.valueAt(Math.max(firstIndex, visibleBars.firstBar()) as TimePointIndex))),
 			to: timePointToTime(ensureNotNull(points.valueAt(Math.min(lastIndex, visibleBars.lastBar()) as TimePointIndex))),
+			barsBefore: visibleBars.firstBar(),
```
timocov: `visibleBars.firstBar()` is an index of the first visible bar, thus if we shift a chart a bit to the history, firstBar will be negative value, so I think we need to have here something like `Math.min(0, -visibleBars.firstBar())` (`Math.min` here to avoid negative values which means nothing if we have the date of the leftmost visible bar).

src/api/time-scale-api.ts
```diff
@@ -55,6 +55,8 @@ export class TimeScaleApi implements ITimeScaleApi, IDestroyable {
...
+			barsAfter: -Math.ceil(this.scrollPosition()),
```
timocov: I guess we need to have Math.min here as well. Also, what's about `visibleBars.lastBar() - lastIndex` value? As you've noticed scrollPosition might be double value, but indexes can't and having logic around indexes is more "convenient".

docs/time-scale.md
```diff
@@ -68,7 +68,7 @@ chart.timeScale().scrollToRealTime();
 
 ### getVisibleRange()
 
-Returns the current visible time range of the chart as an object with the first and last time points of the time range, or returns `null` if the chart has no data at all.
+Returns the current visible time range of the chart as an object with the first and last time points of the time range, or returns `null` if the chart has no data at all. The additional fields `barsBefore` and `barsAfter` denote the amount of bars outside of the viewport in the given direction. If the number is negative, it means that the given amount of bars is missing in the data set on this side in order to fill the viewport completely.
```
timocov: > the amount of bars outside of the viewport in the given direction

Hm, I thought that we want to have amount of bars _before_ the data, if data is ended early, don't we?

timocov:
> Also, I don't think I can fix the failing test in CI by myself, since I had to change the dts in order to introduce the TimeRangeWithBars interface.

`dts-changed`'s fail is ok since it shows that "something is changed is generated dts file so please take a look at the provided changes/API".

> With the new behavior, I can already start loading additional data once only a "half a page" or so of extra bars become available in the direction the user is scrolling

Hm, that sounds reasonable. You meant that you can start to load the data right _before_ an user scroll to the empty space, right? Something like mobile apps or websites do when you scroll down the page with some content (like Twitter for instance).

@tradingview/lightweight-charts-maintainers hey guys, can you please take a look here as well?

gkaindl:
> Hm, that sounds reasonable. You meant that you can start to load the data right _before_ an user scroll to the empty space, right? Something like mobile apps or websites do when you scroll down the page with some content (like Twitter for instance).
> 

Yes, exactly. If you check out the original commit of the pull request (c6fb3c81d1eaad2791c013526a7bd725bbb2116d), I implemented it exactly as you were suggesting in your comments above, e.g. so that `barsBefore`/`barsAfter` is 0 if the viewport is completely filled in this given direction (no missing bars), and turns positive as gaps start to show up, the number denoting the amount of bars missing.

Next, I implemented the actual "endless scrolling" functionality in my own app, based on this. It was set up so that as soon as the `barsBefore`/`barsAfter` field would turn positive, it would load another "page" of data (e.g. two viewports worth of extra bars), wait for the data, display it, and adjust the scroll position if necessary to make it all look seamless. It works nicely.

*However*, then I noticed that it would be nicer if – as you're describing it above – I could already load extra data *before* I'm running out of data on a side. With the added commits (and the changed behavior), my app is now set up to load data as soon as the user scroll so far towards one side that only "half a viewport width" of extra bars is still available outside the viewport. Since my API is rather fast, data will typically arrive before the user reaches the end of the available data, so the whole experience is now much like "endless scrolling" – As you mentioned, exactly like Twitter etc... would do with their timeline.

That's why I changed the behavior after my first commit.

In c6fb3c81d1eaad2791c013526a7bd725bbb2116d, the behavior was

- viewport is filled -> both `barsBefore`, `barsAfter` are 0
- gaps shows up -> `barsBefore` or `barsAfter` turn positive, counting the number of missing bars to fill the viewport again.

So in that commit, the logic was: The fields count the amount of bars needed to fill the whitespace visible in the current viewport, if any. The disadvantage was that you have to wait for whitespace to show up to determine that you're running out of data, which would be too late for proper "endless scrolling".

In 2184c9702f144bc7e3030c37de2a3635371a462a (and subsequently d37a19f77e59cde4dea829da37262df44440a5ee, because I messed up the sign), I changed the behavior to

- viewport is filled -> `barsBefore`/`barsAfter` are positive and count the amount of bars that are still available in the current data, but are outside the viewport. Thus, if `barsBefore` was 10, it would mean that there are 10 bars outside of the viewport to the left.
- viewport has gaps -> `barsBefore`/`barsAfter` turn negative, and their magnitude counts the amount of bars that are needed to fill the gap again

So in that version, the logic/mental model is that the fields denote the amount of bars that are still available in the given direction outside the viewport: If the number is positive, it means there are bars available outside the viewport, if it is negative, it means that you are missing some bars in that direction, so there is a visible gap in the viewport.

I think this behavior, though a bit trickier to describe, is much more useful than just counting the gap width when a gap shows up.

Other use-cases I could think of:

Culling data as the user scrolls far into one direction: (I'm doing this when merging in newly loaded data, but it would be just as valid to say "if I'm scrolling to the right so far that `barsBefore` becomes > 5000, let's remove 1000 items of data, so that we're not consuming too much memory for data outside the viewport).

Showing annotations above a chart of static data: Like, as an example, annotating the left/right chart edges with something like "more data available" or "scroll here"

That's my rationale – I'm of course open to changing the behavior, but I think it would be great if the possibility to do the "preload more data before a gap shows up" kind of "endless scrolling" behavior would be supported by an alternative as well.

src/api/time-scale-api.ts
gkaindl: I replied in a longer post below, since the following three comments are all interlinked and best addressed together.

src/api/itime-scale-api.ts
```diff
@@ -9,6 +9,11 @@ export interface TimeRange {
 	to: Time;
 }
 
+export interface TimeRangeWithBars extends TimeRange {
```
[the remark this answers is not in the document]
gkaindl: Yes, that would be a possibility I was thinking of, too!

My main reason for not adding support for `barsBefore`/`barsAfter` in `setVisibleRange()` is that it will be difficult to communicate the behavior to the user. I see the following issues:

1. Both `from`/`to` and `barsBefore`/`barsAfter` would be sufficient to choose a viewport, so we could either make all of them optional (but that makes the `TimeRange` type a bit strange), or there are two different types and overalds for `setVisibleRange()`, one retaining the current behavior, the other taking a new type like `BarsPosition` or so.

2. `to`/`from` and `barsBefore`/`barsAfter` can be in conflict, if both sets are specified: Let's suppose I have 200 data points between time X and time Y -> if I call `setVisibleRange()` with `from = X` now, but also `barsBefore = 10`, we can't "solve" this, since our data only starts at time X, so we don't have any bars before it. Either one of them takes precedence, or an error is reported/thrown, or the method does nothing if a conflict is found.

However, it's very well possible I'm missing a good use-case here or a way to implement it cleanly, since I haven't needed to extend `setVisibleRange()` for my own use-cases yet and thus didn't spend a lot of thought on it.

timocov:
> The disadvantage was that you have to wait for whitespace to show up to determine that you're running out of data, which would be too late for proper "endless scrolling".

Agree.

> I think this behavior, though a bit trickier to describe, is much more useful than just counting the gap width when a gap shows up.

Yeah, agree 👍

> Culling data as the user scrolls far into one direction: (I'm doing this when merging in newly loaded data, but it would be just as valid to say "if I'm scrolling to the right so far that barsBefore becomes > 5000, let's remove 1000 items of data, so that we're not consuming too much memory for data outside the viewport).

Hmm, interesting point, never thought about that in this case.

Overall I totally agree with you - this is good improvements 👍

I think there are 2 questions we need to solve (just to summarize):

1. What is more intuitive to have positive or negative values for what cases (I meant what mean that `barsBefore` or `barsAfter` is negative or positive at the first look).
1. Should we add `barsBefore`/`barsAfter` in `setVisibleRange` method or not and what meaning of that values should be.

For me, I'd expect that `barsBefore`/`barsAfter` value is amount of bars before/after _the visible range_. This is exactly what you described in the docs.

The second question is more tricky. From one side I understand you that that might be confusing to users, but in other hand it'd be great to have kind of "symmetric" here - calling `timeScale.setVisibleRange(timeScale.getVisibleRange())` should be "no-op" and changes nothing on the chart (to within 1 bar due floating nature of visible range under the hood).

What if we put `barsBefore`/`barsAfter` to the `TimeRange` interface, but for `setVisibleRange` allow to don't specify that values? (even maybe mark them as optional in `TimeRange`, but in `getVisibleRange` and in events change the type to `Required<TimeRange>`). In this case, inside the `setVisibleRange` we can use that values only in case of `from`/`to` dates refers to the first/last bar and if that values are negative. But this might be a bit confusing that `barsBefore`/`barsAfter` should be negative values here, not positive ones. I'm really not sure about that, just thoughts.

timocov:
Just read your reply https://github.com/tradingview/lightweight-charts/pull/370#discussion_r390213918 and understood that you already said what I wrote above 😂 

I requested @tradingview/lightweight-charts-maintainers guys to take a look at this, maybe we'll get some other thoughts.

gkaindl:
> Just read your reply #370 (comment) and understood that you already said what I wrote above 😂

Haha, no worries, I think splitting my answer up into two posts maybe wasn't a clever thing to do! :-D

> calling timeScale.setVisibleRange(timeScale.getVisibleRange()) should be "no-op" and changes nothing on the chart

Yes, with that I absolutely agree, this should definitely be possible. Right now, it would work because `TimeRangeWithBars` extends `TimeRange`, so it is downwards-compatible, but I also really like the idea of using `Required<TimeRange>` and optional properties instead, which additionally seems to be more TypeScript-idiomatic to me.

But yeah, I also think that the negative values would be confusing in `setVisibleRange()`, even though they intuitively make sense with `getVisibleRange()` – It's a bit of mental-model kind of issue, because during the scroll events, positive numbers for "more bars are available" and negative numbers for "bars are missing" make sense, but when setting the viewport, you would use the fields to *add* margins, and using negative numbers to add something feels backwards. Tricky!

Maybe if the others chime in, they might have a good idea how to deal with this in an intuitive way.

timocov:
Hey @gkaindl, sorry for the delay.

I just discussed this with my teammates and we'd like to suggest to change solution a bit in the following way:

1. Don't change nor `TimeScaleApi.getVisibleRange` or `TimeScaleApi.setVisibleRange` - they will still work with _time_ range (e.g. with "data", not indexes).

1. Move `subscribeVisibleTimeRangeChange` and `unsubscribeVisibleTimeRangeChange` to `TimeScaleApi` (to make it consistent - see below) (yeah, this is breaking change, but this PR will be in 3.0 release, so breaking change looks OK here).

1. Add 4 methods to the `TimeScaleApi`:

	- `getVisibleIndexesRange` - returns object with the new type `TimePointIndexRange` with properties `from` and `to` of type `TimePointIndex`
	- `setVisibleIndexesRange` - sets passed `TimePointIndexRange` object as current visible range
	- `subscribeVisibleIndexesRangeChange` - allows you subscribe on visible range changed (similar to `subscribeVisibleTimeRangeChange`, but provides `TimePointIndexRange`)
	- `unsubscribeVisibleIndexesRangeChange` - unsubscribe from indexes range updates (similar to `unsubscribeVisibleTimeRangeChange`)

	The advantages of this new API are that it 1) provides you symmetric API to get/set visible range 2) allows you set visible range with "empty spaces" from left/right. Yeah, it exposes "internal" knowledge about indexes, what we should describe in the docs (you can don't do it in the PR - we'll provide it as soon as the PR will be merged).

1. Add the new method `barsInIndexesRange` into `SeriesApi`, which have the only one argument of type `TimePointIndexRange` and return object with the following properties (I guess we can call it something like `BarsInfo`, but I assume that there might be more suitable name) filled for passed argument:

	- `from`, type `Time`
	- `to`, type `Time`
	- `barsBefore`, type `number`
	- `barsAfter`, type `number`

	This method is pretty similar to what we currently have in the PR in method `TimeScaleApi.getVisibleRange`, but for the only one series. This will allow you know how much bars before/after are for every series separately, and also "convert" indexes to series' times (series could have different sets of the data).

Thus, if you need to know how much bars you need to load, you should do something like that:

```js
const chart = createChart();
const timeScale = chart.timeScale();
const series = chart.addLineSeries();

function onIndexesRangeChanged(newRange) {
	const barsInfo = series.barsInIndexesRange(newRange);
	console.log(barsInfo.barsBefore, barsInfo.barsAfter);
}

timeScale.subscribeVisibleIndexesRangeChange(onIndexesRangeChanged);
```

What do you think?

cc @Nipheris @kirchet @eugene-korobko

gkaindl:
> Hey @gkaindl, sorry for the delay.

No worries at all, it's tough times right now, and I'm glad to see you and the team are doing well!

> I just discussed this with my teammates and we'd like to suggest to change solution a bit in the following way: ...

Yes, I think I understand all of the proposed changes and agree fully, it's a nicer and more flexible way to do it – Thanks for the input! I'll have time for it at the weekend at the latest!

One more suggestion, though: So far, I've kept the `barsBefore`/`barsAfter` fields as integer numbers, but I'd suggest to change this to fractional values, much like how `scrollPosition()` works now.

Since implementing the dynamic loading, I've also worked on a `ChartViewportSyncProvider` (I use React in my app), which should synchronize the viewports (as visible time range) between multiple chart instances – Basically what multiple panes in a single chart would do, but my use-case also includes charts that are at completely different places in the DOM, not necessarily all below each other and of the same width, like panes would work).

I got it working with a lot of hacks (e.g. hooking into various DOM events as well) – The main problem is that right now with the `getVisibleRange()/setVisibleRange()` API, you can only get/set the "integer" viewport, e.g. the first and last bar in the viewport, but charts can be scrolled smoothly, e.g. there can be parts of bars visible at the edges of the viewports. This is exacerbated by the fact that after `setVisibleRange()`, you'd have to wait for one animation frame to adjust `scrollPosition()` for fractional values again, since internally, it takes an animation frame for `setVisibleRange()` to actually update – Thus, this type of viewport sync *can* be done right now, but whenever the visibleRangeChanged event fires, there will at least be "jitter", since adapting the scroll position to account for fractional bars can only happen in the next frame.

If we allow `barsBefore`/`barsAfter` to be fractional, you could

- hook into `subscribeVisibleIndexRangeChange()` to be notified of these fractional changes during scroll/zoom events, e.g. this handler would simply fire way more often than `subscribeVisibleTimeRangeChange()`, which only tracks visible bars, not exact scroll positions.

- do your own calculations to relate the `BarsInfo` of one chart to the charts that should get synchronized to it (e.g. to account for different amounts of bar data, and so on), but that's all in the client app's hands, as the client app controls the data anyway.

- use `setVisibleIndexRange()` to set the new viewports in the sync'd charts, but allowing for fractional bars would also enable the client app to set a viewport that does not start or end directly between two bars, but shows parts of bars at the edges, eg. a "pixel-perfect" viewport sync.

I think that would be a great addition! My "collection" of use-cases where such a "smooth" viewport sync could be useful, but where it couldn't be done by using multiple panes (once they become an official feature):

- multiple charts that are placed at completely different positions in the DOM, or that are not supposed to be of the same width as the other charts in the group

- creating a "minimap", e.g. a smaller chart that shows the entirety of data and overlays a semi-transparent box over it that denotes the current viewport of a larger chart.

- Smooth synchronization of charts in different browser instances, which can be sync'd via webRTC or a websocket-based system (e.g. collaborative chart analysis).

The only downside I could see is that `IndexRange` is a bad naming for something that can have fractional values. Maybe the instances of `IndexRange` could be renamed to something like `PositionRange`, to bring it in line with `scrollPosition()`, which is a similar concept, and that's already fractional.

Anyway, I'll await your feedback and go ahead as outlined above. Implementation-wise, the only difference would be either `Math.floor/Math.ceil`-ing the values exposed to the user API or not, as well as the naming changes, so that's trivial to adapt if the team decides that fractional values are a no-go.

timocov:
May I ask you how do you sync charts between each other? What the point of the synchronization is? Is that left/right bar's time, or it's just "indexes"?

> right now with the getVisibleRange()/setVisibleRange() API, you can only get/set the "integer" viewport, e.g. the first and last bar in the viewport, but charts can be scrolled smoothly, e.g. there can be parts of bars visible at the edges of the viewports

Looks like we've stripped that from the lightweight-charts, but actually "internally" we have terms of "visible range" and "strict visible range". In lightweight-charts we have strict one for now, but looks like we need to put non-strict version back for this...

We'll dig into it soon and return back with results.

gkaindl:
> May I ask you how do you sync charts between each other? What the point of the synchronization is? Is that left/right bar's time, or it's just "indexes"?

Sure! It's probably the similar to "strict visible range", e.g. I synchronize the charts by "left bar's" time and "right bar's time", e.g. the entire viewport, but also take the partial bars into account that happen while scrolling, only that I'm doing it in a complicated process that uses `scrollPosition()` as well.

So what I'm achieving is that two charts show exactly the same viewport, including partial bars.

> Looks like we've stripped that from the lightweight-charts, but actually "internally" we have terms of "visible range" and "strict visible range". In lightweight-charts we have strict one for now, but looks like we need to put non-strict version back for this...
> 
> We'll dig into it soon and return back with results.

Yes, "strict visible range" sounds exactly like what I'm doing with workarounds now – After all, if the `barsBefore`/`barsAfter` fields are allowed to be fractional, and the client app still has access to the data it supplied to a series for rendering, then you can derive all values (times, indices, and partial scroll offsets) from them.

gkaindl:
I've implemented the changes as outlined above.

As discussed, the indexes in `getVisibleIndexRange()/setVisibleIndexRange()` and the related subscriptions are fractional, so that they reflect the visible state (e.g. the strict vs. non-strict range distinction that's been mentioned above). I think that makes a lot of sense and is very helpful, since there's currently no other way to retrieve the "strict visible range", e.g. the range including the fractional scroll offsets. Additionally, I noticed that `TimePointIndexRange` hasn't been used in public API before, so this won't break anything.

Since the values are fractional, it also means that the visible index range subscription will fire more often than the visible time range subscription. I think that's also a useful feature, especially because it's now very easy to use the subscription to synchronize the visible range of multiple charts during scrolling/panning/zooming, etc...

I've attached a debug.html that illustrates all this:

- The visible index range subscription is set up for the top chart and logs the `barsInIndexRange()` output (`barsBefore`/`barsAfter` behave as before).

- When scrolling/panning/zooming the top chart, the bottom chart reflects the visible range changes exactly. This is a simple example, since both charts are exactly the same, but I'm using a similar approach now in my own client app, where I additionally account for different amounts of data in multiple synchronized charts. The important thing is that I can now access the fractional bar information as well, so the visible viewports can be exact across multiple charts.

For me, all my needs for my use-cases are perfectly doable with the API in this pull request, which is great.

Let me know what you think!

src/model/time-data.ts
```diff
@@ -17,6 +17,8 @@ export interface TimePoint {
 	businessDay?: BusinessDay;
 }
 
+export type Time = UTCTimestamp | BusinessDay | string;
```
timocov, line 20: I'm afraid on model level we don't have `string` as Time, so it'd be better to revert this changes back to API level and use `TimePoint` instead.

src/api/iseries-api.ts
```diff
@@ -7,10 +7,18 @@ import {
...
 } from '../model/series-options';
+import { Time, TimePointIndexRange } from '../model/time-data';
 
-import { SeriesDataItemTypeMap, Time } from './data-consumer';
+import { SeriesDataItemTypeMap } from './data-consumer';
 import { IPriceLine } from './iprice-line';
 
+export interface BarsInfo {
+	from: Time;
+	to: Time;
```
timocov, line 17: should it be `TimePoint` instead?

src/api/time-scale-api.ts
```diff
@@ -66,6 +60,18 @@ export class TimeScaleApi implements ITimeScaleApi, IDestroyable {
 		this._model.setTargetTimeRange(convertedRange);
 	}
 
+	public getVisibleIndexRange(): TimePointIndexRange | null {
+		const indexRange = this._timeScale().visibleIndexRange();
+		return indexRange ? clone(indexRange) : null;
+	}
+
+	public setVisibleIndexRange(range: TimePointIndexRange): void {
+		if (range.from > range.to) {
```
timocov, line 69: what do you think about asserting here instead?

src/model/invalidate-mask.ts
```diff
@@ -85,6 +97,9 @@ export class InvalidateMask {
 		if (other._targetTimeRange) {
 			this.setTargetTimeRange(other._targetTimeRange);
 		}
+		if (other._targetIndexRange) {
+			this.setTargetIndexRange(other._targetIndexRange);
```
timocov, line 101: actually looks like we can use one of them here, because we can easily convert them from one to another

src/model/time-scale.ts
```diff
@@ -412,14 +453,13 @@ export class TimeScale {
 	}
 
 	public scrollTo(x: Coordinate): void {
-		this._visibleBarsInvalidated = true;
+		this._invalidateVisibleBars();
```
timocov, line 456: why should be invalidate visible bars in case of doing nothing in method? 🤔

src/model/time-scale.ts
```diff
@@ -142,6 +146,43 @@ export class TimeScale {
...
+		const from = Math.round(range.from);
+		const to = Math.round(range.to);
+
+		const points = this._model.timeScale().points();
+		const firstIndex = ensureNotNull(points.firstIndex());
+		const lastIndex = ensureNotNull(points.lastIndex());
+
+		return {
+			from: timePointToTime(ensureNotNull(points.valueAt(Math.max(firstIndex, from) as TimePointIndex))),
+			to: timePointToTime(ensureNotNull(points.valueAt(Math.min(lastIndex, to) as TimePointIndex))),
```
timocov, line 182: I think it'd better to provide TimePoint here instead

src/api/series-api.ts
```diff
@@ -58,6 +59,36 @@ export class SeriesApi<TSeriesType extends SeriesType> implements ISeriesApi<TSe
 		return this._series.priceScale().coordinateToPrice(coordinate, firstValue.value);
 	}
 
+	public barsInIndexRange(range: TimePointIndexRange): BarsInfo | null {
+		const visibleRange = this._series.model().timeScale().timeRangeForIndexRange(range);
```
timocov, line 63: Actually that's not what we need to have here. `visibleRange` is range for all series currently we have on the chart. To get from/to times, we need to use PlotList (or SeriesData) for this series.

src/api/series-api.ts
```diff
@@ -58,6 +59,36 @@ export class SeriesApi<TSeriesType extends SeriesType> implements ISeriesApi<TSe
...
+
+		const rangeStart = Math.round(range.from) as TimePointIndex;
+		const rangeEnd = Math.round(range.to) as TimePointIndex;
+
+		const bars = this._series.data().bars();
+
+		const firstBar = bars.search(rangeStart);
+		const lastBar = bars.search(rangeEnd);
```
timocov, line 71: Looks like it's exact search, but series might don't have the data for this index. There is different PlotRowSearchMode to looking from right/left instead.

src/api/series-api.ts
```diff
@@ -58,6 +59,36 @@ export class SeriesApi<TSeriesType extends SeriesType> implements ISeriesApi<TSe
...
+		const firstIndex = bars.firstIndex() || 0;
+		const lastIndex = bars.lastIndex() || 0;
+
+		const barsBefore = (null !== firstBar)
+			? firstBar.index - firstIndex
+			: Math.min(lastIndex - firstIndex + 1, rangeStart - firstIndex)
+		;
```
timocov, line 78: shouldn't it be `range.from - firstIndex` or something like that? I just though it might simpler...

timocov:
Hey @gkaindl, sorry for the delay again! I'm back with the news 🙂. We just discussed it with the team and I'd like to make some suggestions according your last comments.

Yeah, we agree that `TimePointIndexRange` not exactly what we can except to have here if we make it floating. Thus, let's rename a bit all stuff. Let's add the new nominal type called `Logical`, which means "floating time point index" and rename all stuff accordingly (getVisibleIndexRange => getVisibleLogicalRange , setVisibleIndexRange => setVisibleLogicalRange, barsInIndexRange => barsInLogicalRange and so on). This looks better and don't violate meanings.

We decide to add the new type because TimePointIndex is index, it's integer value which especially used in data, so we don't want to mix that terms.

That's all we need to change since my previous comment. `barsBefore`, `barsAfter`, `LogicalRange` should be contains floating values, so we can do any stuff to sync several charts (or even apply setVisibleLogicalRange with just got getVisibleLogicalRange value as no-op).

PS: when I said "strict visible range" I meant strict=integer, but it doesn't really make a sense here though 😂.

gkaindl:
Great, that sums up my concern about the "index" types, too! I've refactored everything as you've said, introducing `Logical` and `LogicalRange` at the API level. Internally, there's `LogicalPoint` (better naming suggestions very welcome, but I thought it'd be consistent with `Time` and `TimePoint`). `Logical` basically is the "relaxed" type for the public API, which is either `LogicalPoint` or just `number`, so that API users don't need to type-cast all the time. `LogicalPoint` is the nominal type.

Regarding the other points, I'll respond to them in the change requests!

> PS: when I said "strict visible range" I meant strict=integer, but it doesn't really make a sense here though 😂.

Haha, oh well! Naming truly is the hardest thing after all! 😃

src/model/time-data.ts
gkaindl: Ah, thanks for mentioning this! The relationship between `Time` as a relaxed public type and `TimePoint` as stricter internal type went over my head, I guess! I've undone these changes and moved everything to using `TimePoint` internally.

src/api/iseries-api.ts
gkaindl: Yes, I changed this to `TimePoint`. I think that in the public API, `Time` would be fine too (to be consistent with `TimeRange`, as an example, which also has `from`/`to` and is public), but `TimePoint` is the more explicit type, and potentially easier to work with if your own data uses business days.

src/api/time-scale-api.ts
gkaindl: Good point! I've changed it to `assert()`. It also makes more sense to me, since if you end up calling this with an invalid range, it's probably more useful to be notified of that, rather than silently ignoring it.

src/model/invalidate-mask.ts
gkaindl: Yes, also a good point! I've dropped the time range from the invalidation mask and kept only the logical range (e.g. the more exact one), and I've added a method to `time-scale.ts` to convert a time range to a logical range. That way, we only need to keep one range around, and in the single place where invalidation happened for a time range, I convert the time range to a logical range before, so that the API of `chart-model.ts` isn't changed.

src/model/time-scale.ts
gkaindl: Actually, I do not know 😃 The bars invalidation in the first line of the method is also in current master (and `git blame` traces it back to the initial commit, making it harder to figure out why it’s there) – It was changed in my commit since I replaced the invalidation with a method.

In master, there is a second invalidation after updating the rightOffset, which looked redundant to me, so I removed the second invalidation, but kept the unconditional one before the guard condition – Now I removed the unconditional invalidation instead, and it seems fine to me, so that’s definitely better.

src/model/time-scale.ts
gkaindl: Yes, this got changed in the refactoring anyway.

src/api/series-api.ts
gkaindl: Ah, okay, that was misunderstanding! My train of thought was to provide the currently visible time range (e.g. what's drawn in the time axis of the chart) and how the given series relates to it.

I've changed it now to actually use the time range from the plot list. Also, I've turned `from`/`to` into optional fields here, because if you call with a logical range for which the given series has no data at all, there are also no time points to provide here (as an example, if you call it with the visible logical range when the given series is completely scrolled out of the viewport.

src/api/series-api.ts
gkaindl: Thanks, it's something I've overlooked. I've changed it to search from left/right now at the edges.

src/api/series-api.ts
gkaindl: The idea behind this expression (and the equivalent one for `barsAfter`) is to limit the maximum magnitude to how many bars there actually *are* in the series, e.g. if I take a series with 100 bars and scroll it all the way out to the left, I think that `barsBefore` should never go above 100, since there actually are only 100 bars in the series, so there can't be more bars than that outside of the viewport. Without this limiting expression, the `barsBefore` magnitude would still grow indefinitely.

We could also use the simpler version without the condition and let the `barsBefore`, `barsAfter` grow, but then we would need to document this somehow, because then the meaning between positive and negative values changes subtly, e.g. the way it is now, you can say that “positive values mean that X bars are before the viewport, and negative values mean X bars are missing to fill the viewport” (or simpler: positive values mean this many bars "are there", negative values mean this many bars are missing), but without the condition, we would need to say that “there are bars before the viewport, but these are “logical” bars, not necessarily bars that are actually available in the data).

You can try both behaviors with the debug.html I attached, by looking at the red series (“short series”), and how it behaves as you scroll it in and out of the viewport (e.g. currently, `barsBefore`/`barsAfter` will never get > 600, because there are 600 bars in this series’ data).

timocov:
Hey @gkaindl, thank you so much for this PR.

To avoid the future (I guess it might be confusing) review process I'm going to merge it, but not in master at the first. It'll be branch kind of `fix335`. In this branch we'll make some changes to avoid long review process and to don't boring you with that. After that, we'll make another PR with fixes into that fix335 branch, where we'll mention you if you wish to see changed we made there. After all, we'll merge that all into master and the feature will be part of 3.0 release.

What do you think about that? I just don't want to make the review more painful than it already is and don't disturb you every time. The fixes for the issue looks a bit complex that might be imagined at the first look, so for outside contributors it might be a bit hard. Anyway, you already did really great job with the PR (I must admit that I even didn't expect that from an outside contributor who don't familiar with our codebase) and I don't want to take more your time.

gkaindl:
Thank you very much for your kind words, I'm happy I could contribute a bit! Yes, please mention me when you guys update the branch and finalize it, since I'm very interested to see the changes for issues that I missed an learn more about the codebase – That would also be helpful in case I can contribute something again in the future!

It has been a pleasure to work with you, and I'm looking forward to 3.0 (#209 is another enhancement in that milestone that will be of great help to me, but it already has a PR open, so I suppose there's no contribution needed there anymore).

timocov:
Great! I'm going to merge it and then we'll make some changes there and make a PR with additional fixes. See you soon!

timocov:
Hey @gkaindl, I just created #416 with some fixes.

PR #384, opened 2020-03-30
Refactoring dateFormat

### LINKED ISSUE #368: Refactor date formats
https://github.com/tradingview/lightweight-charts/blob/master/src/formatters/date-format.ts

Here we have a list of available date formats, and this list might be extended every time we want to add new date format. I think we can refactor this code so it will be possible to provide any combination of `yy`/`yyyy`/`MMM`/`mm`/etc within a string and we'll format it in that way (yeah, just use regexes here...).

timocov:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #368
- [x] Includes tests
- [x] Documentation update


timocov:
Needs to add unit-tests for `formatDate` function.

PR #386, opened 2020-04-02
Added ability to customize time scale tick marks formatter

### LINKED ISSUE #226: Time scale marks customization
Currently there is no support to customise the time scale with even at least formatting.

I have two problems:

- My trade data is now showing random intervals of 12:32 and 13:34 rather than just 12:30 and 13:30 
- There doesn't seem to be a way to localise the timescale and convert unix timestamp to local times

Any chance we can get this soon?

timocov:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #226
- [x] Includes tests
- [x] Documentation update

### DIFF
--- a/docs/time-scale.md
+++ b/docs/time-scale.md
@@ -16,24 +16,24 @@ Time scale (or time axis) is a horizontal scale at the bottom of the chart that 
 |`visible`|`boolean`|`true`|If true, the time scale is shown on a chart|
 |`timeVisible`|`boolean`|`false`|If true, the time is shown on the time scale and in the vertical crosshair label|
 |`secondsVisible`|`boolean`|`true`|If true, seconds are shown on the label of the crosshair vertical line in `hh:mm:ss` format on intraday intervals|
-|`tickMarkFormatter`|`(TimePoint, TickMarkType, locale) => string`|Default tick marks formatter|Allows override tick marks formatter (see below)|
+|`tickMarkFormatter`|`(TimePoint, TickMarkType, locale) => string`|Default tick marks formatter|Allows to override the tick marks formatter (see below)|
 
 ### Tick marks formatter
 
-Tick marks formatter might be used to customize tick marks labels on time axis.
+Tick marks formatter can be used to customize tick marks labels on the time axis.
 
-To customize it, you need to provide `tickMarkFormatter` option. It's the function with the following declaration:
+To customize it, you need to provide the `tickMarkFormatter` option. It's a function with the following declaration:
 
 ```typescript
 export type TickMarkFormatter = (timePoint: TimePoint, tickMarkType: TickMarkType, locale: string) => string;
 ```
 
-Where `timePoint` is [Time](./time.md) object, `type` is [TickMarkType](./constants.md#TickMarkType) enum and `locale` is a string with current applied locale.
+Where `timePoint` is [Time](./time.md) object, `type` is [TickMarkType](./constants.md#TickMarkType) enum and `locale` is the currently applied locale of the string type.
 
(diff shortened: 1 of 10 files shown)


docs/time-scale.md
```diff
@@ -16,6 +16,24 @@ Time scale (or time axis) is a horizontal scale at the bottom of the chart that
...
+
+To customize it, you need to provide `tickMarkFormatter` option. It's the function with the following declaration:
+
+```typescript
+export type TickMarkFormatter = (timePoint: TimePoint, tickMarkType: TickMarkType, locale: string) => string;
+```
+
+Where `timePoint` is [Time](./time.md) object, `type` is [TickMarkType](./constants.md#TickMarkType) enum and `locale` is a string with current applied locale.
+
+This function should return a string represents `timePoint` as `tickMarkType` type (year, month, etc) in `locale` locale.
```
ezhukovskiy, line 33:
```suggestion
This function should return `timePoint` as a string formatted according to `tickMarkType` type (year, month, etc) and `locale`.
```

docs/time-scale.md
```diff
@@ -16,6 +16,24 @@ Time scale (or time axis) is a horizontal scale at the bottom of the chart that
...
+
+Note that the returned string should be shortest value as possible and contains no more than (approx.) 8 characters.
```
ezhukovskiy, line 35:
```suggestion
Note that the returned string should be the shortest possible value and should have no more than 8 characters.
```

docs/time-scale.md
```diff
@@ -16,6 +16,24 @@ Time scale (or time axis) is a horizontal scale at the bottom of the chart that
...
+Otherwise, tick marks could be overlap each other.
```
ezhukovskiy, line 36:
```suggestion
Otherwise, the tick marks will overlap each other.
```

src/api/options/time-scale-options-defaults.ts
```diff
@@ -11,4 +14,55 @@ export const timeScaleOptionsDefaults: TimeScaleOptions = {
 	visible: true,
 	timeVisible: false,
 	secondsVisible: true,
+	tickMarkFormatter: (timePoint: TimePoint, tickMarkType: TickMarkType, locale: string) => {
```
kirchet, line 17: Maybe it would be better if we moved this default implementations to a separated named function, maybe even in another file

src/model/formatted-labels-cache.ts
```diff
@@ -36,8 +42,8 @@ export class FormattedLabelsCache {
 		}
 
 		const str = this._format(value);
-		this._cache.set(value.valueOf(), { string: str, tick: this._usageTick });
-		this._tick2Labels.set(this._usageTick, value.valueOf());
+		this._cache.set(cacheKey, { string: str, tick: this._usageTick });
+		this._tick2Labels.set(this._usageTick, cacheKey);
```
kirchet, line 46: Are we checking somewhere that this cache is invalidated when `tickMarkFormatter` is changed?

src/api/options/time-scale-options-defaults.ts
timocov: It doesn't really make a sense (from my prospective), but if you wish I can move it out 👌

src/model/formatted-labels-cache.ts
timocov: Yes, in time scale we clear all caches when new options are applied.

src/api/options/time-scale-options-defaults.ts
kirchet: OK, let's do it 🤝

timocov: Done in 0c6b6bd

PR #387, opened 2020-04-02
Improved tick marks generation algorithm for the first point

timocov:
**Type of PR:** enhancement

**PR checklist:**

- [ ] Addresses an existing issue:
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Span (weight) for the first tick mark previously was always 20, but now it's "smarty" and calculated as spanByTime of the first time point and average time point's diff back.

### DIFF
--- a/src/api/data-layer.ts
+++ b/src/api/data-layer.ts
@@ -175,11 +175,11 @@ const spanDivisors = [
 	},
 ];
 
-function spanByTime(time: TimePoint, previousTime: TimePoint | null): number {
+function spanByTime(time: UTCTimestamp, previousTime: UTCTimestamp | null): number {
 	// function days(count) { return count * 24 * 60 * 60 * 1000; }
 	if (previousTime !== null) {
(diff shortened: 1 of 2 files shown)


src/api/data-layer.ts
```diff
@@ -392,17 +392,27 @@ export class DataLayer {
...
 				span: span,
 				time: time,
 				index: index as TimePointIndex,
 			};
 		});
 
+		if (marks.length > 1) {
+			// let's guess a span for the first mark
+			// let's say the previous point was average time back in the history
+			const averageTimeDiff = Math.ceil(totalTimeDiff / (marks.length - 1));
```
kirchet, line 411: I'm not sure about this algorithm. That's may be not correct if we take into account that sessions can be like 10:00-16:00, bit I cannot suggest a better _simple_ solution, so let's try this solution and if we'll face some issues we will improve this algorithm

timocov: Yeah, nobody's sure here though 😂

PR #290, opened 2019-12-20
Two price scales

eugene-korobko:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #129
- [x] Includes tests
- [x] Documentation update

**Overview of change:**
Now it is possible to show both left and rigth price axises.

PR #394, opened 2020-04-07
378: Add an option to hide ticks on price axis labels

### LINKED ISSUE #378: Remove tick mark from price label
Currently a price label has a line in front of the text:
image

Please add an option to remove this line. For someone it can look like a minus.

eugene-korobko:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #378 
- [x] Includes tests
- [x] Documentation update

**Overview of change:**
Added an option to hide small horizontal lines on price axis labels


src/renderers/price-axis-renderer-options-provider.ts
```diff
@@ -51,6 +52,7 @@ export class PriceAxisRendererOptionsProvider {
 		}
 
 		rendererOptions.color = this._textColor();
+		rendererOptions.drawTicks = this._chartModel.options().priceScale.drawTicks;
```
timocov, line 55: In context of #290 - which one? Is it possible to draw tick marks on one scale and don't on another one?

src/renderers/price-axis-view-renderer.ts
```diff
@@ -118,7 +118,7 @@ export class PriceAxisViewRenderer implements IPriceAxisViewRenderer {
 			ctx.fillStyle = this._data.borderColor;
 			ctx.fillRect(alignRight ? rightScaled - horzBorderScaled : 0, yTopScaled, horzBorderScaled, yBottomScaled - yTopScaled);
 
-			if (this._data.tickVisible) {
+			if (this._data.tickVisible && rendererOptions.drawTicks) {
```
timocov, line 121: why can't we use only `tickVisible` here and patch `tickVisible` from pane view?

eugene-korobko: Hm, it looks more logical for me. Some labels have ticks, some do not, however drawTicks is a global options how to draw labels

timocov: So, why renderer should know about it? I don't think that it's good idea to have several different sources of truth at least in the last mile, in renderers.

src/views/price-axis/price-axis-view.ts
```diff
@@ -78,46 +81,49 @@ export abstract class PriceAxisView implements IPriceAxisView {
...
 		paneRendererData: PriceAxisViewRendererData,
 		commonData: PriceAxisViewRendererCommonData
 	): void;
 
-	private _updateRendererDataIfNeeded(): void {
+	private _updateRendererDataIfNeeded(priceScale: PriceScale): void {
 		if (this._invalidated) {
 			this._updateRendererData(this._axisRendererData, this._paneRendererData, this._commonRendererData);
-			this._invalidated = false;
+			this._axisRendererData.tickVisible = this._axisRendererData.tickVisible && priceScale.options().drawTicks;
```
timocov, line 125: Looks like there is a bug - if I apply options with disabling `drawTicks` and then enable it, it stays invisible:

```
chart.applyOptions({ rightPriceScale: { drawTicks: false } })
// ...
chart.applyOptions({ rightPriceScale: { drawTicks: true } })
```

src/views/price-axis/price-axis-view.ts
```diff
@@ -78,46 +81,49 @@ export abstract class PriceAxisView implements IPriceAxisView {
...
+		this._updateRendererDataIfNeeded(priceScale);
 		this._axisRenderer.setData(this._axisRendererData, this._commonRendererData);
 		this._paneRenderer.setData(this._paneRendererData, this._commonRendererData);
 
 		return this._axisRenderer;
 	}
 
 	public paneRenderer(): IPriceAxisViewRenderer {
-		this._updateRendererDataIfNeeded();
+		this._updateRendererDataIfNeeded(ensureNotNull(this._getSource().priceScale()));
```
timocov, line 107: `_getSource()` in case of cross-hair never has price scale so we'll fail here somewhen. We don't fail right now, but we could. It might be need to update `tickVisible` only in case of requesting renderer, but I'm not sure.

src/views/price-axis/price-axis-view.ts
eugene-korobko: Fixed

timocov: Could you please add a test case for that?

PR #399, opened 2020-04-09
Separate handling of missing/zero width/height in chart-widget constr…

### LINKED ISSUE #353: Impossible to override the only width or height in constructor
**Lightweight Charts Version:** latest

**Steps/code to reproduce:**

```javascript
	var chart = LightweightCharts.createChart(container, {
		width: 1000,
	});
```

**Actual behavior:**

Chart's height is 50.

**Expected behavior:**

Chart's height is container's height.

The problem is in somewhere here https://github.com/tradingview/lightweight-charts/blob/71d096804a79243172c451c142fc696d4bb33768/src/gui/chart-widget.ts#L78-L95.

gkaindl:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #353 
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

I noticed this issue when browsing the `3.0` milestone and saw it had the `help-wanted` tag.

Width/height were only adjusted to the container size if both are 0 or missing, but not if only one of them is missing. This fixes it.

I think it would make sense to move the highDPI-fix to the unconditional path or mention it in the docs that width/height should be chosen so that `(width-or-height * devicePixelRatio) % 2 === 0`, or that could be a gotcha for people specifying their own width/height.

If the intention of not doing so is to ensure the resulting DOM element has exactly the requested width/height, we could add the potentially lost pixel as padding on the right or bottom (where the price scale and time scale are), so that the enclosing element still has exactly the same width/height as specified (that might be important to people doing absolute positioning to build an interface, as an example, so they don't get missing single pixel lines that they can't explain).

### DIFF
--- a/src/gui/chart-widget.ts
+++ b/src/gui/chart-widget.ts
@@ -75,19 +75,20 @@ export class ChartWidget implements IDestroyable {
 		let width = this._options.width;
 		let height = this._options.height;
 
-		if (width === 0 && height === 0) {
+		if (width === 0 || height === 0) {
 			const containerRect = container.getBoundingClientRect();
 			// TODO: Fix it better
 			// on Hi-DPI CSS size * Device Pixel Ratio should be integer to avoid smoothing
 			// For chart widget we decreases because we must be inside container.
 			// For time axis this is not important, since it just affects space for pane widgets
-			width = Math.floor(containerRect.width);
-			if (width % 2) {
-				width -= 1;
+			if (width === 0) {
+				width = Math.floor(containerRect.width);
+				width -= width % 2;
 			}
-			height = Math.floor(containerRect.height);
-			if (height % 2) {
-				height -= 1;
+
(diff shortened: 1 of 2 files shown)


timocov:
> If the intention of not doing so is to ensure the resulting DOM element has exactly the requested width/height, we could add the potentially lost pixel as padding on the right or bottom (where the price scale and time scale are), so that the enclosing element still has exactly the same width/height as specified (that might be important to people doing absolute positioning to build an interface, as an example, so they don't get missing single pixel lines that they can't explain).

@eugene-korobko that's for you 😂

gkaindl:
Sure! I've added a graphics test-case that sub-divides the container vertically to try all 4 combinations of width/height being set or not, creating a test image like this:

2 test

timocov:
Awesome! Thank you!

PR #410, opened 2020-04-21
392 price range api

### LINKED ISSUE #392: Add ability to modify price scale's price range

eugene-korobko:
**Type of PR:** enhancement

**PR checklist:**

- [x] Includes tests
- [x] Documentation update

**Overview of change:**
Now there is am ability to set certain price range via PriceScaleApi

### DIFF
--- a/docs/price-scale.md
+++ b/docs/price-scale.md
@@ -9,8 +9,30 @@ Left and right price scales could be visible, overlay price scales are always hi
 ## Autoscale
 
 Autoscaling is a feature of automatic adjusting price scale to fit the visible range of data.
-Autoscaling is enabled by default, however you could turn it off by zooming price scale or calling `PriceScaleApi.setOptions` method with `autoScale` field set to false.
+Autoscaling is enabled by default, however you could turn it off by zooming price scale or calling `PriceScaleApi.applyOptions` method with `autoScale` field set to false.
 Overlay price scales are always autoscaled.
+
+## Price Range
+
+`priceScaleApi.applyOptions` method accepts a structure with `priceRange` fields. It allows overriding claculated price range with predefined one. So every time the library will calculate visible price scale range, the specified range will be returned.
+For instance,
+
+```javascript
+    var chart = LightweightCharts.createChart(container, {
+        priceScale: {
+            priceRange: {
+                minValue: 0,
+                maxValue: 100,
+            },
+            scaleMargins: {
+                bottom: 0,
+                top: 0,
+            },
+        },
+    });
+```
+
+Creates a chart with price scale range set to 0..100.
 
 ## PriceScale ID
 
--- a/src/api/price-scale-api.ts
+++ b/src/api/price-scale-api.ts
@@ -3,7 +3,7 @@ import { IDestroyable } from '../helpers/idestroyable';
 import { DeepPartial } from '../helpers/strict-type-checks';
 
 import { ChartModel } from '../model/chart-model';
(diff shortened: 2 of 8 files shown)


src/api/price-scale-api.ts
```diff
@@ -24,11 +24,11 @@ export class PriceScaleApi implements IPriceScaleApi, IDestroyable {
 		return this._priceScale().id();
 	}
 
-	public applyOptions(options: DeepPartial<PriceScaleOptions>): void {
+	public applyOptions(options: DeepPartial<Exclude<PriceScaleOptions, 'autoScale'>> & Partial<PriceScaleRangeOptions>): void {
```
timocov, line 27: Should we:

1. Update iprice-scale-api as well
1. Provide method to enable autoScale?

Maybe it'd better to extract this complex type into separate type

docs/price-scale.md
```diff
@@ -9,9 +9,31 @@ Left and right price scales could be visible, overlay price scales are always hi
 ## Autoscale
 
 Autoscaling is a feature of automatic adjusting price scale to fit the visible range of data.
-Autoscaling is enabled by default, however you could turn it off by zooming price scale or calling `PriceScaleApi.setOptions` method with `autoScale` field set to false.
+Autoscaling is enabled by default, however you could turn it off by zooming price scale or calling `PriceScaleApi.applyOptions` method with `autoScale` field set to false.
```
timocov, line 12: How to enable autoScale back?

src/api/price-scale-api.ts
timocov: Also why type in applyOptions and options are differ?

src/model/price-scale.ts
```diff
@@ -160,18 +172,29 @@ export class PriceScale {
 		return this._id;
 	}
 
-	public options(): Readonly<PriceScaleOptions> {
-		return this._options;
+	public options(): Readonly<PriceScaleOptions & PriceScaleRangeOptions> {
+		const priceScaleMinMax = this._priceRange === null ? null : {
```
timocov, line 176: Does it mean that if we'll request options when range was invalidated we'll get `null` as range?

src/model/price-scale.ts
```diff
@@ -851,12 +874,17 @@ export class PriceScale {
 
 	// tslint:disable-next-line:cyclomatic-complexity
 	private _recalculatePriceRangeImpl(): void {
+		if (this._predefinedPriceRange !== null) {
+			this.setPriceRange(this._predefinedPriceRange);
```
timocov, line 878: I don't think that it's good place to have it here. Afaik this method is used to autoscale only. Why we should have it here? Why we can't set price range right inside applyOptions method?

src/api/price-scale-api.ts
eugene-korobko: 1. Done
2. Actually, setting priceRange does not disable autoscale, it sets predefined result of autoscale.
3. Applying options with priceRange = null restores autoscaling base on visible data
4. Options differ because we could not define price range, however lt always exist. Also we cannot turn on/off autoscale directly, however we can get this value

timocov: > Actually, setting priceRange does not disable autoscale, it sets predefined result of autoscale.

So, what happened if an user manually change the price range by dragging the price axis after "setting predefined autoscale"? I don't even think that this is what we actually need.

> Options differ because we could not define price range, however lt always exist. Also we cannot turn on/off autoscale directly, however we can get this value

I'm not sure that this is what we actually need to have here.

PR #413, opened 2020-04-22
Override autoscale

### LINKED ISSUE #382: Create Grid instance for every pane
For now we use the only one Grid instance per the whole chart, but create pane view for every pane, what led by [using WeakMap to avoid memory leaks](https://github.com/tradingview/lightweight-charts/blob/5e328f4982675c78d8a7599a240a7d108cc84034/src/model/grid.ts#L26-L31). If we'll create Grid instance per pane, we don't need to use WeakMap, so -1 required polyfill 🙂.

### LINKED ISSUE #392: Add ability to modify price scale's price range

eugene-korobko:
**Type of PR:** enhancement

**PR checklist:**

- [x] Includes tests
- [x] Documentation update

**Overview of change:**
Now there is an option to override series autoscale value.


docs/series-basics.md
```diff
@@ -118,6 +118,62 @@ chart.applyOptions({
...
+            priceRange: {
+                minValue: 0,
+                maxValue: 100,
+            },
+        };
+    },
+});
+```
+
+So, you could just add a function that returns object with `priceRange` field to make chart thinking the series always has values from 0 to 100. These values aer in prices.
```
timocov, line 139:
```suggestion
So, you could just add a function that returns object with `priceRange` field to make chart thinking the series always has values from 0 to 100. These values are in prices.
```

docs/series-basics.md
```diff
@@ -118,6 +118,62 @@ chart.applyOptions({
...
+            margins: {
+                above: 10,
+                below: 10,
+            },
+        };
+    },
+});
+```
+
+Actually, `autoscaleProvider` function accepts a parameter with original implementation, so you can call default and ajust the result:
```
timocov, line 160:
```suggestion
Actually, `autoscaleProvider` function accepts a parameter with original implementation, so you can call default and adjust the result:
```

docs/series-basics.md
```diff
@@ -118,6 +118,62 @@ chart.applyOptions({
...
+        if (res.priceRange != null) {
+            res.priceRange.minValue -= 10;
+            res.priceRange.maxValue += 10;
+        }
+        return res;
+    },
+});
+```
+
+Please pay your attention that both `priceRange` and `margins` could be null in the default result.
```
timocov, line 175:
```suggestion
Please pay your attention that both `priceRange` and `margins` could be `null` in the default result.
```

docs/series-basics.md
```diff
@@ -118,6 +118,62 @@ chart.applyOptions({
...
+});
+```
+
+Actually, `autoscaleProvider` function accepts a parameter with original implementation, so you can call default and ajust the result:
+
+```javascript
+var firstSeries = chart.addLineSeries({
+    autoscaleProvider: (original) => {
+        var res = original();
+        if (res.priceRange != null) {
```
timocov, line 166:
```suggestion
        if (res.priceRange !== null) {
```

src/model/series-options.ts
```diff
@@ -163,6 +164,20 @@ export interface NonOverlaySeriesSpecificOptions {
...
+}
+
+export interface AutoscaleInfo {
+	priceRange: PriceRange | null;
+	margins: AutoScaleMargins | null;
+}
+
+type AutoscaleProviderImplementation = () => AutoscaleInfo | null;
+
+type AutoscaleProvider = (baseImplementation: AutoscaleProviderImplementation) => AutoscaleInfo | null;
```
timocov, line 179: AutoscaleInfoProvider?

src/model/series-options.ts
```diff
@@ -163,6 +164,20 @@ export interface NonOverlaySeriesSpecificOptions {
...
+	minValue: number;
+	maxValue: number;
+}
+
+export interface AutoscaleInfo {
+	priceRange: PriceRange | null;
+	margins: AutoScaleMargins | null;
+}
+
+type AutoscaleProviderImplementation = () => AutoscaleInfo | null;
```
timocov, line 177: Maybe reduce that type and just inline it in AutoscaleProvider? It's name is a bit strange :)

src/model/series.ts
```diff
@@ -395,33 +395,16 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
...
-		let barsMinMax: MinMax | null;
-		if (priceSource !== null) {
-			barsMinMax = this.data().bars().minMaxOnRangeCached(startTimePoint, endTimePoint, [{ name: priceSource, offset: 0 }]);
-		} else {
-			barsMinMax = this.data().bars().minMaxOnRangeCached(startTimePoint, endTimePoint, [{ name: 'low', offset: 0 }, { name: 'high', offset: 0 }]);
-		}
-
-		let range = barsMinMax !== null ? new PriceRange(barsMinMax.min, barsMinMax.max) : null;
+	public autoscaleInfo(startTimePoint: TimePointIndex, endTimePoint: TimePointIndex): AutoscaleInfoImpl | null {
+		if (this._options.autoscaleProvider) {
```
timocov, line 399:
```suggestion
		if (this._options.autoscaleProvider !== undefined) {
```

src/model/series-options.ts
eugene-korobko: ok

src/model/series-options.ts
eugene-korobko: done

docs/series-basics.md
```diff
@@ -118,6 +118,62 @@ chart.applyOptions({
...
 ```
 
+### Overriding autoscale
+
+By default, chart scales data automatically, based on visible range of data. However for some reasons one could need overriding this behavior.
+There is an options called `autoscaleProvider` that allows overriding visible price range for series
+
+```javascript
+var firstSeries = chart.addLineSeries({
+    autoscaleProvider: () => {
```
timocov, line 128:
```suggestion
    autoscaleInfoProvider: () => {
```

docs/series-basics.md
```diff
@@ -118,6 +118,62 @@ chart.applyOptions({
...
+});
+```
+
+So, you could just add a function that returns object with `priceRange` field to make chart thinking the series always has values from 0 to 100. These values are in prices.
+
+You could also provide additional margins in pixels. Please be careful and never mix prices and pixels.
+
+```javascript
+var firstSeries = chart.addLineSeries({
+    autoscaleProvider: () => {
```
timocov, line 145:
```suggestion
    autoscaleInfoProvider: () => {
```

docs/series-basics.md
```diff
@@ -118,6 +118,62 @@ chart.applyOptions({
...
+            margins: {
+                above: 10,
+                below: 10,
+            },
+        };
+    },
+});
+```
+
+Actually, `autoscaleProvider` function accepts a parameter with original implementation, so you can call default and adjust the result:
```
timocov, line 160:
```suggestion
Actually, `autoscaleInfoProvider` function accepts a parameter with original implementation, so you can call default and adjust the result:
```

docs/series-basics.md
```diff
@@ -118,6 +118,62 @@ chart.applyOptions({
...
+
+```javascript
+var firstSeries = chart.addLineSeries({
+    autoscaleProvider: (original) => {
```
timocov, line 164:
```suggestion
    autoscaleInfoProvider: (original) => {
```

PR #416, opened 2020-04-23
Improves for #370

timocov:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: #335

**Overview of change:**

It's fixes atop of fix335 branch by @gkaindl.

The fixes contains:

- Removing some unused methods/types
- Moving types between projects for architectural purposes
- Changing to from `TimePoint` to `Time` in `BarsInfo` (after some overview of typing I noticed that we mostly use `Time` and it's better for API consistency)
- Small refactoring of `barsInLogicalRange`
- Having the only 1 "source of truth" in TimeScale in case of visible range (previously we had 2 independent visible bars and visible logical range)

As we previously discussed in #370, I just made some changed in @gkaindl's fixes and make this PR to overview them (the PR is NOT in master, it's in fix335 branch).


timocov:
According the tests there are some visible range changes which we need to overview and quite possible fix the bugs, but I don't think that there will be big changes.

tests/unittests/timed-data.spec.ts
```diff
@@ -15,7 +15,7 @@ function visibleTimedValuesCase(
 	expectedTo: number,
 	times: number[]
 ): void {
-	const barsRange = new BarsRange(rangeFrom as TimePointIndex, rangeTo as TimePointIndex);
+	const barsRange = new Range(rangeFrom as TimePointIndex, rangeTo as TimePointIndex);
```
kirchet, line 18: `Range<TimePointIndex>`?

src/model/time-scale.ts
```diff
@@ -765,22 +732,15 @@ export class TimeScale {
...
 
-	private _setVisibleBars(visibleBars: BarsRange | null): void {
-		if (visibleBars === null && this._visibleBars === null) {
-			return;
-		}
-
-		const oldVisibleBars = this._visibleBars;
-		this._visibleBars = visibleBars;
+	private _setVisibleRange(newVisibleRange: TimeScaleVisibleRange): void {
+		const oldVisibleBars = this._visibleRange;
```
kirchet, line 736: Minor: `oldVisibleRange`

tests/unittests/timed-data.spec.ts
kirchet, line 18: `Range<TimePointIndex>`?

src/model/time-scale.ts
```diff
@@ -573,30 +575,11 @@ export class TimeScale {
...
-			ensureNotNull(points.indexOf(Math.max(firstPoint, range.from.timestamp) as UTCTimestamp, true)),
-			ensureNotNull(points.indexOf(Math.min(lastPoint, range.to.timestamp) as UTCTimestamp, true))
-		);
-		this.setVisibleRange(barRange);
-	}
-
-	public setLogicalIndexRange(range: LogicalPointRange): void {
-		const barRange = new BarsRange(
+	public setLogicalRange(range: LogicalRange): void {
+		const barRange = new Range(
```
kirchet, line 582: Maybe we should add `assert(range.isValid())` here

kirchet, line 582: `new Range<TimePointIndex>`

src/model/time-scale.ts
```diff
@@ -148,39 +149,40 @@ export class TimeScale {
...
 	}
 
-	public visibleBars(): BarsRange | null {
-		if (this._visibleBarsInvalidated) {
-			this._visibleBarsInvalidated = false;
-			this._updateVisibleBars();
+	// strict range: integer indices of the bars in the visible range rounded in more wide direction
+	public visibleStrictRange(): Range<TimePointIndex> | null {
+		if (this._visibleRangeInvalidated) {
+			this._visibleRangeInvalidated = false;
```
kirchet, line 155: Maybe we should move `this._visibleRangeInvalidated = false;` to `_updateVisibleRange` to prevent copy-paste

tests/unittests/timed-data.spec.ts
timocov: It's the same. TypeScript understand that type correctly :)

image

src/model/time-scale.ts
timocov: https://github.com/tradingview/lightweight-charts/pull/416#discussion_r413987818

src/model/time-scale.ts
timocov: Agree.

src/model/time-scale.ts
timocov: range is simple object with from/to values. `isValid()` is in TimeScaleVisibleRange only

timocov:
Needs to add docs as well for new methods.

tests/unittests/timed-data.spec.ts
kirchet: Yes, I understand this, this is just a code style question. You provide `<T>` in some cases, but don't do the same in anothers

timocov: Ok, I'll remove `<Logical>` in `_updateVisibleRange` from `new Range<Logical>` - it looks exceeded and the only one place where I added it where it could be stripped :) In all other places it's types, so we can't don't use specifier for generic

timocov:
Done.

docs/time-scale.md
```diff
@@ -141,3 +161,76 @@ Returns an object with options currently applied to the time scale.
...
+
+Integer part of logical index means index of the fully visible bar (see image below).
+Thus, if we have 5.2 as the last visible logical index (`to` field), that means that the last visible bar is bar with index 5, but also we have partially visible (for 20%) 6th bar.
+Half (e.g. 1.5, 3.5, 10.5) means exactly middle of the bar.
+
+![Logical range](./assets/logical-range.png "Logical range")
+
+Red vertical lines here are borders between bars.
+
+So, in the image above visible logical range is from -4.73 and to 5.05.
```
kirchet, line 236: @ezhukovskiy look at docs please

timocov:
Hey @gkaindl, did you have a time to look at this PR? Maybe you have some questions/suggestions?

docs/series-basics.md
```diff
@@ -403,6 +403,44 @@ const priceLine = series.createPriceLine({ price: 80.0 });
...
+
+The returned value is an object with the following properties:
+
+- `from` - a [Time](./time.md) of the first series' bar inside passed logical range or `undefined` if no bars found in requested range
+- `to` - a [Time](./time.md) of the last series' bar inside passed logical range or `undefined` if no bars found in requested range
+- `barsBefore` - a number of bars between `from` index of passed logical range and the first series' bar
+- `barsAfter` - a number of bars between `to` index of passed logical range and the last series' bar
+
+Positive value in `barsBefore` field means that there are some bars before (out of logical range from the left) the `from` logical index in the series.
+Negative value means that the first series' bar is inside the passed logical range, and between the first series' bar and the `from` logical index are some bars.
```
gkaindl, line 418: I think the behavior of negative values isn't really described correctly here: If the value is negative, it does not necessarily mean that the first bar in the series is within the logical range. It can very well be outside the logical range "to the right". In that case, the magnitude of the negative value will just increase as you keep scrolling to the left.

Maybe something like this would be more correct and clearer (Could maybe be rephrased to be even more succinct, but it's the best I can do, I think!)

"Negative value means that the first series bar is either inside the logical range or outside of it on the right. The magnitude of the negative value is equivalent to the amount of bar data required to completely fill the viewport from the left up to the first current series bar."

docs/series-basics.md
```diff
@@ -403,6 +403,44 @@ const priceLine = series.createPriceLine({ price: 80.0 });
...
+
+Positive value in `barsAfter` field means that there are some bars after (out of logical range from the right) the `to` logical index in the series.
+Negative value means that the last series' bar is inside the passed logical range, and between the last series' bar and the `to` logical index are some bars.
```
gkaindl, line 421: Same as above, only with left/right exchanged.

docs/time-scale.md
```diff
@@ -141,3 +161,76 @@ Returns an object with options currently applied to the time scale.
...
+```
+
+## Logical range
+
+Logical range is an object with 2 properties: `from` and `to`, which are numbers and represent logical indexes on the time scale.
+
+The start point of the time scale's logical range is the data item of all series.
+Before that point all indexes are negative, starts from the point - positive.
+
+Indexes might be floating, for instance 4.2.
```
gkaindl, line 226: I think "Indexes might have fractional parts, for instance 4.2, due to the time-scale being continuous rather than discrete" might be clearer here.

src/api/series-api.ts
```diff
@@ -61,33 +63,44 @@ export class SeriesApi<TSeriesType extends SeriesType> implements ISeriesApi<TSe
...
+		}
+
+		const dataFirstBarInRange = bars.search(correctedRange.left(), PlotRowSearchMode.NearestRight);
+		const dataLastBarInRange = bars.search(correctedRange.right(), PlotRowSearchMode.NearestLeft);
+
+		const dataFirstIndex = ensureNotNull(bars.firstIndex());
+		const dataLastIndex = ensureNotNull(bars.lastIndex());
+
+		const barsBefore = (dataFirstBarInRange === null || dataFirstBarInRange.index === dataFirstIndex)
+			? range.from - dataFirstIndex
```
gkaindl, line 88: Makes sense to take out the limiting to the amount of bar data items in the series here – On second thought, that didn't really serve a useful purpose, and with the description of this method in the docs now, this behavior makes perfect sense!

src/model/time-data.ts
```diff
@@ -17,19 +17,18 @@ export interface TimePoint {
 	businessDay?: BusinessDay;
 }
 
-export interface TimePointsRange {
-	from: TimePoint;
-	to: TimePoint;
+export interface Range<T> {
```
gkaindl, line 20: Great addition (as well as `RangeImpl`)! – Much nicer to have these as generics, rather than repeating the shape!

src/model/time-scale.ts
```diff
@@ -765,22 +730,15 @@ export class TimeScale {
...
-		if (this._visibleBars === null || oldVisibleBars !== null && !this._visibleBars.equals(oldVisibleBars)) {
+		if (!areRangesEqual(oldVisibleRange.strictRange(), this._visibleRange.strictRange())) {
 			this._visibleBarsChanged.fire();
 		}
 
-		const logicalRange = this.visibleLogicalRange();
-
-		if (logicalRange?.from !== this._prevVisibleLogicalRange?.from || logicalRange?.to !== this._prevVisibleLogicalRange?.to) {
-			this._prevVisibleLogicalRange = logicalRange;
+		if (!areRangesEqual(oldVisibleRange.logicalRange(), this._visibleRange.logicalRange())) {
```
gkaindl, line 741: That's also a great improvement, not having to carry around the extra `_prevVisibleLogicalRange` (which was bothering me) by creating a single "unified" range instead!

gkaindl:
Thanks for pinging me! (and sorry for my late response!)

The improvements are great and make it much cleaner, especially unifying both range types into a single objects in the time-scale, which is more robust and also easier to understand, I think!

I've left a few comments regarding the documentation for the new functionality (especially the `barsInIndexRange()` method), which I think reflect the behavior more accurately.

I've also tested your changes with the `debug.html` from my original pull request, as well as with the current state of the app I'm building, which previously used my branch – It's working well, and I'm using some rather complex chart setups there now with multiple series per chart, which each load additional data individually, and it's all working well, so that's a great sign.

Regarding the failing tests with the visible range, yes, sorry, this was most likely because I was using `round()` instead of `floor()/ceil()` to convert the ranges, but I saw that you fixed this right away :-)

Also, when 3.0 comes out, I think it would be great to eventually add the new ability to dynamically load new data during scrolling to the demo page at https://en.tradingview.com/lightweight-charts/ too – From my perspective as a user, that was the first thing I was looking at to check out what the library can do, and since there was no example about dynamic loading, that's what made me check the issues and eventually pick up the task to add it – Maybe just something where when `barsBefore/barsAfter` turn negative, it add prepends/appends more randomly generated data after a 1 second timeout to "simulate" the loading?

docs/time-scale.md
```diff
@@ -141,3 +161,76 @@ Returns an object with options currently applied to the time scale.
...
+```
+
+### subscribeVisibleLogicalRangeChange
+
+Allows you subscribe on visible [logical range](#logical-range) changed events.
+The argument is a handler function, which will be called with the new visible logical range.
+
+The argument passed to your handler is the new visible logical range, which might be either:
+
+- an object with properties `from` and `to` of type number (see [Logical range](#logical-range) section) which means the first and last visible logical index accordingly
```
ezhukovskiy, line 199:
```suggestion
- an object with numerical properties `from` and `to` of type number (see [Logical range](#logical-range) section), that are the first and last visible logical indexes, respectively
```

docs/time-scale.md
```diff
@@ -141,3 +161,76 @@ Returns an object with options currently applied to the time scale.
...
+function onVisibleLogicalRangeChanged(newVisibleLogicalRange) {
+    console.log(newVisibleLogicalRange);
+}
+
+chart.timeScale().subscribeVisibleLogicalRangeChange(onVisibleLogicalRangeChanged);
+```
+
+### unsubscribeVisibleLogicalRangeChange
+
+Allows you unsubscribe from previously subscribed handler on visible logical range changed events.
```
ezhukovskiy, line 212:
```suggestion
Allows to unsubscribe the previously subscribed handler from the visible logical range changed event.
```

docs/time-scale.md
```diff
@@ -141,3 +161,76 @@ Returns an object with options currently applied to the time scale.
...
+```javascript
+chart.timeScale().unsubscribeVisibleLogicalRangeChange(onVisibleLogicalRangeChanged);
+```
+
+## Logical range
+
+Logical range is an object with 2 properties: `from` and `to`, which are numbers and represent logical indexes on the time scale.
+
+The start point of the time scale's logical range is the data item of all series.
+Before that point all indexes are negative, starts from the point - positive.
```
ezhukovskiy, line 224:
```suggestion
Before that point all indexes are negative, starting from that point - positive.
```

docs/time-scale.md
```diff
@@ -141,3 +161,76 @@ Returns an object with options currently applied to the time scale.
...
+
+```javascript
+chart.timeScale().unsubscribeVisibleLogicalRangeChange(onVisibleLogicalRangeChanged);
+```
+
+## Logical range
+
+Logical range is an object with 2 properties: `from` and `to`, which are numbers and represent logical indexes on the time scale.
+
+The start point of the time scale's logical range is the data item of all series.
```
ezhukovskiy, line 223:
```suggestion
The starting point of the time scale's logical range is the first data item among all series.
```

docs/time-scale.md
```diff
@@ -141,3 +161,76 @@ Returns an object with options currently applied to the time scale.
...
+Before that point all indexes are negative, starts from the point - positive.
+
+Indexes might be floating, for instance 4.2.
+
+Integer part of logical index means index of the fully visible bar (see image below).
+Thus, if we have 5.2 as the last visible logical index (`to` field), that means that the last visible bar is bar with index 5, but also we have partially visible (for 20%) 6th bar.
```
ezhukovskiy, line 229:
```suggestion
Thus, if we have 5.2 as the last visible logical index (`to` field), that means that the last visible bar has index 5, but we also have partially visible (for 20%) 6th bar.
```

docs/time-scale.md
ezhukovskiy, line 236:
```suggestion
Thus, the visible logical range in the image above is from -4.73 to 5.05.
```

docs/series-basics.md
timocov: Oh, really, you're right! I guess it'd better to fix `barsInLogicalRange` method to return more expected result (as described in the docs)?

Let's we have series with the data on indexes `[0..10, 30..60]` and we request bars in `[15, 25]` range. I'd say here we should have the following result:

```js
{
	barsBefore: 15,
	barsAfter: 35,
}
```

_(yeah, with empty `from` and `to` fields because there is no data in requested range)_

What do you think?

timocov:
@gkaindl thanks for having time to taking a look at PR and your feedback!

> Regarding the failing tests with the visible range, yes, sorry, this was most likely because I was using round() instead of floor()/ceil() to convert the ranges, but I saw that you fixed this right away :-)

No problem at all, I even didn't think that you've broke that tests, because I seen them only after my changes though 😂 Anyway, they are fixed now so we don't need to worry about that 🙂

> Also, when 3.0 comes out, I think it would be great to eventually add the new ability to dynamically load new data during scrolling to the demo page at en.tradingview.com/lightweight-charts too – From my perspective as a user, that was the first thing I was looking at to check out what the library can do, and since there was no example about dynamic loading, that's what made me check the issues and eventually pick up the task to add it – Maybe just something where when barsBefore/barsAfter turn negative, it add prepends/appends more randomly generated data after a 1 second timeout to "simulate" the loading?

Great suggestion! I totally agree with that and I think we'll add it. Thank you!

docs/time-scale.md
timocov: Thanks! Done in 99e01ad03bc2b76304871b44c60451bc45a55ee8

docs/series-basics.md
timocov:
```diff
 		const dataFirstIndex = ensureNotNull(bars.firstIndex());
 		const dataLastIndex = ensureNotNull(bars.lastIndex());
 
+		// this means that we request data in the data gap
+		// e.g. let's say we have series with data [0..10, 30..60]
+		// and we request bars info in range [15, 25]
+		// thus, dataFirstBarInRange will be with index 30 and dataLastBarInRange with 10
+		if (dataFirstBarInRange !== null && dataLastBarInRange !== null && dataFirstBarInRange.index > dataLastBarInRange.index) {
+			return {
+				barsBefore: range.from - dataFirstIndex,
+				barsAfter: dataLastIndex - range.to,
+			};
+		}
+
 		const barsBefore = (dataFirstBarInRange === null || dataFirstBarInRange.index === dataFirstIndex)
 			? range.from - dataFirstIndex
 			: dataFirstBarInRange.index - dataFirstIndex;
```

Something like that.

gkaindl: Yes, thank you for the added fix, that looks good now! :-)

PR #419, opened 2020-04-28
Override autoscale

### LINKED ISSUE #392: Add ability to modify price scale's price range

eugene-korobko:
**Type of PR:** enhancement

**PR checklist:**

- [x] Includes tests
- [x] Documentation update

Added ability to override autoscale result for series

### DIFF
--- a/docs/series-basics.md
+++ b/docs/series-basics.md
@@ -125,7 +125,7 @@ There is an options called `autoscaleProvider` that allows overriding visible pr
 
 ```javascript
 var firstSeries = chart.addLineSeries({
-    autoscaleProvider: () => {
+    autoscaleInfoProvider: () => {
         return {
             priceRange: {
                 minValue: 0,
@@ -136,13 +136,13 @@ var firstSeries = chart.addLineSeries({
 });
 ```
 
-So, you could just add a function that returns object with `priceRange` field to make chart thinking the series always has values from 0 to 100. These values aer in prices.
+So, you could just add a function that returns object with `priceRange` field to make chart thinking the series always has values from 0 to 100. These values are in prices.
 
 You could also provide additional margins in pixels. Please be careful and never mix prices and pixels.
 
 ```javascript
 var firstSeries = chart.addLineSeries({
-    autoscaleProvider: () => {
+    autoscaleInfoProvider: () => {
         return {
             priceRange: {
                 minValue: 0,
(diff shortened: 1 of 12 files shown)


docs/series-basics.md
```diff
@@ -118,6 +118,62 @@ chart.applyOptions({
...
+            priceRange: {
+                minValue: 0,
+                maxValue: 100,
+            },
+        };
+    },
+});
+```
+
+So, you could just add a function that returns object with `priceRange` field to make chart thinking the series always has values from 0 to 100. These values are in prices.
```
ezhukovskiy, line 139:
```suggestion
So, you can just add a function that returns an object with `priceRange` field, if you want to override the original series' price range. For example, you can set the range from 0 to 100. These values are in prices.
```

docs/series-basics.md
```diff
@@ -118,6 +118,62 @@ chart.applyOptions({
...
+
+You could also provide additional margins in pixels. Please be careful and never mix prices and pixels.
```
ezhukovskiy, line 141:
```suggestion
You can also provide additional margins in pixels. Please be careful and never mix prices and pixels.
```

docs/series-basics.md
```diff
@@ -118,6 +118,62 @@ chart.applyOptions({
...
+            margins: {
+                above: 10,
+                below: 10,
+            },
+        };
+    },
+});
+```
+
+Actually, `autoscaleInfoProvider` function accepts a parameter with original implementation, so you can call default and adjust the result:
```
ezhukovskiy, line 160:
```suggestion
Actually, `autoscaleInfoProvider` function has an argument `original` which is the default implementation, so you can call it and adjust the result:
```

docs/series-basics.md
```diff
@@ -118,6 +118,62 @@ chart.applyOptions({
...
+        if (res.priceRange !== null) {
+            res.priceRange.minValue -= 10;
+            res.priceRange.maxValue += 10;
+        }
+        return res;
+    },
+});
+```
+
+Please pay your attention that both `priceRange` and `margins` could be `null` in the default result.
```
ezhukovskiy, line 175:
```suggestion
Note that both `priceRange` and `margins` could be `null` in the default result.
```

src/model/price-scale.ts
```diff
@@ -879,7 +879,7 @@ export class PriceScale {
 				}
 
 				if (autoScaleInfo !== null) {
-					const margins = autoScaleInfo.margins();
+					const margins = autoScaleInfo.margins() || null;
```
timocov, line 882: When this helps us?

eugene-korobko: If user overrides the autoscaleInfoProvider in options and the custom function does not return margins

timocov: So why we don't add `|| null` in AutoscaleInfoImpl?

Rostovsoft: Ok, we had a typing issue around it, now fixes

eugene-korobko: We had a typing issue around here, now fixed

Issue #221, opened 2019-09-24
How to set the time zone of the X axis


liubin595338764:
https://jsfiddle.net/TradingView/bmL0vont/

![image](https://user-images.githubusercontent.com/5274428/65504580-e6af9380-def9-11e9-9f8c-158308dba087.png)

How can I set the time zone of the X axis?
My local is the +8 time zone, but it shows the time in the 0 time zone.

`time: 1556877600`  Should show  `2019/05/03 18:00:00`

timocov:
Sorry, but we don't understand Chinese. Please use English.

liubin595338764:
@timocov hi, I have changed to English.

timocov:
> How can I set the time zone of the X axis?

https://github.com/tradingview/lightweight-charts/issues/125#issuecomment-511307557

> And the x-axis time will show an exception on some browsers.

What the browser you have?

liubin595338764:
QQ browser

liubin595338764:
QQ browser : 
`console.log(new Date().toLocaleString())` => `Tue Sep 24 2019 22:46:07 GMT+0800 (CST)`

chrome browser: 
`console.log(new Date().toLocaleString())` => `2019/9/24 下午10:45:48`

timocov:
> QQ browser && HUAWEI mobile phone browser:

Can you please try to set `formatOptions` like we use them here: https://github.com/tradingview/lightweight-charts/blob/b88c8b5844f0aeb171903c8afc784b9f30b49541/src/model/time-scale.ts#L649-L687

It's possible that this browser doesn't support them.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString#Browser_compatibility

liubin595338764:
hi, timocov. I don't know how to use this method, can you give me a list?

liubin595338764:
@timocov 
```html
<div id="time"></div>
<script>
var time = document.querySelector('#time');
time.innerHTML = new Date().toLocaleString('cn', { hour12: false, hour: '2-digit', minute: '2-digit' });
</script>
```
chrome browser:
`13:42`

QQ browser && HUAWEI mobile phone browser:
`Wed Sep 25 2019 13:42:09 GMT+0800 (CST)`

![image](https://user-images.githubusercontent.com/5274428/65572583-1793d500-df9b-11e9-9695-5a7f6d7366c2.png)

timocov:
[...]

PR #425, opened 2020-05-01
Fixed incorrect timestamp locale format

YaoKaiLun:
**Type of PR:** bugfix

**PR checklist:**

fixes # https://github.com/tradingview/lightweight-charts/issues/221

### DIFF
--- a/src/model/default-tick-mark-formatter.ts
+++ b/src/model/default-tick-mark-formatter.ts
@@ -45,7 +45,7 @@ export function defaultTickMarkFormatter(timePoint: TimePoint, tickMarkType: Tic
 	const localDateFromUtc = new Date(
 		date.getUTCFullYear(),
 		date.getUTCMonth(),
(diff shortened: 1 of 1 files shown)


YaoKaiLun:
Does it be supported in 3.0? I cannot use it by npm.

timocov:
It will be available in 3.0, which is still in development. You can subscribe on releases via github so you be notified when the new release happened though.

YaoKaiLun:
> It will be available in 3.0, which is still in development. You can subscribe on releases via github so you be notified when the new release happened though.

Thanks a lot, but i am anxious to fix this problem, is there any other method i can do to handle this situation?

timocov:
Even if I merged this PR it doesn't help you until we publish the new version. Right now it seems that it's impossible to override it in 2.0.0, but you can correct your times to add an offset to them with timezone you'd like to have).

PR #432, opened 2020-05-08
Removed non-exported const enum's JS code

timocov:
**Type of PR:** enhancement

After we enabled `preserveConstEnums` TS has started to generate a code for `const enum`s, even they aren't exported from source file. This PR fixes that and removes code for const enums from JS code if `const enum` isn't exported.

`ts-transformer-strip-const-enums` itself support to strip `const enum` if it isn't exported from some entry point only (so it might remove some additional `const enum`s which are "shared" between modules internally, but aren't exported from the library), but let's do it in another PR after introducing https://github.com/timocov/ts-transformer-properties-rename.

The code of the transformer: https://github.com/timocov/ts-transformer-strip-const-enums

### DIFF
--- a/src/gui/time-axis-widget.ts
+++ b/src/gui/time-axis-widget.ts
@@ -227,6 +227,7 @@ export class TimeAxisWidget implements MouseEventHandlers, IDestroyable {
 	}
 
 	public update(): void {
+		// this call has side-effect - it regenerates marks on the time scale
 		this._chart.model().timeScale().marks();
 	}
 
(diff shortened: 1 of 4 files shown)


timocov:
изображение

👍 👏

src/gui/time-axis-widget.ts
```diff
@@ -228,17 +227,7 @@ export class TimeAxisWidget implements MouseEventHandlers, IDestroyable {
...
-		if (!tickMarks) {
-			return;
-		}
-
-		this._minVisibleSpan = MarkSpanBorder.Year;
-
-		tickMarks.forEach((tickMark: TimeMark) => {
-			this._minVisibleSpan = Math.min(tickMark.span, this._minVisibleSpan);
-		});
+		this._chart.model().timeScale().marks();
```
Nipheris, line 230: Maybe add some comments about side-effects of this call? Looks like the whole method can be completely removed.

timocov, line 230:
```suggestion
		// this call has side-effect - it regenerates marks on the time scale
		this._chart.model().timeScale().marks();
```

timocov: Agreed. Done.

PR #436, opened 2020-05-15
Introduced ts-transformer-properties-rename instead of ts-transformer-minify-privates

timocov:
It allows to minify everything aren't exported to the public space, so it reduces bundle size a lot (by several KB which is several % in min.gz)

**Type of PR:** enhancement

**Overview of change:**

There are several "big" changes:

1. To make prod build we started to use non-composite project (see below).
1. Replaced `ts-transformer-minify-privates` with `ts-transformer-properties-rename` to rename properties while compiling
1. Changed "workflow". Scripts `tsc-all` and `tsc-all-watch` have been dropped.
	Previously these scripts were used to verify that everything is compiled successfully.
	Now, to ensure this, you should use `tsc-verify` script - it does pretty similar to `tsc-all`, but the difference is meaning of the script (see below).
1. `d.ts` bundle right now bundled from `d.ts` files (previously it takes `.ts` files directly) and from non-composite build (see below).
1. Slightly changed `verify` script and now it should be a bit faster (in my env it's 55s vs 65s).
1. Pre-commit hook uses `tsc-verify` script so it might take more time than usual because you won't compile the library with composite project because we use non-composite one (see 1).
1. A bit refactored tsconfigs.
	- We have `tsconfig.options.json` instead of `tsconfig.base.json` with all options, without `plugins` and `outDir` specified.
	- We have `tsconfig.composite.base.json` - the base config for composite sub-projects (it has specified `outDir` option and set `composite` to `true`, so you don't need to set it in every config).
	- We have `tsconfig.prod.json` - it's _non-composite_ config to make a prod build. It has `plugins` option for `ttsc` and `stripInternals` for `dts-bundle-generator`.
1. Structure of `lib` output folder has been a bit changed - previously we had `lib/src` and `lib/tests` where we had all compiled js/dts code. Now there are `lib/prod` and `lib/composite` - output folders for different builds (prod and composite ones).
1. I have disabled `no-redundant-jsdoc` tslint rule because it failed with error on `/** @public */` comments ¯\_(ツ)_/¯
1. `ts-transformer-strip-const-enums` added in #432 now detects whether `const enum` is exported from entry point (`src/index.ts`) and it allows us strip some additional `const enum`s.

**Composite vs non-composite**

Composite projects are great solution - it allows you split your project on different areas and compile them "independently" with some references between them. Another cool features of composite projects are built-in support to detect cyclic dependencies between projects and detecting whether you're able to import module from other project or not. This helps us several (2-3) times while develop and disallow, for instance, import a module from api level in model or even gui sub-project. That's why we don't want to opt out them back.

In opposite, non-composite project has the whole program information while compilation - which symbols are exported _from entry point_, which aren't (composite projects know about current compiled project and don't know about project which will use this project).

So, to make kind of "whole program optimization" we should use non-composite projects, but to be sure about correct imports/dependencies and structure we should use composite ones.

That's why this PR exist in the way you can see it.

To build the library we use non-composite project, but to verify that everything is ok - composite ones.

I hope I didn't miss something 🙂

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,12 +80,9 @@ jobs:
     executor: node-latest-executor
     steps:
       - checkout-with-deps
-      - run: npm run tsc-all
-
-      # make sure that the project is compiled with non-composite project
-      # it doesn't output anything, but is used by IDE or tools
-      - run: ./node_modules/.bin/tsc -p tsconfig.json
-
+      # make sure that the project is compiled successfully with composite projects
+      # so we don't have cyclic deps between projects and wrong imports
+      - run: npm run tsc-verify
       - run: npm run build:prod
       - persist_to_workspace:
           root: ./
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -6,8 +6,7 @@ The minimal supported version of [NodeJS](https://nodejs.org/) for development i
 
 - `npm run tsc` - compiles the source code only (excluding tests)
 - `npm run tsc-watch` - runs the TypeScript compiler in the watch mode for source code (same as `tsc`, but in the watch mode)
-- `npm run tsc-all` - compiles everything (source code and tests)
-- `npm run tsc-all-watch` - runs the TypeScript compiler in watch mode for source code and tests (same as `tsc-all`, but in watch mode)
+- `npm run tsc-verify` - compiles everything (source code and tests) with composite projects config to ensure that no invalid imports or cyclic deps are found
 
 ## Bundling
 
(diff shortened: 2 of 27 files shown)


timocov:
image

We make the library even smaller than it was before all changes in 3.0. Good point for blog post though 😂

timocov:
We need to build locally (and in e2e tests) with non-composite projects, because we want to get _the same state_ before and after minifying so we'll be sure that everything is fine locally (we don't need to build production build for that like with google closure compiler).

timocov:
Ah, goodbye debugging production build 😂 

image

PR #445, opened 2020-05-19
Add internal "time" and "price" options to "axisPressedMouseMove"

### LINKED ISSUE #440: Disabling/enabling scaling axis for both price and time

It's not possible to disable scaling axis for only price or time.

Instead of disabling/enabling scaling axis like that:

```
chart.applyOptions({
    handleScale: {
        axisPressedMouseMove: false
    }
});
```

We can seperate scaling axis to options for only price or time like that:

```
chart.applyOptions({
    handleScale: {
        priceAxisPressedMouseMove: false,
        timeAxisPressedMouseMove: true
    }
});
```

**Additional context**

I would like to create an PR for that if it's okay.

yusufsahinhamza:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #440
- [x] Includes tests
- [x] Documentation update

**Overview of change:**

Added internal `time` and `price` options to `axisPressedMouseMove`.

I think documentation is could be better than i did, you guys can change it.


docs/customization.md
```diff
@@ -318,13 +318,22 @@ You can also set `handleScroll` to `true` or `false` to enable or disable all th
 
 |Name                        |Type   |Default  |Description|
 |----------------------------|-------|---------|-|
-|`axisPressedMouseMove`|`boolean`|`true`|If true, axis scaling with left mouse button pressed is allowed|
+|`axisPressedMouseMove`|`boolean`|`true`|If true, [time and price](#axis-scaling-options) axis scaling with left mouse button pressed is allowed|
```
timocov, line 321: > `boolean`

I guess the type here should be

```
`boolean` &#124; [`{ time, price }`](#axis-scaling-options)
```
?

src/gui/price-axis-widget.ts
```diff
@@ -313,7 +314,7 @@ export class PriceAxisWidget implements IDestroyable {
 	}
 
 	private _mouseDownEvent(e: TouchMouseEvent): void {
-		if (this._priceScale === null || this._priceScale.isEmpty() || !this._pane.chart().options().handleScale.axisPressedMouseMove) {
+		if (this._priceScale === null || this._priceScale.isEmpty() || !(this._pane.chart().options().handleScale as HandleScaleOptionsInternal).axisPressedMouseMove.price) {
```
timocov, line 317: I think that `this._pane.chart().options()` should return "internal" options and you don't need to "cast" type here (and everywhere in non-api code).

timocov: You can see as it is done for `handleScale` and `handleScroll` options.

tests/e2e/graphics/test-cases/applying-options/axis-scaling-boolean-accordingly.js
```diff
@@ -0,0 +1,31 @@
...
+	mainSeries.setData(generateData());
+
+	chart.applyOptions({
+		handleScale: {
+			axisPressedMouseMove: false,
+		},
+	});
+
+	return new Promise((resolve) => {
+		setTimeout(() => resolve(), 1000);
```
timocov, line 29: I guess you don't need to return a Promise at all - this feature is used sometimes when you need to do something async. I don't even think that with the current test runner we can test it properly, so we can drop these tests...

src/model/chart-model.ts
```diff
@@ -37,10 +37,21 @@ export interface HandleScrollOptions {
 export interface HandleScaleOptions {
 	mouseWheel: boolean;
 	pinch: boolean;
-	axisPressedMouseMove: boolean;
+	axisPressedMouseMove: AxisPressedMouseMoveOptions | boolean;
```
timocov, line 40: I think it'd be better to create one more type: `HandleScaleOptionsInternal`, use it in `ChartOptionsInternal`. In "internal" type we don't need to have `| boolean` part, we should migrate it on API level and don't pass/use it outside of API.

timocov: In this case you don't need to use casting though.

docs/customization.md
yusufsahinhamza: Well yes or we just need to make type and default value like: 

type: [`{ time, price }`](#axis-scaling-options)
default: `{ time: true, price: true }`

Because in the line 335; we are already said "You can also set axisPressedMouseMove to true or false to enable or disable all the above options.", just like handleScale?

timocov: > Because in the line 335; we are already said "You can also set axisPressedMouseMove to true or false to enable or disable all the above options.", just like handleScale?

Yeah, agree.

src/model/chart-model.ts
```diff
@@ -37,10 +37,21 @@ export interface HandleScrollOptions {
 export interface HandleScaleOptions {
 	mouseWheel: boolean;
 	pinch: boolean;
-	axisPressedMouseMove: boolean;
+	axisPressedMouseMove: AxisPressedMouseMoveOptions;
```
timocov, line 40:
```suggestion
	axisPressedMouseMove: AxisPressedMouseMoveOptions | boolean;
```


Shouldn't be like this? Otherwise there is no difference between `HandleScaleOptionsInternal` and `HandleScaleOptions` though.

timocov: Ok, I'll fix it after merge, don't worry.

timocov:
Looks good for me. Thank you for contributing, @yusufsahinhamza!

Issue #437, opened 2020-05-15
Custom font families for watermarks


meetmangukiya:
**Is your feature request related to a problem? Please describe.**

Just would like to be able to change the font

I think should be easier to implement, just adding another argument to the `makeFont` call from the brief look at it?

timocov:
Yeah, it needs to add the new option `fontFamily` to [WatermarkOptions](https://github.com/tradingview/lightweight-charts/blob/a6c3e8134be68ee9e79f0382540eba054d1e35cb/src/model/watermark.ts#L9-L22 ) and use it [in pane view](https://github.com/tradingview/lightweight-charts/blob/a6c3e8134be68ee9e79f0382540eba054d1e35cb/src/views/pane/watermark-pane-view.ts#L59).

Feel free to make a PR.

PR #448, opened 2020-05-23
Add support for custom font families and styling to watermarks

meetmangukiya:
Closes https://github.com/tradingview/lightweight-charts/issues/437

**Type of PR:** 

**PR checklist:**

- [x] Addresses an existing issue: fixes https://github.com/tradingview/lightweight-charts/issues/437
- [ ] Includes tests
- [x] Documentation update

**Overview of change:**

### DIFF
--- a/src/api/options/watermark-options-defaults.ts
+++ b/src/api/options/watermark-options-defaults.ts
@@ -1,10 +1,12 @@
+import { defaultFontFamily } from '../../helpers/make-font';
+
 import { WatermarkOptions } from '../../model/watermark';
 
 export const watermarkOptionsDefaults: WatermarkOptions = {
 	color: 'rgba(0, 0, 0, 0)',
 	visible: false,
 	fontSize: 48,
-	fontFamily: '',
+	fontFamily: defaultFontFamily,
 	fontStyle: '',
 	text: '',
 	horzAlign: 'center',
--- a/src/api/options/watermark-options-defaults.ts
+++ b/src/api/options/watermark-options-defaults.ts
@@ -4,6 +4,8 @@ export const watermarkOptionsDefaults: WatermarkOptions = {
 	color: 'rgba(0, 0, 0, 0)',
 	visible: false,
 	fontSize: 48,
+	fontFamily: '',
+	fontStyle: '',
 	text: '',
 	horzAlign: 'center',
 	vertAlign: 'center',
--- a/src/views/pane/watermark-pane-view.ts
+++ b/src/views/pane/watermark-pane-view.ts
@@ -56,7 +56,7 @@ export class WatermarkPaneView implements IUpdatablePaneView {
 		data.lines = [
 			{
 				text: options.text,
-				font: makeFont(options.fontSize),
+				font: makeFont(options.fontSize, options.fontFamily, options.fontStyle),
 				lineHeight: options.fontSize * 1.2,
 				vertOffset: 0,
 				zoom: 0,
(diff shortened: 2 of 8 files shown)


meetmangukiya:
I am not sure how to fix the dts-changes build, any suggestions?

src/api/options/watermark-options-defaults.ts
```diff
@@ -4,6 +4,8 @@ export const watermarkOptionsDefaults: WatermarkOptions = {
 	color: 'rgba(0, 0, 0, 0)',
 	visible: false,
 	fontSize: 48,
+	fontFamily: '',
```
timocov, line 7: This doesn't fit with default value from the docs. I guess we need to use `defaultFontFamily` from `make-font.ts`?

src/views/pane/watermark-pane-view.ts
```diff
@@ -56,7 +56,7 @@ export class WatermarkPaneView implements IUpdatablePaneView {
 		data.lines = [
 			{
 				text: options.text,
-				font: makeFont(options.fontSize),
+				font: makeFont(options.fontSize, options.fontFamily, options.fontStyle),
```
timocov, line 59: If `options.fontFamily` isn't `undefined` (default value is empty string?), then `makeFont` will take it as the value (I guess we expect to use default one instead in this case?). I guess we need 1) to change default font family and 2) change `makeFont` so it shouldn't allow empty font family (not sure about that, what do you think?).

timocov:
> I am not sure how to fix the dts-changes build, any suggestions?

That's expected. That test means that we have some changes in public API so we need to carry about that and see that everything is public as expected. In your case this is expected so please don't worry about it.

src/api/options/watermark-options-defaults.ts
meetmangukiya: right, should've been undefined for the default one to get used: https://github.com/tradingview/lightweight-charts/blob/master/src/helpers/make-font.ts#L18

timocov: Yeah, maybe `undefined` is good option as well. https://github.com/tradingview/lightweight-charts/pull/448#discussion_r429868583

timocov:
Hey @meetmangukiya, do you have any updates here? I'm asking because we're going to prepare to 3.0 release soon and I think it'd be awesome to put it into that release as well. If you haven't enough time for it, please let me know and I think we'll finish it.

meetmangukiya:
Hey @timocov sorry for the delay, even though its a small change, haven't been able to find the time to do it and would need some more time to get around this.

timocov:
@meetmangukiya sure, no worries and no problem at all. Just to be sure - you're going to finish it but you need a bit time, right? (to avoid misunderstanding).

meetmangukiya:
@timocov when is the release scheduled for?

timocov:
We don't have specific date for it, but I hope it should be within 1-2 weeks.

timocov:
Hey @meetmangukiya, any updates on this? As I said, we can finish it if you'd like (no pressure, just a reminder 🙂).

meetmangukiya:
@timocov I pushed the changes. Not sure if I got the graphics test right. Also, don't think it will work because this PR introduces visual difference, not sure if it is above or below the tolerance threshold. Sorry for the delay.

timocov:
Hey @meetmangukiya, thanks for update and no worries! I just added some changes ahead of yours, but nothing critical.

> Not sure if I got the graphics test right. 

You got them completely right!

> Also, don't think it will work because this PR introduces visual difference, not sure if it is above or below the tolerance threshold

That's ok. That tests might be broken, they just help us see what's actually changed in visual part.

timocov:
Merged. @meetmangukiya thank you for your contribution!

PR #469, opened 2020-06-08
Fix types inconsistency

timocov:
Proper fix should be added in #470.


docs/time-scale.md
```diff
@@ -16,7 +16,7 @@ Time scale (or time axis) is a horizontal scale at the bottom of the chart that
 |`visible`|`boolean`|`true`|If true, the time scale is shown on a chart|
 |`timeVisible`|`boolean`|`false`|If true, the time is shown on the time scale and in the vertical crosshair label|
 |`secondsVisible`|`boolean`|`true`|If true, seconds are shown on the label of the crosshair vertical line in `hh:mm:ss` format on intraday intervals|
-|`tickMarkFormatter`|`(TimePoint, TickMarkType, locale) => string`|Default tick marks formatter|Allows to override the tick marks formatter (see below)|
+|`tickMarkFormatter`|`(TimePoint, TickMarkType, locale) => string` &#124; `undefined`|`undefined`|Allows to override the tick marks formatter (see below)|
```
kirchet, line 19: Seems like a typo `undefined`|`undefined`

timocov: Actually one `undefined` is for "type" column, another one is for default value :)

PR #481, opened 2020-06-15
TextWidthCache refactoring & tests

subzey:
**Type of PR:** 

**PR checklist:**

- [x] Includes tests
- [x] Documentation update (not required)

**Overview of change:**

No signature changes.

Added some tests.

Changed the cache from `Record<string, { tick: number, width: number}>` to just `Record<string, number>`

Fixed `__proto__` and `prototype` were read from cache object proto

Changed a ticks2indices record with even growing indices into a cyclic array structure,

-10 bytes .min.gz :grin: 

The `private _cache` can be a `Map<string, number>` as well. It's an easy change and I can apply it if you wish.

### DIFF
--- a/src/model/text-width-cache.ts
+++ b/src/model/text-width-cache.ts
@@ -2,51 +2,45 @@ const defaultReplacementRe = /[2-9]/g;
 
 export class TextWidthCache {
 	private readonly _maxSize: number;
-	private _actualSize: number = 0;
-	private _usageTick: number = 1;
-	private _oldestTick: number = 1;
-	private _tick2Labels: Record<number, string> = {};
-	private _cache: Record<string, { width: number; tick: number }> = {};
+	private _cache: Record<string, number> = Object.create(null);
+	/** A "cyclic buffer" of cache keys */
+	private _keys: string[] = [];
+	/** Current index in the "cyclic buffer" */
+	private _keysIndex: number = 0;
 
 	public constructor(size: number = 50) {
 		this._maxSize = size;
 	}
 
 	public reset(): void {
-		this._actualSize = 0;
-		this._cache = {};
-		this._usageTick = 1;
-		this._oldestTick = 1;
-		this._tick2Labels = {};
+		this._cache = Object.create(null);
+		this._keys = [];
+		this._keysIndex = 0;
 	}
 
 	public measureText(ctx: CanvasRenderingContext2D, text: string, optimizationReplacementRe?: RegExp): number {
 		const re = optimizationReplacementRe || defaultReplacementRe;
 		const cacheString = String(text).replace(re, '0');
 
-		if (this._cache[cacheString]) {
-			return this._cache[cacheString].width;
-		}
-
-		if (this._actualSize === this._maxSize) {
-			const oldestValue = this._tick2Labels[this._oldestTick];
-			delete this._tick2Labels[this._oldestTick];
-			delete this._cache[oldestValue];
-			this._oldestTick++;
-			this._actualSize--;
+		if (cacheString in this._cache) {
+			return this._cache[cacheString];
 		}
 
 		const width = ctx.measureText(cacheString).width;
 
-		if (width === 0 && !!text.length) {
+		if (width === 0 && text.length !== 0) {
 			// measureText can return 0 in FF depending on a canvas size, don't cache it
 			return 0;
 		}
 
-		this._cache[cacheString] = { width: width, tick: this._usageTick };
-		this._tick2Labels[this._usageTick] = cacheString;
-		this._actualSize++;
-		this._usageTick++;
-		return width;
+		if (this._keysIndex < this._keys.length) {
+			// Cleanup the oldest value
+			delete this._cache[this._keys[this._keysIndex]];
+		}
+
+		this._keys[this._keysIndex] = cacheString;
+		// Advance the index so it always points the oldest value
+		this._keysIndex = (this._keysIndex + 1) % this._maxSize;
+		return this._cache[cacheString] = width;
 	}
 }
--- a/src/model/text-width-cache.ts
+++ b/src/model/text-width-cache.ts
@@ -1,15 +1,9 @@
+export type CanvasCtxLike = Pick<CanvasRenderingContext2D, 'measureText'>;
+
 const defaultReplacementRe = /[2-9]/g;
 
-export interface TextMetricsLike {
-	width: number;
-}
-
-export interface CanvasCtxLike {
-	measureText(text: string): TextMetricsLike;
-}
-
 export class TextWidthCache {
-	private _cache: Map<string | undefined, number> = new Map();
+	private _cache: Map<string, number> = new Map();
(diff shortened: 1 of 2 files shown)


src/model/text-width-cache.ts
```diff
@@ -2,51 +2,45 @@ const defaultReplacementRe = /[2-9]/g;
...
-		return width;
+		if (this._keysIndex < this._keys.length) {
+			// Cleanup the oldest value
+			delete this._cache[this._keys[this._keysIndex]];
+		}
+
+		this._keys[this._keysIndex] = cacheString;
+		// Advance the index so it always points the oldest value
+		this._keysIndex = (this._keysIndex + 1) % this._maxSize;
+		return this._cache[cacheString] = width;
```
timocov, line 44:
```suggestion
		this._cache[cacheString] = width;
		return width;
```

src/model/text-width-cache.ts
```diff
@@ -2,51 +2,45 @@ const defaultReplacementRe = /[2-9]/g;
...
 			// measureText can return 0 in FF depending on a canvas size, don't cache it
 			return 0;
 		}
 
-		this._cache[cacheString] = { width: width, tick: this._usageTick };
-		this._tick2Labels[this._usageTick] = cacheString;
-		this._actualSize++;
-		this._usageTick++;
-		return width;
+		if (this._keysIndex < this._keys.length) {
```
timocov, line 36: I'm think that it might be a bit complicated to understand why we can it here and what does it actually mean here. I'd like to add comment here (something like "if that's true that means that the buffer is filled already so we need to clear the oldest value").

src/model/text-width-cache.ts
```diff
@@ -2,51 +2,45 @@ const defaultReplacementRe = /[2-9]/g;
...
+			// Cleanup the oldest value
+			delete this._cache[this._keys[this._keysIndex]];
+		}
+
+		this._keys[this._keysIndex] = cacheString;
```
timocov, line 41: Is that good to extend an array in this way? Can't we push instead? Maybe even prepare an array while reset and fill to avoid wholes...

tests/unittests/text-width-cache.spec.ts
```diff
@@ -0,0 +1,153 @@
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+
+import { TextWidthCache } from '../../src/model/text-width-cache';
+
+function fakeMeasureText(str: string): { width: number } {
```
timocov, line 6:
```suggestion
function fakeMeasureText(str: string): Pick<TextMetrics, 'width'> {
```

? 🤓

timocov: If so, then needs to replace every `{ width: number }` in this file...

tests/unittests/text-width-cache.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+				invocations[str] = (invocations[str] || 0) + 1;
+				return { width: 42 };
+			},
+		} as unknown as CanvasRenderingContext2D;
+
+		const textWidthCache = new TextWidthCache(3);
+
+		textWidthCache.measureText(fakeCtx, '__proto__');
+		textWidthCache.measureText(fakeCtx, 'prototype');
+		textWidthCache.measureText(fakeCtx, 'hasOwnProperty');
```
timocov, line 116: Maybe we can just use `Map` instead of `Object.create(null)` as cache value and avoid this on purpose? It's possible that it might even a bit faster though...

src/model/text-width-cache.ts
subzey: I've commented everything in my new commit. It should be more clear waht happens now

tests/unittests/text-width-cache.spec.ts
subzey: According to SOLID we could shrink the required interface in the module itself

```diff
-public measureText(ctx: CanvasRenderingContext2D, text: string, optimizationReplacementRe?: RegExp): number {
+export interface TextMetricsLike {
+  width: number;
+}
+export interface MeasureTextHost {
+  measureText(text: string): TextMetricsLike;
+}
+public measureText(ctx: MeasureTextHost, text: string, optimizationReplacementRe?: RegExp): number {
```

so `CanvasRenderingContext2D` implements `TextMeasurer`. And by the way get rid of ugly `as unknown as CanvasRenderingContext2D` in the tests

src/model/text-width-cache.ts
subzey: Here's a little research about V8 internal array sturcture.

> `new Array(100)` is `HOLEY_ELEMENTS`.

`.fill` and even creating an array copy do not change the structure:

> `new Array(100).fill(undefined)` is `HOLEY_ELEMENTS`

> `new Array(100).fill(undefined).slice()` is `HOLEY_ELEMENTS` although it's a new array and has no holes

`Array.from` allows to pre-allocate the array and keep in packed:

> `Array.from(new Array(100))` is `PACKED_ELEMENTS`

the optional filler function can create a packed array of arbitrary values:

> `Array.from(new Array(100), () => 42)` is `PACKED_SMI_ELEMENTS`

Surprisingly, it works only if the array is created from the array, not the ArrayLike:

> `Array.from({length:100})` is `HOLEY_ELEMENTS`

tests/unittests/text-width-cache.spec.ts
subzey: Yup, now Map is used

src/model/text-width-cache.ts
subzey: Used a fixed preallocated array. PTAL

src/model/text-width-cache.ts
```diff
@@ -1,52 +1,77 @@
 const defaultReplacementRe = /[2-9]/g;
 
+export interface TextMetricsLike {
+	width: number;
+}
+
+export interface CanvasCtxLike {
+	measureText(text: string): TextMetricsLike;
```
timocov, line 8: Ha-ha, our props renamer can't handle it and rename it. Currently both `measureText` and `width` are renamed to `_internal_*` and this can't work. There are 2 options:

- add `@public` JSDoc comment to them (you can see example in colors.ts)
- change types to make them "related" to the original types from `lib.dom.d.ts` (they should refer to them actually), something like this `type CanvasCtxLike = Pick<CanvasRenderingContext2D, 'measureText'>` (yeah, we can't change a type of the returned value, but I don't think that it's big deal though)

subzey: Changed it to `Pick<CanvasRenderingContext2D, 'measureText'>`

```ts
type TextMetricsLike = Pick<TextMetrics, 'width'>;
interface CtxLike extends Partial<CanvasRenderingContext2D> {
  measureText(text: string): TextMetricsLike & TextMetrics;
}
```
seems to work fine, but it's hacky as hell.

Tests now uses `as undefined as TextMetrics` again, but I guess that's not a big deal.

timocov:
Thanks!

PR #486, opened 2020-06-18
Data layer refactoring

timocov:
**Type of PR:** bugfix, enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #355, #270
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

There is a lot of changes so I think it's better to review changes commit-by-commit.

PR #487, opened 2020-06-18
Added updating watermark views

### LINKED ISSUE #485: Applying watermark in setTimeout doesn't make an effect

**Steps/code to reproduce:**
It is not effective when i use setTimeout to set applyOptions

**Actual behavior:**

**Expected behavior:**

**Screenshots**
image
image

timocov:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #485
- [x] Includes tests
- [ ] Documentation update

### DIFF
--- a/src/model/chart-model.ts
+++ b/src/model/chart-model.ts
@@ -251,7 +251,7 @@ export class ChartModel implements IDestroyable {
 		return this._grid;
 	}
 
-	public watermarkSource(): Watermark | null {
+	public watermarkSource(): Watermark {
 		return this._watermark;
 	}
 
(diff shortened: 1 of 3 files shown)


src/model/chart-model.ts
```diff
@@ -545,6 +545,7 @@ export class ChartModel implements IDestroyable {
 	}
 
 	public recalculateAllPanes(): void {
+		this._watermark.updateAllViews();
```
timocov, line 548: This is here instead of `updateAllPaneViews` to avoid conflicts with #486

PR #492, opened 2020-06-23
Fixes for Shadow DOM support

### LINKED ISSUE #427: Scaling via mouse click and drag doesn't work if chart is inside shadow root.
**Lightweight Charts Version:** 2.0.0

**Steps/code to reproduce:**

Put any chart inside a custom html element using the custom element api of modern browsers.

**Actual behavior:**

If a chart is located inside the shadow root of a custom element scaling time or price bar with the mouse (click + drag) won't work

**Expected behavior:**

Scaling should work as normal

**Screenshots**

timocov:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #427 
- [ ] Includes tests
- [ ] Documentation update

/cc @subzey @trading-peter

### DIFF
--- a/src/gui/mouse-event-handler.ts
+++ b/src/gui/mouse-event-handler.ts
@@ -354,9 +354,15 @@ export class MouseEventHandler implements IDestroyable {
 				if (!this._handler.mouseDownOutsideEvent) {
 					return;
 				}
+
+				if (event.composed && this._target.contains(event.composedPath()[0] as Element)) {
+					return;
+				}
+
 				if (event.target && this._target.contains(event.target as Element)) {
 					return;
 				}
+
 				this._handler.mouseDownOutsideEvent();
 			};
 
--- a/src/gui/mouse-event-handler.ts
+++ b/src/gui/mouse-event-handler.ts
@@ -41,7 +41,6 @@ export interface TouchMouseEvent {
 	// TODO: remove this after rewriting MouseEventHandler to handle touch and mouse event separately
 	readonly type: 'touch' | 'mouse';
 
-	target: MouseEvent['target'];
 	view: MouseEvent['view'];
 }
 
@@ -519,7 +518,6 @@ export class MouseEventHandler implements IDestroyable {
 
 			type: event.type.startsWith('mouse') ? 'mouse' : 'touch',
 
-			target: eventLike.target,
 			view: event.view,
 		};
 	}


src/gui/chart-widget.ts
```diff
@@ -29,7 +29,25 @@ export interface MouseEventParamsImpl {
...
 
+function getShadowRootHostForElement(element: HTMLElement): Element | null {
+	// IE doesn't support shadow dom but we don't need to fail here, so let's check whether a method exist
+	// tslint:disable-next-line:strict-type-predicates
+	if (element.getRootNode === undefined) {
+		return null;
+	}
+
+	const rootNode = element.getRootNode();
+	if (rootNode === element.ownerDocument) {
```
subzey, line 40: Looks like https://developer.mozilla.org/en-US/docs/Web/API/Element/shadowRoot fits better for this case

src/gui/pane-widget.ts
```diff
@@ -109,6 +109,7 @@ export class PaneWidget implements IDestroyable {
 		this._mouseEventHandler = new MouseEventHandler(
 			this._topCanvasBinding.canvas,
 			this,
+			chart.getShadowRootHost(),
```
subzey, line 112: Here, and similar places: Couldn't shadow root be obtained from the element that is passed as the 1st arg?

src/gui/chart-widget.ts
```diff
@@ -317,6 +338,14 @@ export class ChartWidget implements IDestroyable {
 		return ensureNotNull(priceAxisWidget).getWidth();
 	}
 
+	/**
+	 * Returns a shadow root host of the container.
+	 * Might be used to correctly handle outside mouse events (to skip composed events properly)
+	 */
+	public getShadowRootHost(): Element | null {
```
subzey, line 345: Uhh.. that's really looks like an overly specific method for a chartwidget.

timocov: Yes! But the only chart widget knows the container where a chart will be placed, because we attach an element to DOM after a widget is initialized in it's parent, and if an element is detached from tree, we'll get `null` as shadow root.

src/gui/pane-widget.ts
timocov: No, it can't. An element couldn't be attached to the DOM yet.

src/gui/chart-widget.ts
timocov: It seems that it works for the root element only (that's the first what I've tried).

image

PR #499, opened 2020-06-26
Added whitespaces support

### LINKED ISSUE #209: Omit value in candlestick chart

timocov:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #209
- [x] Includes tests
- [x] Documentation update

### DIFF
--- a/src/api/data-layer.ts
+++ b/src/api/data-layer.ts
@@ -14,7 +14,7 @@ import {
 	SeriesDataItemTypeMap,
 	Time,
 } from './data-consumer';
-import { getSeriesPlotRowCreator } from './get-series-plot-row-creator';
+import { getSeriesPlotRowCreator, isSeriesPlotRow, WhitespacePlotRow } from './get-series-plot-row-creator';
 import { fillWeightsForPoints } from './time-scale-point-weight-generator';
 
 type TimedData = Pick<SeriesDataItemTypeMap[SeriesType], 'time'>;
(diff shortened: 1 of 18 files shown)


src/api/data-layer.ts
```diff
@@ -143,7 +143,7 @@ interface TimePointData {
 
 	// actually the type of the value should be related to the series' type (generic type)
 	// here, in data layer all data for us is "mutable" by default, but to the chart we provide "readonly" data, to avoid modifying it
-	mapping: Map<Series, Mutable<SeriesPlotRow>>;
+	mapping: Map<Series, Mutable<SeriesPlotRow | WhitespacePlotRow>>;
```
timocov, line 146: Mapping `_pointDataByTimePoint` in contains all series points (including whitespaces), rather than `_seriesRowsBySeries` where we can exactly rows we "send" to the series (without whitespaces).

PR #509, opened 2020-07-03
Incroduces size-limit instead of codechecks

### LINKED ISSUE #356: Migrate from codechecks
Looks like codechecks isn't supported anymore. Needs to migrate to something else.

timocov:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #356 
- [x] Includes tests
- [ ] Documentation update


timocov:
Example of failed size-limit test: https://app.circleci.com/pipelines/github/tradingview/lightweight-charts/845/workflows/c8e98a44-621c-417f-8809-e7749c1bc3a2/jobs/9608

scripts/size-limit.js
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env node
```
Nipheris, line 1: Maybe rename this file to `compare-size-with-merge-base.js`

scripts/size-limit.js
```diff
@@ -0,0 +1,105 @@
...
+	if (headRev.length === 0 || headRev === 'HEAD') {
+		headRev = runForOutput('git rev-parse HEAD');
+	}
+
+	const revToCheck = runForOutput(`git merge-base origin/master "${headRev}"`);
+
+	console.log(`Using "${revToCheck}" as base\n`);
+
+	console.log(`Switching to ${revToCheck}`);
+	runForSuccess(`git checkout ${revToCheck}`);
```
Nipheris, line 75: What about checking out to the separate `worktree`? We can pass path to directory using the script argument (to be able to configure it for different environments).

scripts/size-limit.js
timocov: Agree.

scripts/size-limit.js
timocov: Yeah, maybe, but there are some questions: 1) caching node_modules (in different worktrees it will be different folders and we need to wait until everything is installed twice) 2) where that worktrees should be (afaik they can't be in the repo dir)

I think we need to reconsider that, but in different issue and for all tests.

scripts/size-limit.js
timocov: Done in c8448c8

Issue #497, opened 2020-06-25
Add example to how to use chart with timezones


timocov:
UPD: don't forget about #529.

timocov:
Fixed in #781

Issue #414, opened 2020-04-22
Add API to get chart values (data, markers, etc)


allartdo:
Why is there no getPrice?
or getMarkers();

Would really like them, makes it much easyer for me to use.

Example:
This is how I need to get the current price now:
        `let currentPrice = Array.from(Array.from(priceArea._dataUpdatesConsumer.de.ee)[priceArea._dataUpdatesConsumer.de.ee.size - 1][1].mapping)[0][1].value;`

Just write getPrice(); would make it much more easyer and would have saved me at least 3 hours of searching.

timocov:
Why can't you store data somewhere in your code before passing it to series?

I just want to understand the case where it can be real issue.

Currently, we don't just store the data in series and use it everywhere. The data processed via our data-layer, which split that data and then convert to data series can understand. So, even if we can return the data back to you, it will be different type/array/etc. So I need to understand use case here.

allartdo:
Its not all issue's, its also a big time saver and userbility upgrades. Like I said, it took me 3 hours the get the current price in my chart. If you would have a build-in function that litirally does what I wrote in my in first commend but then the priceArea variable whould be changed to a lineSerie variable like the code below it would save other people alot of time.
```Javascript
/**
 * Returns the current value on the given line (yAxis).
 * 
 * @param {Object} lineSerie //Your line serie, in the image example: the green or blue line.
 */
function getCurrentValue(lineSerie) {
    return Array.from(Array.from(lineSerie._dataUpdatesConsumer.de.ee)[lineSerie._dataUpdatesConsumer.de.ee.size - 1][1].mapping)[0][1].value;
}
```
To make it more clear: you would get the value you see at the end of this line (or the label) and you can get it of the line you pass trough in the function. The lineSerie variable handles this. So in this case it would return -124.99 if you pass the green line, and -138.43 if you pass the blue line.

![image](https://user-images.githubusercontent.com/32697889/80212370-a81cbb80-8637-11ea-891d-c6630177f68e.png)
This is ofcourse just an example of how the get the current price, it might be so that you know a better way so feel free to change it.

If I look at a function like setMarkers(); then I think this is a very weird function which in my opinion does not work as great as I could imagine.

1. Why do you have to add all markers all over again when you want to implement just one?
2. Why is there only this setMarkers function and not a getMarkers function or getMarker?
[...]

Issue #315, opened 2020-02-04
Add run-time validation of inputs in debug mode


timocov:
Needs add run-time checks for passed data in DEBUG mode only.

Examples of data validation:

- order of data by time
- valid date/values (numbers, not strings)

Related to #313 https://github.com/tradingview/lightweight-charts/commit/84154b8fa2f37af191d4c9baf7f9ffdf00a83c90 https://github.com/tradingview/lightweight-charts/issues/348#issuecomment-591408462

PR #529, opened 2020-07-20
Explicitly label unix timestamp as a number

Slurpgoose:
I missed "prevent errors, you should cast the numeric type of the time to UTCTimestamp" when reading this initally. This extra comment may help future library users avoid accidentally cast timestamp as a string and open new issues.

**Type of PR:** 

**PR checklist:**

- [ ] Includes tests
- [x] Documentation update

**Overview of change:**

### DIFF
--- a/docs/time.md
+++ b/docs/time.md
@@ -13,6 +13,7 @@ Example:
 ```js
 const timestamp = 1529884800; // June 25, 2018
 ```
+- `timestamp` (`number`) - UTC
 
 ## Business day object
 


timocov:
I'm afraid that they shouldn't be UTC actually. It's preferred to provide them in UTC, but to emulate "timezones", you might to pass date with offset (we'll provide examples/docs for that in #497). Thus, I think that it's better to write about that in a fix for that issue to avoid confusing and more misunderstanding.

timocov:
I mean, the nature of timestamp is UTC-based by-default (it's number of seconds since `01-01-1970 00:00:00 UTC`), what else there could be? But to emulating timezones you can use some tricks, but it isn't default.

Slurpgoose:
I apologize if the intention of the commit was unclear, I had an issue with the error message stating "string format must be in yyyy-mm-dd format" when a Unix timestamp is passed as a string. 

I spent an extra 20 minutes looking through the docs to see if there was a configuration setting that had to be passed to accept Unix based timestamps.

It was not until I found issue [#414](https://github.com/tradingview/lightweight-charts/issues/414). Before I found my stupid mistake. I agree that it is a mistake on my end because epoch timestamps should not be stored as strings. Unfortunately the error provided by the library made it difficult for me to understand what specifically was incorrect. I thought that by adding the one line to the documentation it would a good hint for library users in the future to quickly check their data type first.

**I understand your concerns with the `UTC` label as it may mislead people to think that the library will localize timestamps by default.** 

thank you for the reference to the other issue as it is something that I have not considered either. As well thanks for maintaining this great library :)

timocov:
> I had an issue with the error message stating "string format must be in yyyy-mm-dd format" when a Unix timestamp is passed as a string.

I guess we'll help solve such issues faster in #315. But probably we should say that it should be _number_ - 123 (not string-like-number value `'123'`) then in this PR?

> Unfortunately the error provided by the library made it difficult for me to understand what specifically was incorrect

Agree. Do you have full error message you got here? I thought that we log passed date value as well in the message?

> I thought that by adding the one line to the documentation it would a good hint for library users in the future to quickly check their data type first.

Yes, sure! But I think that it shouldn't be about whether that value should be UTC or not, that note should be about type of the value IMO (string vs number).

Slurpgoose:
> But probably we should say that it should be number - 123 (not string-like-number value '123') then in this PR?

I completely agree, I think that makes the most sense. I attempted to copy the format from [Business day object](https://github.com/tradingview/lightweight-charts/blob/master/docs/time.md#business-day-object) and just missed the mark.

> Do you have full error message you got here? 

`Error: Invalid date string=1595284200, expected format=yyyy-mm-dd`

you are correct, the issue is simply user error and I have already put way more thought into it than its worth. 

If this issue happens frequently maybe adding  a second condition `isNan(parseInt(timestamp)` to thow an error message specifically for a number passed as a string value at line [src/api/data-layer.ts Line 84](https://github.com/tradingview/lightweight-charts/blob/db0472d2c9b514a7523499c0a11f16cea0474756/src/api/data-layer.ts#L84) would eliminate all confusion. 

Although after looking at the code and running the example it seems pointless to cover an edge case that the new Date constructor does not even handle.

Cheers, sorry for taking up your time.

timocov:
> If this issue happens frequently maybe adding a second condition isNan(parseInt(timestamp) to thow an error message specifically for a number passed as a string value at line src/api/data-layer.ts Line 84 would eliminate all confusion.

Not sure what exactly condition we need to add there, because `parseInt` will handle both of strings properly:

image

But I think it's a good direction to handle it in #315 anyway, thanks!

PR #555, opened 2020-08-28
bugfix #549. Correct time scale behavior on setData

### LINKED ISSUE #549: timeScale() changes visible range on setData()
Based on https://stackoverflow.com/questions/63557823/lightweight-charts-timescale-changes-visiblerange-on-setdata/63562298#63562298
 
**Lightweight Charts Version:** 3.1.5

**Actual behavior:**

`setData` move to right visible range of `timeScale`

**Expected behavior:**

`setData` shouldn't set visible range of `timeScale`

**Screenshots:**

kOl3nMpiwY

https://jsfiddle.net/beholderrk/916g3tcf/3/

beholderrk:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #549
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Correct time scale behavior on setData. The time scale doesn't shift when data added to the left.

### DIFF
--- a/src/model/chart-model.ts
+++ b/src/model/chart-model.ts
@@ -483,13 +483,11 @@ export class ChartModel implements IDestroyable {
 		// (and actually we cannot)
 		if (visibleBars !== null && oldFirstTime !== null && newFirstTime !== null) {
 			const isLastSeriesBarVisible = visibleBars.contains(currentBaseIndex);
-			const isFirstSeriesBarVisible = visibleBars.contains(0 as TimePointIndex);
-			const isFirstTimeChanged = oldFirstTime.timestamp > newFirstTime.timestamp;
-			const isDataAdded = newBaseIndex > currentBaseIndex;
-			const isDataAddedToRight = isDataAdded && !isFirstTimeChanged;
-			const isVisibleRangeFilled = !isLastSeriesBarVisible && !isFirstSeriesBarVisible;
+			const isLeftBarShiftToLeft = oldFirstTime.timestamp > newFirstTime.timestamp;
+			const isSeriesPointsAdded = newBaseIndex > currentBaseIndex;
+			const isSeriesPointsAddedToRight = isSeriesPointsAdded && !isLeftBarShiftToLeft;
 
-			if (isDataAddedToRight && isVisibleRangeFilled) {
+			if (isSeriesPointsAddedToRight && !isLastSeriesBarVisible) {
 				const compensationShift = newBaseIndex - currentBaseIndex;
 				this._timeScale.setRightOffset(this._timeScale.rightOffset() - compensationShift);
 			}
--- a/src/model/chart-model.ts
+++ b/src/model/chart-model.ts
@@ -467,9 +467,13 @@ export class ChartModel implements IDestroyable {
 	}
 
 	public updateTimeScale(newBaseIndex: TimePointIndex, newPoints?: readonly TimeScalePoint[]): void {
+		const oldFirstTime = this._timeScale.indexToTime(0 as TimePointIndex);
+
 		if (newPoints !== undefined) {
 			this._timeScale.update(newPoints);
 		}
+
+		const newFirstTime = this._timeScale.indexToTime(0 as TimePointIndex);
 
 		const currentBaseIndex = this._timeScale.baseIndex();
 		const visibleBars = this._timeScale.visibleStrictRange();
@@ -477,12 +481,16 @@ export class ChartModel implements IDestroyable {
 		// if time scale cannot return current visible bars range (e.g. time scale has zero-width)
 		// then we do not need to update right offset to shift visible bars range to have the same right offset as we have before new bar
 		// (and actually we cannot)
-		if (visibleBars !== null) {
+		if (visibleBars !== null && oldFirstTime !== null && newFirstTime !== null) {
 			const isLastSeriesBarVisible = visibleBars.contains(currentBaseIndex);
+			const isFirstSeriesBarVisible = visibleBars.contains(0 as TimePointIndex);
+			const isFirstTimeChanged = oldFirstTime.timestamp > newFirstTime.timestamp;
+			const isDataAdded = newBaseIndex > currentBaseIndex;
+			const isDataAddedToRight = isDataAdded && !isFirstTimeChanged;
+			const isVisibleRangeFilled = !isLastSeriesBarVisible && !isFirstSeriesBarVisible;
 
-			if (newBaseIndex > currentBaseIndex && !isLastSeriesBarVisible) {
+			if (isDataAddedToRight && isVisibleRangeFilled) {
 				const compensationShift = newBaseIndex - currentBaseIndex;
-
 				this._timeScale.setRightOffset(this._timeScale.rightOffset() - compensationShift);
 			}
 		}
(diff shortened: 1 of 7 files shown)


src/model/chart-model.ts
```diff
@@ -467,22 +467,30 @@ export class ChartModel implements IDestroyable {
...
 
 		// if time scale cannot return current visible bars range (e.g. time scale has zero-width)
 		// then we do not need to update right offset to shift visible bars range to have the same right offset as we have before new bar
 		// (and actually we cannot)
-		if (visibleBars !== null) {
+		if (visibleBars !== null && oldFirstTime !== null && newFirstTime !== null) {
 			const isLastSeriesBarVisible = visibleBars.contains(currentBaseIndex);
+			const isFirstSeriesBarVisible = visibleBars.contains(0 as TimePointIndex);
+			const isFirstTimeChanged = oldFirstTime.timestamp > newFirstTime.timestamp;
+			const isDataAdded = newBaseIndex > currentBaseIndex;
```
timocov, line 488: I guess that `newBaseIndex > currentBaseIndex` doesn't really mean that "data was added". base index - it's a term to show the latest _real_ bar across all series on the chart (because of whitespaces, where timescale might have data additional data from the right, but no series has data there).

Thus, I think it's something like "added new bars from the right" or so (but actually this also doesn't mean that the data was added to the right though...)

src/model/chart-model.ts
```diff
@@ -467,22 +467,30 @@ export class ChartModel implements IDestroyable {
...
+			const isDataAddedToRight = isDataAdded && !isFirstTimeChanged;
+			const isVisibleRangeFilled = !isLastSeriesBarVisible && !isFirstSeriesBarVisible;
```
timocov, line 490: Maybe we need to call it something like `areDataEndsVisible`? @Nipheris can you please take a look here?

src/model/chart-model.ts
timocov: Or maybe we can just "inline" it to the `if` statement below (like it was previously)?..

tests/e2e/graphics/test-cases/time-scale/add-data-to-right.js
```diff
@@ -0,0 +1,106 @@
...
+
+	// visible logical range should be changed
+	newRange = timeScale.getVisibleLogicalRange();
+	console.assert(newRange.from === 3, `from index should be changed, expected=${3}, actual=${newRange.from}`);
+	console.assert(newRange.to === 20, `to index should be changed, expected=${20}, actual=${newRange.to}`);
+
+	let timeRangeBeforeSetData = timeScale.getVisibleRange();
+
+	// add 3 bars to right
+	data = [...data, ...generateData(3, (data[data.length - 1].time + 24 * 60 * 60) * 1000)];
```
timocov, line 73: can we extract `24 * 60 * 60` to a variable?

tests/e2e/graphics/test-cases/time-scale/add-data-to-right.js
```diff
@@ -0,0 +1,106 @@
...
+	timeScale.setVisibleLogicalRange({ from: 50, to: 80 });
+	await whenRangeChanged(timeScale);
+
+	timeRangeBeforeSetData = timeScale.getVisibleRange();
+
+	// add 3 bars to right
+	data = [...data, ...generateData(3, (data[data.length - 1].time + 24 * 60 * 60) * 1000)];
+	areaSeries.setData(data);
+
+	newRange = timeScale.getVisibleLogicalRange();
```
timocov, line 99: maybe we need to rename it to `newLogicalRange` to avoid confusing with `newTimeRange`?

timocov (changes requested):
sorry for late reply!

The PR looks good for me, thank for your time and work. Just some naming comments here.

src/model/chart-model.ts
```diff
@@ -467,22 +467,30 @@ export class ChartModel implements IDestroyable {
...
 		const visibleBars = this._timeScale.visibleStrictRange();
 
 		// if time scale cannot return current visible bars range (e.g. time scale has zero-width)
 		// then we do not need to update right offset to shift visible bars range to have the same right offset as we have before new bar
 		// (and actually we cannot)
-		if (visibleBars !== null) {
+		if (visibleBars !== null && oldFirstTime !== null && newFirstTime !== null) {
 			const isLastSeriesBarVisible = visibleBars.contains(currentBaseIndex);
+			const isFirstSeriesBarVisible = visibleBars.contains(0 as TimePointIndex);
+			const isFirstTimeChanged = oldFirstTime.timestamp > newFirstTime.timestamp;
```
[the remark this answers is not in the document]
beholderrk: > And add check that the old oldFirstTime.timestamp exist on the timescale

Sorry, not sure that I understood.
```suggestion
			const isDataAddedToLeft = oldFirstTime.timestamp > newFirstTime.timestamp && visibleBars.contains(???);
```

src/model/chart-model.ts
beholderrk: `areDataEndsVisible` means like letf end of end and right end of data visible inside current timescale? 
```suggestion
			const areDataEndsVisible = isLastSeriesBarVisible && isFirstSeriesBarVisible;
```
But we need shift timescale when all ends of data out of timescale and bars added to right side of data.
Expression like this `if (isDataAddedToRight && !areDataEndsVisible )` will perform the shift when the right edge is visible, which we don't need.

May be `areDataEndsInvisible`?

src/model/chart-model.ts
beholderrk: Ok. I can also suggest "inline" it to the `isDataAddedToRight`
```suggestion
			const isDataAddedToRight = newBaseIndex > currentBaseIndex && !isDataAddedToLeft;
```

tests/e2e/graphics/test-cases/time-scale/add-data-to-right.js
beholderrk: ok

tests/e2e/graphics/test-cases/time-scale/add-data-to-right.js
beholderrk: ok

src/model/chart-model.ts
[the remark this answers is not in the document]
timocov: I think we need to use `timeScale.timeToIndex(oldFirstTime) !== null` - it returns index of the point inside time scale points. We don't need to check that old point was in the visible range (we check it right above).

src/model/chart-model.ts
timocov: Why it's so `&& !isDataAddedToLeft`? 🙂 I think that we can add from both side at the same time.

src/model/chart-model.ts
timocov: > May be areDataEndsInvisible?

Yeah, I think it was typo, I'm sorry 😂 In terms of "data ends are visible" I guess we need to add _any_ there (`IsAnyDataEndVisible`) and make check "negative"

beholderrk:
LCWXxQmLr2
New tests with 2 series and whitespaces

nalepae:
What can be done to help this Pull Request to be merged?

Thanks a lot @beholderrk for your work!

timocov (approved):

nalepae:
@timocov perfect, thank you very much !

Do you know when the next official release is scheduled ?

timocov:
@nalepae we have to finish [3.2 milestone](https://github.com/tradingview/lightweight-charts/milestone/11), I hope it will be within couple of weeks (2-3).

PR #590, opened 2020-10-11
Add crosshair marker property for border and background color

### LINKED ISSUE #310: Add possibility to chose crosshairMarker color, so it will be independent from line-series color
Solution that would be nice its additional field "crosshairMarkerColor" to addLineSeries settings
```
const lineSeries = chart.addLineSeries({
    crosshairMarkerVisible: true,
    crosshairMarkerRadius: 6,
    crosshairMarkerColor: '#111',
});
```

mecm1993:
**Type of PR:** Enhancement

**PR checklist:**

- [X] Addresses an existing issue: fixes #310 
- [x] Includes tests
- [X] Documentation update

**Overview of change:**

lightweight


src/model/series.ts
```diff
@@ -470,6 +474,14 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
 		return 0;
 	}
 
+	private _markerBorderColor(): string {
+		return (this._options as (LineStyleOptions | AreaStyleOptions)).crosshairMarkerBorderColor;
+	}
+
+	private _markerBackgroundColor(): string {
+		return (this._options as (LineStyleOptions | AreaStyleOptions)).crosshairMarkerBackgroundColor;
```
timocov, line 482: I'm not sure that having the only this way to set color is what we previously want. What's about making these properties optional, and use them only if they are set? Otherwise we can use the default behaviour, i.e. `barColorer().barStyle(timePointIndex).barColor` for line color and `this._chartModel.options().layout.backgroundColor` for background color.

src/model/series.ts
```diff
@@ -470,6 +474,14 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
 		return 0;
 	}
 
+	private _markerBorderColor(): string {
+		return (this._options as (LineStyleOptions | AreaStyleOptions)).crosshairMarkerBorderColor;
```
timocov, line 478: What if the series isn't line or area? Why we don't check it like it done in `_markerRadius`?

timocov (changes requested):
Thanks for the PR. A few comments and also, could you provide e2e tests as well please? If you'll have any help with them, just let me know.

[Here is readme](https://github.com/tradingview/lightweight-charts/tree/master/tests/e2e/graphics#graphics-tests) of them, and [this test case](https://github.com/tradingview/lightweight-charts/blob/master/tests/e2e/graphics/test-cases/initial-options/crosshair-label-color.js) you can use as an example.

I'd say we need to have several test cases here:
- initial color for both area and line series
- applying colors for both of them (see `applying-options` folder)
- changing from/to the default behaviour (when we use series' color instead)

src/model/series.ts
mecm1993: Yeah, I can create the test cases.

mecm1993: > I'm not sure that having the only this way to set color is what we previously want. What's about making these properties optional, and use them only if they are set? Otherwise we can use the default behaviour, i.e. `barColorer().barStyle(timePointIndex).barColor` for line color and `this._chartModel.options().layout.backgroundColor` for background color.

Agree, I will perform the changes accordingly.

src/model/series.ts
mecm1993: Sure, I will add the validation as well

src/views/pane/crosshair-marks-pane-view.ts
```diff
@@ -82,8 +82,8 @@ export class CrosshairMarksPaneView implements IUpdatablePaneView {
 			}
 
 			const firstValue = ensureNotNull(s.firstValue());
-			data.lineColor = s.barColorer().barStyle(timePointIndex).barColor;
-			data.backColor = this._chartModel.options().layout.backgroundColor;
+			data.lineColor = seriesData.backgroundColor ? seriesData.backgroundColor : s.barColorer().barStyle(timePointIndex).barColor;
+			data.backColor = seriesData.borderColor ? seriesData.borderColor : this._chartModel.options().layout.backgroundColor;
```
timocov, line 86: Can't we handle this inside `_markerBorderColor` and `_markerBackgroundColor` methods? I think it'd be better to have this logic in one place, so here we'll have something like `data.backColor = seriesData.borderColor;` (without any condition) and anyone else who wants to know that color can get it in the same way without duplicating this lines.

mecm1993: Yes, sorry for the delay, I pushed the changes in this [commit](https://github.com/tradingview/lightweight-charts/pull/590/commits/a8d5a7cf9fc1fdc9378a3e6c3b946b0a5820e345). So now the logic is in one place rather than over the place.

timocov:
Hi @mecm1993, can you please add tests cases as well (see https://github.com/tradingview/lightweight-charts/pull/590#pullrequestreview-506372235)? Everything else looks good for me though.

mecm1993:
Hey @timocov I have been struggling with the tests due to I do not know how to run the e2e tests locally. I dont really understand this part,

```
Running tests
This tests runs against 2 versions of the library - "golden" and "test". Golden is LKG version, test - current tested version of the library.

To run this tests you need use runner.js:

./runner.js ./path/to/golden/standalone/module.js ./path/to/test/standalone/module.js
Each path to the standalone module might be either to a local file (relative/absolute path to a file) or remote file (via http/https). If file is local then local server will be runner to serve that file (see serve-local-files.js module).
```

timocov:
> I do not know how to run the e2e tests locally. I dont really understand this part,

@mecm1993 Ah, sure, let me help you with that.

Our e2e graphics tests are based on comparison screenshots from a baseline and test builds. A baseline build is basically a build from `merge-base` commit your branch based on (if your branch is up-to-date, then it just HEAD of `master` branch). A test build is a build for HEAD of your branch (or any branch which is going to be merged).
Then we open 2 exactly the same pages, but for different builds and wait until code on the page is finished to work and then compare screenshots from both pages. It's kind of regression tests, we want to be sure that we break nothing or a change is intended.

So, to run this tests you need to do the following:

1. Make a build for `merge-base` (or for HEAD of `master`) branch. We need the only standalone build, so you can, for instance, do the following: `npm run clean && npm run build && mv dist merge-base-dist`
1. Make a build for HEAD of your branch: `npm run clean && npm run build`.
1. At this step we have 2 folders `dist` with a build for your branch and `merge-base-dist` for a build for `master` branch. Now we can run tests: `node ./tests/e2e/graphics/runner.js ./merge-base-dist/lightweight-charts.standalone.development.js ./dist/lightweight-charts.standalone.development.js`.

That's it!

Also, if you use VSCode, we provide a run target named `Graphics tests`, you can use it (it will ask you for 2 paths for that 2 files).

Please let me know if you have any question.

timocov:
Hey @mecm1993, do you have any updates here?

mecm1993:
Hi @timocov sorry for the delay on this, I will try to get it done by the end of the week

timocov:
@mecm1993 no pressure at all, just checking the status to understand what we can help with

tests/e2e/graphics/test-cases/applying-options/make-crosshair-colors-change.js
```diff
@@ -0,0 +1,30 @@
...
+	const chart = LightweightCharts.createChart(container);
+	
+	const areaSeries = chart.addAreaSeries({
+        crosshairMarkerBorderColor: 'yellow',
+        crosshairMarkerBackgroundColor: 'red',
+	});
+
+	const lineSeries = chart.addLineSeries({
+        crosshairMarkerBorderColor: 'blue',
+        crosshairMarkerBackgroundColor: 'green',
```
timocov, line 25:
```suggestion
		crosshairMarkerBorderColor: 'yellow',
		crosshairMarkerBackgroundColor: 'red',
	});

	const lineSeries = chart.addLineSeries({
		crosshairMarkerBorderColor: 'blue',
		crosshairMarkerBackgroundColor: 'green',
```

tests/e2e/graphics/test-cases/initial-options/crosshair-border-and-background-color.js
```diff
@@ -0,0 +1,30 @@
...
+	const chart = LightweightCharts.createChart(container);
+	
+	const areaSeries = chart.addAreaSeries({
+        crosshairMarkerBorderColor: '#ffffff',
+        crosshairMarkerBackgroundColor: '#2296f3',
+	});
+
+	const lineSeries = chart.addLineSeries({
+        crosshairMarkerBorderColor: '#ffffff',
+        crosshairMarkerBackgroundColor: '#2296f3',
```
timocov, line 25:
```suggestion
		crosshairMarkerBorderColor: '#ffffff',
		crosshairMarkerBackgroundColor: '#2296f3',
	});

	const lineSeries = chart.addLineSeries({
		crosshairMarkerBorderColor: '#ffffff',
		crosshairMarkerBackgroundColor: '#2296f3',
```

tests/e2e/graphics/test-cases/applying-options/make-crosshair-colors-default.js
```diff
@@ -0,0 +1,44 @@
...
+	const chart = LightweightCharts.createChart(container);
+	
+	const areaSeries = chart.addAreaSeries({
+        crosshairMarkerBorderColor: 'yellow',
+        crosshairMarkerBackgroundColor: 'red',
+	});
+
+	const lineSeries = chart.addLineSeries({
+        crosshairMarkerBorderColor: 'blue',
+        crosshairMarkerBackgroundColor: 'green',
```
timocov, line 25:
```suggestion
		crosshairMarkerBorderColor: 'yellow',
		crosshairMarkerBackgroundColor: 'red',
	});

	const lineSeries = chart.addLineSeries({
		crosshairMarkerBorderColor: 'blue',
		crosshairMarkerBackgroundColor: 'green',
```

tests/e2e/graphics/test-cases/applying-options/make-crosshair-colors-default.js
```diff
@@ -0,0 +1,44 @@
...
+
+	return new Promise(resolve => {
+		setTimeout(() => {
+			areaSeries.applyOptions({
+                crosshairMarkerBorderColor: '#ffffff',
+                crosshairMarkerBackgroundColor: '#2296f3',
+			});
+			lineSeries.applyOptions({
+                crosshairMarkerBorderColor: '#ffffff',
+                crosshairMarkerBackgroundColor: '#2296f3',
```
timocov, line 39:
```suggestion
				crosshairMarkerBorderColor: '#ffffff',
				crosshairMarkerBackgroundColor: '#2296f3',
			});
			lineSeries.applyOptions({
				crosshairMarkerBorderColor: '#ffffff',
				crosshairMarkerBackgroundColor: '#2296f3',
```

timocov (commented):
Thanks! May I ask what's the difference between `tests/e2e/graphics/test-cases/initial-options/crosshair-border-and-background-color.js` and `tests/e2e/graphics/test-cases/applying-options/make-crosshair-colors-change.js` test cases except colors? They look pretty similar...

timocov:
It's interesting, but it seems that the only in one of three tests there is visual changes (in `make-crosshair-colors-change`)...

mecm1993:
```
 - initial color for both area and line series
 - applying colors for both of them (see applying-options folder)
 - changing from/to the default behaviour (when we use series' color instead)
```

I was trying to test the first point in `tests/e2e/graphics/test-cases/initial-options/crosshair-border-and-background-color.js`, but you are right both of them expect the same output at a certain point.

timocov:
Hey @mecm1993,

my apologize - I confused you in https://github.com/tradingview/lightweight-charts/pull/590#discussion_r503101665 when told that we need to make them optional. I just looked at other options we have, especially at `priceLineColor` and it has "special" value `''` (empty string) which the library interprets like "use default fallback". I think we need to have it in the same way and fallback in case of empty string.

As soon as that's my fault, I decided to fix it myself, so you don't need to do anything.

You can see the changed I made in 3cc0447bb604acb5f07f961ac64821d40044f851.

timocov (approved):

timocov:
@mecm1993 thanks for your contribution!

PR #599, opened 2020-10-23
71: Use observer for auto resizing and hide "% 2" magic with a strategy

### LINKED ISSUE #71: Handle resize with ResizeObserver if it's exist in window
Latest

**Steps/code to reproduce:**
Just rotate the device

**Actual behavior:**
Fixed width always

**Expected behavior:**
Need to be responsive

**Screenshots**
Here also: https://www.tradingview.com/lightweight-charts/

### LINKED ISSUE #367: Refactoring resize the chart
https://github.com/tradingview/lightweight-charts/blob/a1c8d55e710783766d1d55a56ff296e8d5ad5642/src/gui/chart-widget.ts#L80-L91

Here we have hidpi fixes for chart's size, but that size applies in constructor only. Thus, if we'll resize the chart, we don't apply that "magic" and can get smoothing.

eugene-korobko:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #71, fixes #367
- [x] Includes tests
- [x] Documentation update

**Overview of change:**
- Added an option `useObserver` that makes the chart resizing automatically to fill parent's client area
- Extracted magic with keeping sizes odd to separate module and encapsulate with interface

One graphic test has been broken because we start applying "odd magic" to explicitly defined sizes


src/model/chart-model.ts
```diff
@@ -80,6 +80,11 @@ export interface ChartOptions {
 	width: number;
 	/** Height of the chart */
 	height: number;
+	/**
+	 * Use observer to resize chart to its parent client area
+	 * Calling code is responsible for providing pilyfill if native implementation is not avaibable
+	 * */
+	useObserver: boolean;
```
timocov, line 87: I'm not sure that `useObserver` is good name for handling resize. Maybe something like `autosize` or `handleResize`?

src/model/chart-model.ts
```diff
@@ -80,6 +80,11 @@ export interface ChartOptions {
 	width: number;
 	/** Height of the chart */
 	height: number;
+	/**
+	 * Use observer to resize chart to its parent client area
+	 * Calling code is responsible for providing pilyfill if native implementation is not avaibable
+	 * */
```
timocov, line 86:
```suggestion
	 * Use ResizeObserver to handle container's resize to resize a chart automatically.
	 * Note that the library WON'T ADD own polyfill for `ResizeObserver` if it doesn't exist in the global scope. So you have to be sure that you enable this option only if a browser supports it or you provide a polyfill instead. Otherwise, a warn message will be printed and no handling resize will be applied.
	 **/
```

Maybe something like that?

timocov: Probably we need to update docs as well accordingly.

src/gui/chart-widget.ts
```diff
@@ -78,29 +85,31 @@ export class ChartWidget implements IDestroyable {
...
-		if (width === 0 || height === 0) {
-			const containerRect = container.getBoundingClientRect();
-			// TODO: Fix it better
-			// on Hi-DPI CSS size * Device Pixel Ratio should be integer to avoid smoothing
-			// For chart widget we decreases because we must be inside container.
-			// For time axis this is not important, since it just affects space for pane widgets
-			if (width === 0) {
-				width = Math.floor(containerRect.width);
-				width -= width % 2;
+		if (options.useObserver) {
```
timocov, line 88: What's about enabling this option on-the-fly? I guess it might be a trick, for instance if you'd like to load polyfills async and then handle resize events.

src/gui/chart-widget.ts
```diff
@@ -78,29 +85,31 @@ export class ChartWidget implements IDestroyable {
...
+			// eslint-disable-next-line no-restricted-syntax
+			if (!('ResizeObserver' in window)) {
+				warn('Options contains "useObserver" flag, but the browser does not support this');
```
timocov, line 91: > the browser does not support this

I think we need to write a bit more about that. Doesn't support what?

src/gui/chart-widget.ts
```diff
@@ -78,29 +85,31 @@ export class ChartWidget implements IDestroyable {
...
+			} else {
+				this._observer = new ResizeObserver((entries: ResizeObserverEntry[]) => {
+					const containerEntry = entries.find((entry: ResizeObserverEntry) => entry.target === container);
+					if (!containerEntry || !containerEntry.devicePixelContentBoxSize) {
```
timocov, line 95: `containerEntry.devicePixelContentBoxSize` Do we need to check it here, even if we don't use it anywhere?

src/gui/chart-widget.ts
```diff
@@ -78,29 +85,31 @@ export class ChartWidget implements IDestroyable {
...
+					this.resize(containerEntry.contentRect.width, containerEntry.contentRect.height);
+				});
+				this._observer.observe(container, { box: 'border-box' });
 			}
-
-			if (height === 0) {
-				height = Math.floor(containerRect.height);
-				height -= height % 2;
+		} else {
+			let size = new Size(this._options.width, this._options.height);
```
timocov, line 103: do we actually need to create this `Size` variable? Can't we use 2 `number` variables? 🤔

src/gui/chart-widget.ts
```diff
@@ -78,29 +85,31 @@ export class ChartWidget implements IDestroyable {
...
+				const containerRect = container.getBoundingClientRect();
+				size = new Size(containerRect.width, containerRect.height);
 			}
-		}
 
-		// BEWARE: resize must be called BEFORE _syncGuiWithModel (in constructor only)
-		// or after but with adjustSize to properly update time scale
-		this.resize(width, height);
+			// BEWARE: resize must be called BEFORE _syncGuiWithModel (in constructor only)
+			// or after but with adjustSize to properly update time scale
```
timocov, line 110: is it comment important while using `ResizeObserver`? Will a `callback` be called async?

src/gui/internal-layout-sizes-hints.ts
```diff
@@ -0,0 +1,40 @@
+import { Size } from './canvas-utils';
+
+export interface InternalLayoutSizeHints {
```
timocov, line 3: What does "internal" here mean? is it inner? or something else?

src/gui/internal-layout-sizes-hints.ts
```diff
@@ -0,0 +1,40 @@
...
+	public suggestTimeScaleHeight(originalHeight: number): number {
+		return originalHeight + originalHeight % 2;
+	}
+
+	public suggestPriceScaleWidth(originalWidth: number): number {
+		return originalWidth + originalWidth % 2;
+	}
+}
+
+export class InternalLayoutSizeHintsKeepIriginal implements InternalLayoutSizeHints {
```
timocov, line 30: Is it used anywhere?

timocov (changes requested):

src/model/chart-model.ts
eugene-korobko: fitContent?

timocov: Nah, we already have that (method, but anyway) 😂 I think we might to look at others libraries, probably they already have this.

src/gui/chart-widget.ts
eugene-korobko: Yes, I think so. Call from observer callback is the same as external resize call

src/gui/internal-layout-sizes-hints.ts
eugene-korobko: This means this hints relate to internal chart layout - panes/scales/etc

src/gui/internal-layout-sizes-hints.ts
eugene-korobko: no, it is just to allow easy disabling of %2 magic...
And I have a Typo here...

timocov: > no, it is just to allow easy disabling of %2 magic...

easy for whom? if it isn't used anywhere, it looks like deadcode, which will be removed from the bundle and you can't access it anyway (or even apply it somehow) 🤔

src/gui/chart-widget.ts
timocov: Don't we need to do a first resize in the widget constructor to have an initial size? I guess we should test it

src/model/chart-model.ts
eugene-korobko: Well, will rename to autoSize
If anyboydy has better name - speak now or forever hold your peace )

src/gui/chart-widget.ts
eugene-korobko: Yes.
ResizeObserver fires the first event immediately not for all the cases, so we have to set initial size with both cases

src/gui/internal-layout-sizes-hints.ts
eugene-korobko: ok

docs/customization.md
```diff
@@ -21,6 +21,28 @@ If you want the chart size to be adjusted when the web page is resized, use the
...
+
+Please pay your attention that `userObserver` option and explicit sizes options `width` and `height` conflict one with others.
+If you specify `userObserver` flag, `width` and `height` options will be ignored in common case
+and could only be used as fallback if using `ResizeObserver` has failed.
+
+The flag `userObserver` could also be set with and unset with `applyOptions` function.
+
+```js
+const chart = LightweightCharts.createChart(document.body, {
+    useObserver: true,
```
timocov, line 42:
```suggestion
### Auto Size

Setting `autoSize` flag to `true` makes chart monitoring container and changing its size on every container resize.

This feature requires [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) class to be available in the global scope.

Note that calling code is responsible for providing a polyfill if required. If the global scope does not have `ResizeObserver`, a warning will appear and the flag will be ignored.

Please pay your attention that `autoSize` option and explicit sizes options `width` and `height` conflict one with others.

If you specify `autoSize` flag, `width` and `height` options will be ignored in common case and could only be used as fallback if using `ResizeObserver` has failed.

The flag `autoSize` could also be set with and unset with `applyOptions` function.

```js
const chart = LightweightCharts.createChart(document.body, {
    autoSize: true,
```

src/gui/chart-widget.ts
```diff
@@ -52,8 +56,14 @@ export class ChartWidget implements IDestroyable {
 	private _clicked: Delegate<MouseEventParamsImplSupplier> = new Delegate();
 	private _crosshairMoved: Delegate<MouseEventParamsImplSupplier> = new Delegate();
 	private _onWheelBound: (event: WheelEvent) => void;
+	private _observer: ResizeObserver | null = null;
+	// replace InternalLayoutSizeHintsKeepOdd with InternalLayoutSizeHintsKeepOriginal to turn "odd magic" off
```
timocov, line 60:
```suggestion
```

src/gui/chart-widget.ts
```diff
@@ -199,6 +207,14 @@ export class ChartWidget implements IDestroyable {
 		const height = options.height || this._height;
 
 		this.resize(width, height);
+
+		if (options.autoSize && !this._observer) {
+			// installing observer will override resize if successfull
+			this._installObserver();
+		}
+		if (options.autoSize === false && this._observer) {
```
timocov, line 215:
```suggestion
		if (options.autoSize && this._observer === null) {
			// installing observer will override resize if successful
			this._installObserver();
		}

		if (!options.autoSize && this._observer !== null) {
```

src/gui/chart-widget.ts
```diff
@@ -199,6 +207,14 @@ export class ChartWidget implements IDestroyable {
 		const height = options.height || this._height;
 
 		this.resize(width, height);
```
timocov, line 209: Shouldn't explicit resize disable autoSize?

src/model/chart-model.ts
```diff
@@ -80,6 +80,14 @@ export interface ChartOptions {
 	width: number;
 	/** Height of the chart */
 	height: number;
+	/**
+	 * Settings this flag to true makes chart monitoring container
+	 * and changing its size on every container resize
+	 * This feature requires `ResizeObserver` class to be avaiable in the global scope
+	 * Calling code is responsibe for providing a polyfill if requried
+	 * If the global scope does not contain `ResizeObserver` object, a warning will appear the the flag will be ignored
```
timocov, line 88:
```suggestion
	 * Setting this flag to `true` makes chart monitoring container and changing its size on every container resize.
	 * This feature requires [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) class to be available in the global scope.
	 * Note that calling code is responsible for providing a polyfill if required. If the global scope does not have `ResizeObserver`, a warning will appear and the flag will be ignored.
	 * Please pay your attention that `autoSize` option and explicit sizes options `width` and `height` conflict one with others.
	 * If you specify `autoSize` flag, `width` and `height` options will be ignored in common case and could only be used as fallback if using `ResizeObserver` has failed.
	 * The flag `autoSize` could also be set with and unset with `applyOptions` function.
```

_(or whatever will be in the docs)_

tests/e2e/graphics/test-cases/initial-options/use-observer.js
```diff
@@ -0,0 +1,47 @@
...
+
+		box.style.position = 'absolute';
+		box.style.top = `${i * 25}%`;
+		box.style.left = 0;
+		box.style.right = 0;
+		box.style.height = '25%';
+
+		container.appendChild(box);
+
+		const chart = LightweightCharts.createChart(box, { useObserver: true, ...config });
```
timocov, line 30:
```suggestion
		const chart = LightweightCharts.createChart(box, { autoSize: true, ...config });
```

timocov (changes requested):
It seems there is some typos here

src/gui/chart-widget.ts
eugene-korobko: probably yes, I've changed the logic a bit

adgower:
I pulled this branch ran npm install then npm run build:prod I could not get autoSize to work? Any tips? @eugene-korobko

adgower:
@eugene-korobko This works great!

SlicedSilver:
It is expected that the graphics tests will have a single failing test for `use-observer` as this is a new test specific for this feature.

SlicedSilver:
@eugene-korobko I've made a few changes so that this PR could be merged. Could you please review if you are happy with the changes.
@Nipheris This PR makes some minor changes to the `% 2` magic code. Could you please check if you are happy with these changes as a temporary measure until we can implement `fancy-canvas` perfectly.

src/model/chart-model.ts
```diff
@@ -247,6 +247,24 @@ export interface ChartOptions {
 	 */
 	height: number;
 
+	/**
+	 * Setting this flag to `true` will make the chart watch the chart container's size and automatically resize the chart to fit it's container whenever the size changes.
```
romfrancois, line 251: typo - it's container => its container

src/model/chart-model.ts
```diff
@@ -247,6 +247,24 @@ export interface ChartOptions {
...
+	 *
+	 * This feature requires [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) class to be available in the global scope.
+	 * Note that calling code is responsible for providing a polyfill if required. If the global scope does not have `ResizeObserver`, a warning will appear and the flag will be ignored.
+	 *
+	 * Please pay your attention that `autoSize` option and explicit sizes options `width` and `height` don't conflict with one another.
```
romfrancois, line 256: ~~your~~

src/gui/internal-layout-sizes-hints.ts
```diff
@@ -0,0 +1,20 @@
+import { Size, size } from 'fancy-canvas';
+
+// on Hi-DPI CSS size * Device Pixel Ratio should be integer to avoid smoothing
+// For chart widget we decreases because we must be inside container.
```
romfrancois, line 4: decrease~~s~~

romfrancois (commented):
Very minor typos

SlicedSilver:
@romfrancois Fixed the typos 👍

romfrancois (approved):

Issue #268, opened 2019-11-26
Zooming by mousewheel with saving rightmost bar position


ws420:
We have some technical issues with the chart and I am wondering if maybe you could help us?

When zooming out on the chart, it looks like this:
![image](https://user-images.githubusercontent.com/48076840/69673008-5a3f5f80-1067-11ea-8fc9-8c65b72f6114.png)
There is a parameter called fixLeftEdge, which prevents scrolling to the left of the first historical bar. Is there a way to reproduce the same behavior, but for the right edge?

The charts on tradingview.com have this by default - zooming out looks like this:
![image](https://user-images.githubusercontent.com/48076840/69673025-62979a80-1067-11ea-8842-82145b8c2c18.png)
Thanks!

timocov:
Did you search the similar issues before? 🙂 https://github.com/tradingview/lightweight-charts/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+fixRightEdge - see #218.

Also, on www.tradingview.com we have different kind of zooming - it zooms around rightmost point (but we're trying to add other way to zoom), in lightweight-charts we zoom around point under the cursor.

ws420:
The thread you linked doesn't provide a solution to fix the issue... it says to make a proposal?

timocov:
> The thread you linked doesn't provide a solution to fix the issue

That issue is pretty similar to what you've requested here, so this issue is duplicate actually.

> it says to make a proposal?

Yeah, it means that implementation might be unclear and we need to make design of the changes first.

> So there is no way to get charts with zooming like on tradingview.com?

Currently no and we didn't have a plan to make it here. Let's re-open it and wait for more feedback.

SlicedSilver:
I believe that this issue is resolved with these API options:
- https://tradingview.github.io/lightweight-charts/docs/next/api/interfaces/TimeScaleOptions#fixrightedge
- https://tradingview.github.io/lightweight-charts/docs/next/api/interfaces/TimeScaleOptions#rightbarstaysonscroll

PR #601, opened 2020-10-25
adds fixRightEdge option

### LINKED ISSUE #218: Add option to fix right edge
Is there a way to fix the right edge? Similar to what you achieve with fixLeftEdge but on the right side.
Or even better, fix both edges?!?

sathoro:
**Type of PR:** enhancement

Fixes #218

**Overview of change:**

Adds `fixRightEdge` option to the timescale as requested in #268 and #218.

I'm using it in conjuction with the `fixLeftEdge` option to prevent zooming the chart out and scrolling the chart into the future.

### DIFF
--- a/src/api/options/time-scale-options-defaults.ts
+++ b/src/api/options/time-scale-options-defaults.ts
@@ -4,6 +4,7 @@ export const timeScaleOptionsDefaults: TimeScaleOptions = {
 	rightOffset: 0,
 	barSpacing: 6,
 	fixLeftEdge: false,
+	fixRightEdge: false,
 	lockVisibleTimeRangeOnResize: false,
 	rightBarStaysOnScroll: false,
 	borderVisible: true,
--- a/src/model/time-scale.ts
+++ b/src/model/time-scale.ts
@@ -69,6 +69,7 @@ export interface TimeScaleOptions {
 	rightOffset: number;
 	barSpacing: number;
 	fixLeftEdge: boolean;
+	fixRightEdge: boolean;
 	lockVisibleTimeRangeOnResize: boolean;
 	rightBarStaysOnScroll: boolean;
 	borderVisible: boolean;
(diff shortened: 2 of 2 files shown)


src/api/options/time-scale-options-defaults.ts
```diff
@@ -4,6 +4,7 @@ export const timeScaleOptionsDefaults: TimeScaleOptions = {
 	rightOffset: 0,
 	barSpacing: 6,
 	fixLeftEdge: false,
+	fixRightEdge: false,
```
timocov, line 7: Would you mind to add documentation for this new option please?

src/model/time-scale.ts
```diff
@@ -420,6 +421,16 @@ export class TimeScale {
 		// zoom in/out bar spacing
 		this.setBarSpacing(newBarSpacing);
 
+		if (this._options.fixRightEdge) {
+			const newRightOffset = this.rightOffset();
+
+			// if we will get pushed into the future then revert the bar spacing
+			if (newRightOffset > 0) {
+				this.setBarSpacing(barSpacing);
```
timocov, line 429: I'm not sure why we should have this code here. It seems that we do here exactly what `_maxRightOffset` is for? It allows to override max value for right offset, so you don't need to handle it yourself in any other places. Did you try to change it according this new option?

timocov (changes requested):
Hi, thanks for your PR.

1. It seems that it requires to add applying an option in `TimeScale::applyOptions` method (in the same way as `fixLeftEdge` or similar)
1. Would you mind to add test cases for this brand new option please? You can use test cases for `fixLeftEdge` as a baseline for your test cases (find for `fixLeftEdge` in e2e tests).
1. It seems that we can override `_maxRightOffset` to handle this option in all cases?

timocov:
> Adds fixRightEdge option to the timescale as requested in #268 and #218.

Actually #268 is about different issue.

timocov:
Hey @sathoro, do you have any updates here?

timocov:
I'm going to close the PR. If you'll have any updates - feel free to re-open it or just ping me.

PR #603, opened 2020-10-26
597: Fixed drawing single-pointed

### LINKED ISSUE #597: One-point line series is invisible
Hello.

If the data consists of a single value it is not displayed on the chart.

Снимок экрана от 2020-10-21 16-54-17

you need to search for this point with the mouse.

Снимок экрана от 2020-10-21 16-54-20

how do I make this one point visible?

Help please.
Thanks.

eugene-korobko:
**Type of PR:** 

**PR checklist:**

- [x] Addresses an existing issue: fixes #597
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**
Processed special case - 1 bar - while drawing line or area


src/renderers/line-renderer.ts
```diff
@@ -40,7 +42,14 @@ export class PaneRendererLine extends ScaledRenderer {
 		ctx.lineJoin = 'miter';
 
 		ctx.beginPath();
-		walkLine(ctx, this._data.items, this._data.lineType, this._data.visibleRange);
+		if (this._data.items.length === 1) {
```
timocov, line 45: I think it'd be better to add a test case where data has more than 1 item, but visible range is for 1 point:

- 2 series
- the first one has data items at 0 and 2 time
- the second one has data item at 1 time

2 test cases:
- set visible range from 0 to 1
- set visible range from 1 to 2

timocov (changes requested):
See review comment

src/renderers/line-renderer.ts
eugene-korobko: Added three testcases:
The only visible point is the first bar
The only visible point is the last bar
There is no visible bars, however there are bars before and after the visible range

All three cases for both line and area.

These tests are green now and the picture is correct. So the condition to  compare data items length, not visible ones, is correct

timocov (approved):

PR #610, opened 2020-11-01
Update package.json

ronaldwerd:
I think you have a typing mistake regarding tsc


timocov:
Actually it's not, `ttsc` is stands for `ttypescript` - https://github.com/cevek/ttypescript (like TypeScript, but with it allows use Transformation API via tsconfig). But anyway thanks for your attention on this!

PR #540, opened 2020-08-01
Update .gitignore

mahal200:
**Type of PR:** 

**PR checklist:**

- [ ] Addresses an existing issue: fixes #
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**

**Is there anything you'd like reviewers to focus on?**

PR #613, opened 2020-11-12
Added proper support for ES module exports (with some fixes for NodeJS context)

### LINKED ISSUE #488: Added test to check that the library importable in server-side context
Related issue #446.

In #484 @NekitCorp fixed importing the library in server-side context, but we need to be sure that it won't happen in the future. To avoid this, we need to add a test in CI to check that the library might be imported in nodejs context without any error (just imported, don't need to create chart or whatever).

### LINKED ISSUE #543: DOCS: How to use in SSR context
**Lightweight Charts Version: 3.1.3**

**Steps/code to reproduce:**
Encapsulating lightweight-charts in an external library.

**Actual behavior:**
I have my own library which encapsulates lightweight-charts and when installing it generates this error, I am using rollup to compress the files with a cjs format I cannot understand the conflict.

**Expected behavior:**
Module should be found.

I don't know what I am missing

**Screenshots:**
#### Error
image

#### Rollup config
image

timocov:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #543, fixes #488
- [x] Includes tests
- [x] Documentation update

**Overview of changes:**

This PR includes a test in CI to make sure that you're able to import `lightweight-charts` package in NodeJS context, `exports` map in `package.json` (including different files for dev/prod builds 🎉 - now it should work out of the box almost for everybody), the published `package.json` file will have `"type": "module"`.


cipriancaba:
Hey @timocov Do you have any updates on this?

timocov:
@cipriancaba not yet, we're waiting for changes in `fancy-canvas` library so it will be able to be imported in SSR. cc @Nipheris

cipriancaba:
Appreciate the update

zouxuoz:
@timocov https://github.com/tradingview/fancy-canvas/pull/7 merged, can we merge this PR too?

timocov:
@zouxuoz this PR leads for breaking change in the package so we cannot merge right now since we're preparing the next minor update for the library. I think after this version we can start to prepare the next major version with all breaking change we're going to do.

zouxuoz:
@timocov do you have ETA for it? 😔 can I help?

Zidious:
Hey @timocov,

Hope all is well, 

I saw your comment here: https://stackoverflow.com/a/65230044 in regards to the below error in a nodeJS context: 

```console
import { bindToDevicePixelRatio } from 'fancy-canvas/coordinate-space';
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

I was wondering if there is any update to this at all? Specifically, being able to create charts and save image to file. 

Thank you in advance, 

Gabe

timocov:
@Zidious we're planning to fix the issue in v4. We didn't do it before because it might be a breaking change for users.

timocov:
It seems that in fancy-canvas we have a bug that it's impossible to import it in modules.

To merge this branch we're waiting for:
- a fix for the issue above
- upgrading fancy-canvas to the next major version

timocov:
I believe this PR is ready to be rebased and updated accordingly the latest changes regarding the docs (but I have no access to the branch anymore so fyi lwc maintainers).

SlicedSilver:
### FYI: Results of running [`publint`](https://publint.dev).
```bash
npm run prepare-release
npx publint
```

#### Master
```
lightweight-charts lint results:
Suggestions:
1. pkg.module is used to output ESM, but pkg.exports is not defined. As NodeJS doesn't read pkg.module, the ESM output may be skipped. Consider adding pkg.exports to export the ESM output. pkg.module can usually be removed alongside too. (This will be a breaking change)
Warnings:
1. /dist/lightweight-charts.esm.production.js is written in ESM, but is interpreted as CJS. Consider using the .mjs extension, e.g. /dist/lightweight-charts.esm.production.mjs
2. /dist/lightweight-charts.esm.development.js is written in ESM, but is interpreted as CJS. Consider using the .mjs extension, e.g. /dist/lightweight-charts.esm.development.mjs
```

#### This PR
```
lightweight-charts lint results:
All good!
```

🎉

romfrancois (approved):

BUILDING.md
```diff
@@ -44,6 +44,7 @@ There are several included e2e tests available which can be run individually. Pl
   Note that at this step the website cannot work since it uses unpublished so far version. It will be fixed in the next steps.
 1. Create a git tag for this version with the format `vMAJ.MIN.PATCH` (see other tags).
 1. Run `npm run prepare-release` in the root folder.
+1. Run `npx publint` and ensure that there aren't any issues with the generated `package.json`.
```
timocov, line 47: Maybe it is worth it to add this to validate in CI?

package.json
```diff
@@ -13,11 +13,23 @@
     "url": "https://github.com/tradingview/lightweight-charts.git"
   },
   "module": "dist/lightweight-charts.esm.production.js",
-  "main": "index.js",
   "typings": "dist/typings.d.ts",
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "development": {
```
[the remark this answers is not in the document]
timocov: Btw does it work well with all modern bundlers out there? To make sure they all work in the same way (and handle this convention equally).

website/plugins/enhanced-codeblock/theme/CodeBlock/import-lightweight-charts-version.ts
```diff
@@ -2,7 +2,7 @@
 import type { Version } from '../../../../versions';
 
 export type LightweightChartsApi38 = typeof import('lightweight-charts-3.8');
-export type LightweightChartsApiCurrent = typeof import('../../../../..');
+export type LightweightChartsApiCurrent = typeof import('../../../../../dist/lightweight-charts.esm.development');
```
timocov, line 5: Does it have proper types tho? I'd expect it to be resolved into any as typings are located in a different file?

SlicedSilver: Resolved now
Screenshot 2023-01-16 at 13 58 21

BUILDING.md
SlicedSilver: Didn't add this to the CI yet because `npx publint` doesn't exit with an error when there are only `warnings` or `suggestions`.

package.json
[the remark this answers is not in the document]
Nipheris: @SlicedSilver Couldn't find any clear statement about importing `production` or `development` subpath automagically depending on `NODE_ENV` value (if you mean that). It looks like it's just a community convention.
That means everyone who relies on `NODE_ENV` and `index.js` when using current (`3.x.x`) public release should switch to subpaths instead. So, this is a breaking change.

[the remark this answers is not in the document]
Nipheris: @SlicedSilver Migration guide sounds like a better solution in the long run. But we should check that most of the tools in frontend ecosystem are configurable to support `development`/`production` conditions (or have some built-in support).

What I have found so far:
- Node.js: [`--conditions`](https://nodejs.org/api/packages.html#resolving-user-conditions) CLI flag, stable [since 14.x](https://nodejs.org/docs/latest-v14.x/api/packages.html#packages_resolving_user_conditions);
- Rollup: [`exportConditions`](https://github.com/rollup/plugins/blob/c2afcee39e7edcd5291664f6218abd7041911aee/packages/node-resolve/README.md#exportconditions) config options, [PR #540](https://github.com/rollup/plugins/pull/540), since [@rollup/plugin-node-resolve@11.0.0](https://www.npmjs.com/package/@rollup/plugin-node-resolve/v/11.0.0);
- Webpack: [builtin support since 5](https://webpack.js.org/guides/package-exports/#optimizations)?

More to check: Vite, esbuild, Parcel (does Parcel [support exports at all](https://github.com/parcel-bundler/parcel/issues/4155)?).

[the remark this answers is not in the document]
SlicedSilver: @Nipheris I'm unlikely to have time today, but I would like to create a few dummy projects to test the library imports for:
- CJS
- ESM
- Development / Production env vars
- Rollup
- Webpack
- Parcel

Then we will know for sure if there are any problems.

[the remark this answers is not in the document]
timocov: I'd add https://nextjs.org/ to the list as well as esbuild and vite 😉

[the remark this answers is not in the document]
SlicedSilver: NextJS could also be nice to include here: https://tradingview.github.io/lightweight-charts/tutorials#framework-integrations (so maybe two birds with one stone)
Vite uses rollup so hopefully that means we can skip it, but I've actually already got a Vite-LWC example so why not?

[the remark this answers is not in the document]
Nipheris: Btw, if we stop using `NODE_ENV` variable to choose the build configuration (prod vs dev), it seems useful to invent a different approach for providing build-time constants (see [`_build-time-constants`](https://github.com/tradingview/lightweight-charts/blob/86fbcbf028fa45663d1958ee0324fac5206dee97/src/typings/_build-time-constants/index.d.ts) and [`rollup.config.js`](https://github.com/tradingview/lightweight-charts/blob/86fbcbf028fa45663d1958ee0324fac5206dee97/rollup.config.js#L42)). Current setup should remain working (thanks to black magic provided by `@rollup/plugin-replace`), but it becomes misleading to use `process.env.NODE_ENV` just as a build time replacement string 😆 

For example, we could try [`rollup-plugin-consts`](https://github.com/NotWoods/rollup-plugin-consts) to provide such build-time constants.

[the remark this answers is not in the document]
Nipheris: And pass some custom CLI keys to rollup to choose the configuration, like [here](https://rollupjs.org/guide/en/#configuration-files) (search for `configDebug` to find a snippet), instead of calling `cross-env NODE_ENV=production npm run build`.

SlicedSilver:
@Nipheris I've added back `index.cjs`.
It's specified as the `main` key on the package.json as a fallback for older versions of node, and under the `"require"` sections of `"exports"`.
This prevent the PR being a breaking change for users of `require`.

SlicedSilver:
- Added `cjs` builds (thanks @Nipheris), and added a standalone esm version.

The standalone esm version would allows users to use the import syntax directly in the browser without a build step.
**For example:**
```html
<script type="module">
    import { createChart } from './dist/lightweight-charts.standalone-esm.production.js';
    // or import { createChart } from 'https://unpkg.com/lightweight-charts/dist/lightweight-charts.standalone-esm.production.js';

    var chart = createChart(document.getElementById('container'));
   // ...
</script>
```

---

- Removed `"type": "module"` from the generated `package.json` because we are using the `cjs` as the `"main"` entry point. Older versions of node and bundlers will use this as the fallback while the newer versions will use the values defined in the `exports` field.

timocov:
> Older versions of node and bundlers will use this as the fallback while the newer versions will use the values defined in the exports field.

Tbh from my perspective I don't think you need to even handle this case - nodejs supports modules since v16 (right?) which is current LTS (nodejs 14 will shut down in 2 months https://nodejs.dev/en/about/releases/), all modern bundles should support exports map or some sort of. Looks like unnecessary work for you :) But it is up to you ofc 😉

SlicedSilver:
> > Older versions of node and bundlers will use this as the fallback while the newer versions will use the values defined in the exports field.
> 
> Tbh from my perspective I don't think you need to even handle this case - nodejs supports modules since v16 (right?) which is current LTS (nodejs 14 will shut down in 2 months https://nodejs.dev/en/about/releases/), all modern bundles should support exports map or some sort of. Looks like unnecessary work for you :) But it is up to you ofc 😉

In this case, it was very little effort to add the `cjs` support and there doesn't appear to any downside to providing the additional builds. The main concern was whether any of these changes would cause an unnecessary breaking change so we are perhaps erring on the side of caution here.

Whilst we do specify node 16 in the `"engines"` section of the `package.json`, it is advisory only and will only produce warnings so it is likely that some of the users of this package are actually on older versions of node and having this fallback might be a 'nice' thing and potentially allow more people to use the library (and maybe result in few issues raised). 

Obviously we value your opinion and if you think it would be better to exclude these builds and fallback then we take another look at this.

timocov:
> Whilst we do specify node 16 in the "engines" section of the package.json, it is advisory only and will only produce warnings 

Afaik it is removed from published `package.json` file as the main purpose of `lightweight-charts` package is (was?) to be bundled or shipped for browsers i.e. it doesn't matter what nodejs version you use to build your app. But having `engines` in `package.json` in the repo specifies what nodejs should be used to build and ship this `lightweight-charts` package as there are dependencies with specific requirements and scripts that are using version-specific API (like in tests).

> whether any of these changes would cause an unnecessary breaking change 

Adding `exports` in `package.json` is a breaking change already as it will affect modules resolution in consumer's builds. My original idea for this PR "to make it right once" and not to maintain old redundant stuff (like `module` field which is not even part of standard) and forget about these bad days 😂 I noticed that some time ago lots of packages started to migrate from "commonjs only" or "dual esm+cjs" publishing towards "esm only" and it seems to work for them and probably it could help the whole ecosystem to migrate to esm quickly. I cannot provide examples of big enough packages (or even "widely used" ones) that are esm-only now, but you can take a look at https://github.com/ai/nanoid for instance.

Another thought was that this package was (still is?) unusable in nodejs context at all as it requires lots of browser-specific API. The only purpose of having it passing "being imported in nodejs" is to make SSR happy that is widely used by some frameworks. But they cannot use this package now in SSR because of issues anyway. And _most likely_ these frameworks require modern-enough nodejs for work anyway or _process imported files differently without relying on pure nodejs require/import mechanism_ (e.g. as they need to handle imports of css modules as well as others). And keep in mind that nodejs 14 is LTS for 2 months since now and I assume maintainers of other packages/frameworks might start to migrate their min nodejs versions to 16.

But obviously it requires having an understanding of your customers and whether you can or cannot drop cjs support for them in the next major release, which I cannot have unlike you.

> having this fallback might be a 'nice' thing and potentially allow more people to use the library (and maybe result in few issues raised).

Yes, this is very fair point 👍

---

@SlicedSilver And please don't consider any of my words as nothing more as just an opinion of another random from the Internet. I'm not a maintainer anymore and out of context for a long time and I'm not in position to tell you what to do. And obviously I don't want you to change your mind 🙂 At the end you'll be maintaining this project after any merged changes 😂 

At this point I'm shutting up in this thread 😂

.size-limit.js
```diff
@@ -1,11 +1,21 @@
...
+	},
 	{
 		name: 'ESM',
-		path: 'dist/lightweight-charts.esm.production.js',
+		path: 'dist/lightweight-charts.production.mjs',
 		limit: '44.0 KB',
 	},
+	{
+		name: 'Standalone-ESM',
+		path: 'dist/lightweight-charts.standalone-esm.production.js',
```
Nipheris, line 16: Why not `lightweight-charts.standalone.production.mjs` for this case? Sounds more straightforward imho.

SlicedSilver: These imports would be used directly in the browser without any build step. My feeling is that `.js` is a better file extension to use in a non-node environment. Whilst `.mjs` should work fine in browsers, it is important that the server hosting the files provides the correct MIME type. This is more likely to do done correctly by the server automatically with `.js`.

It's mostly a preference thing so happy to change it.

Nipheris: Very good point about MIME types 🤔 . Let's check unpkg. For the link https://unpkg.com/fancy-canvas@2.1.0/index.mjs it sends the following Content-Type response header:
```
content-type: application/javascript; charset=utf-8
```
Also `mjs` is listed in [Common MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types), so it sounds as common one nowadays.

And obviously we would be more explicit about the content and module type if we choose `mjs`) Why invent another naming convention if there is the official one? Especially in the SAME library, where we already have `mjs`)

SlicedSilver: Good points, I'll change it to `mjs` 👍

Nipheris: [The default MIME types for nginx 1.23.3](https://github.com/nginx/nginx/blob/release-1.23.3/conf/mime.types) - only `js` is in the list 😔 , but there is a [ticket](https://trac.nginx.org/nginx/ticket/2216) already. And Apache guys [have already implemented](https://github.com/apache/httpd/pull/318) support for `mjs` and have served the improvements in [2.5.1 release](https://github.com/apache/httpd/blob/trunk/CHANGES#L26). So, the support is one the way already)

Nipheris (approved):
Looks good to me! Maybe we could also improve `Installing` section in `README.md`, to be clear about available alternatives. To summarize, now we have the following:
| | |ES module|CommonJS|IIFE (`window.LightweightCharts`)|
|-|-|-|-|-|
|Deps NOT included|PROD|`production.mjs`|`production.cjs`|N/A|
|Deps NOT included|DEV|`development.mjs`|`development.cjs`|N/A|
|Deps included (standalone)|PROD|`standalone.production.mjs`|-|`standalone.production.js`|
|Deps included (standalone)|DEV|`standalone.development.mjs`|-|`standalone.development.js`|

PR #675, opened 2021-01-11
Fix bars renderer (#673)

### LINKED ISSUE #673: Incorrect bar height when its value is more than chart's height
https://github.com/tradingview/lightweight-charts/blob/5e6327f6acaa157a4a3889dae7f71920fa284696/src/renderers/bars-renderer.ts#L72

I think there is a typo on this line.
You should use **bar.highY** instead of **bar.high**

The current implementation does not work correctly on data series when values are higher than 200.
This can be reprodused using the example code on [JSFiddle](https://jsfiddle.net/TradingView/v0p52anL/) by changing the data in the series to values greater than 200, something like that:

```
barSeries.setData ([
{time: '2018-10-19', open: 280.34, high: 280.99, low: 278.57, close: 279.85},
...
```

dubroff:
**Type of PR:** fix typo

**PR checklist:**

- [ ] Addresses an existing issue: fixes #673

### DIFF
--- a/src/renderers/bars-renderer.ts
+++ b/src/renderers/bars-renderer.ts
@@ -69,7 +69,7 @@ export class PaneRendererBars implements IPaneRenderer {
 			const bodyRight = bodyLeft + bodyWidth - 1;
 
 			const high = Math.min(bar.highY, bar.lowY);
-			const low = Math.max(bar.high, bar.lowY);
+			const low = Math.max(bar.highY, bar.lowY);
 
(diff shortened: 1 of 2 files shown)


timocov:
Thanks for prompt fix!

Can you please add a test case for this issue please? See https://github.com/tradingview/lightweight-charts/blob/master/tests/e2e/graphics/README.md for docs and https://github.com/tradingview/lightweight-charts/blob/master/tests/e2e/graphics/test-cases/series/bar.js for sample test case (I think you might either adopt this test case or just copy-paste and adopt the new one).

Issue #140, opened 2019-07-21
Ability to fill area between line series


bmatusiak:
Is it possible to have a  "Area" with a range to color between 2 points

like this

![image](https://i.imgur.com/DwItcWa.png)

bmatusiak:
so after thinking a little harder,     there would be 2 more options needed    `colorHigh` & `colorLow`
im guessing API would be similar or integrated into AreaSeries

---------------------
think of the value as
```
{
value : [2,1]
}
```
first point is HIGH so `colorHigh`   would be the fill color

---------------------
```
{
value : [1,2]
}
```
first point is LOW so `colorLow`   would be the fill color

---------------------
and if value are the same,, there is no area to color between

bmatusiak:
here is a fiddle for an example case,,
https://jsfiddle.net/vfcped0z/27/
the goal for the fiddle is to color inside the triangles
using blue as a baseline,  the top triangle should be a different color fill then below

putraxor:
any progress on this?

timocov:
@putraxor No. If you need this feature, you can try to provide proposal here how it could be implemented from API side (not internally of course) to cover all cases (or almost all). I hope it can be our start point of the discussion.

s2x:
> here is a fiddle for an example case,,
> https://jsfiddle.net/vfcped0z/27/
> the goal for the fiddle is to color inside the triangles
> using blue as a baseline, the top triangle should be a different color fill then below

I realize that this is not the best solution - but currently I use candles to generate the kumo cloud in my projects. An example below in the link:

https://jsfiddle.net/xwvs1gb4/

rudin-kr:
Is this only the way to fill between two lines?

IndiluStudio:
how to do it any idea with kotlin?

arunpresdo:
This would be pretty useful. Any chance this will be implemented soon?

edew:
We're happy to add something like this but we don't have any immediate plan to implement it.



> @putraxor No. If you need this feature, you can try to provide proposal here how it could be implemented from API side (not internally of course) to cover all cases (or almost all). I hope it can be our start point of the discussion.

Likewise if you have any suggestion about how the public API for this could look then please feel free to suggest it.

ziofat:
A reasonable solution is to change baseline series to support dynamic baseline values.

UweR70:
Based on 
https://jsfiddle.net/vfcped0z/27/
I´ve made minimal changes 
and created
[...]

PR #696, opened 2021-02-08
Cloud Area Chart

mastermobin:
Nowadays, Ichimoku Cloud is an inseparable part of our technical analyses, so I implemented it for myself in last 2 days.

I think this PR could be useful for many people using this precious library.

Thank you.


mastermobin:
Are there problems for merging this feature? (Any bugs or something like that, which I could fix) :thinking:

timocov:
Hi @mastermobin, sorry for late reply.

First of all, thanks for your interest and contribution in lightweight-charts.

I do understand why and where it might be useful, but I'm not sure that it should be done in this way. This is exactly why we put "Outside contributors may implement enhancement/features only after an approval (label "help wanted") by Lightweight Charts project maintainers" in our [contribution guide](https://github.com/tradingview/lightweight-charts/blob/master/CONTRIBUTING.md#pull-requests). The closest feature request for this PR it seems is #140 which has a "need proposal" label what means that we need to discuss API first to be sure that everything is covered.

I don't think that we need to create another series type for that (#140). _Perhaps_ that it might be done more easier via adding something like "series linking" which allows you set 2 series to link to and colors rather than adding a new series type.

If you'd like to continue with this feature, I'd suggest to continue a discussion (to start from the beginning 🙂) in the issue or create a new GitHub Discussion to chat about this feature's API (probably some part of your code might be applied as-is with the new approach).

What do you think?

mastermobin:
I'm in :D. 
I'm Ok with either continuing discussion in the issue or creating a new Github discussion.

timocov:
@mastermobin I'm sorry for late reply. I just created a discussion https://github.com/tradingview/lightweight-charts/discussions/714.

PR #755, opened 2021-06-09
Fix #752 Series data still in memory after using setdata

### LINKED ISSUE #752: Removing data for the last series doesn't actually remove the data

**Steps/code to reproduce:**
Take 2 lines series : serie 1 and serie 2 both line type series of a couple hundred points.

1. Draw serie 1 `serie1.setData([... data points serie 1 ...])`
2. Draw serie 2 `serie2.setData([... data points serie 2 ...])`
3. Clear serie 1 `serie1.setData([])`
4. Clear serie 2 `serie2.setData([])`
5. Draw serie 1 `serie1.setData([... data points serie 1 ...])`

**Actual behavior:**

Serie 1 appear on chart as expected but serie 2 is rendered too (although I _deleted_ the data using setData([]))

**Expected behavior:**

Only serie 1 is drawn as a result of the serie1.setData([... points ...])

https://jsfiddle.net/kjg1zvnx/

Maybe setData([]) isn't made for deleting the data? 

This works :
Instead of calling setData([]) draw a fake point
```
serie1.setData([
"time": 1622933040, // first time from other series, avoid 'invalid date' error on time axis...
"value": undefined
}])
```

edew:
**Type of PR:** Bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #752 
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Currently when a series is remove (i.e. it is updated with empty data `[]`) the library does not properly handle removing that data.

Internally this is because `DataLayer._syncIndexesAndApplyChanges` doesn't include the removed series in the returned `DataUpdateResponse`.

This PR fixes the logic so that removed series are included in the update response.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ debug.html
 
 # graphics tests out data
 /tests/e2e/graphics/.gendata/
+
+# OS X .DS_Store files
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,3 @@ debug.html
 
 # graphics tests out data
 /tests/e2e/graphics/.gendata/
-
-# OS X .DS_Store files
-.DS_Store
--- a/src/api/data-layer.ts
+++ b/src/api/data-layer.ts
@@ -160,6 +160,8 @@ export class DataLayer {
 	// this is kind of "dest" values (in opposite to "source" ones) - we don't need to modify it manually, the only by calling _syncIndexesAndApplyChanges method
 	private _sortedTimePoints: readonly TimeScalePoint[] = [];
 
+	private _previousBaseIndexValue: TimePointIndex = 0 as TimePointIndex;
+
 	public destroy(): void {
 		this._pointDataByTimePoint.clear();
 		this._seriesRowsBySeries.clear();
@@ -384,12 +386,19 @@ export class DataLayer {
 	private _getBaseIndex(): TimePointIndex {
 		let baseIndex = 0 as TimePointIndex;
 
-		this._seriesRowsBySeries.forEach((data: SeriesPlotRow[]) => {
-			if (data.length !== 0) {
-				baseIndex = Math.max(baseIndex, data[data.length - 1].index) as TimePointIndex;
-			}
-		});
+		if (this._seriesRowsBySeries.size === 0) {
+			// all series have been removed so use the previous base index value
+			baseIndex = this._previousBaseIndexValue;
+		} else {
+			// calculate a new base index value from series data
+			this._seriesRowsBySeries.forEach((data: SeriesPlotRow[]) => {
+				if (data.length !== 0) {
+					baseIndex = Math.max(baseIndex, data[data.length - 1].index) as TimePointIndex;
+				}
+			});
+		}
 
+		this._previousBaseIndexValue = baseIndex;
 		return baseIndex;
 	}
 
--- a/src/api/data-layer.ts
+++ b/src/api/data-layer.ts
@@ -335,8 +335,6 @@ export class DataLayer {
 	private _updateTimeScalePoints(newTimePoints: TimeScalePoint[]): number {
 		let firstChangedPointIndex = -1;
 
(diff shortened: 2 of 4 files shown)


tests/unittests/data-layer.spec.ts
```diff
@@ -695,4 +712,61 @@ describe('DataLayer', () => {
...
+		const series2 = createSeriesMock();
+
+		const data1 = generateData();
+		const data2 = generateData();
+
+		dataLayer.setSeriesData(series1, data1);
+		dataLayer.setSeriesData(series2, data2);
+
+		it('update result has empty data for series1', () => {
+			const updateResult = dataLayer.setSeriesData(series1, []);
```
timocov, line 741: what if tests will be run in random order? Probably now the result depends on order of tests and probably this is not good. Can we either merge all `it`s into one or make tests independent of each other?

src/api/data-layer.ts
```diff
@@ -420,6 +420,10 @@ export class DataLayer {
 				dataUpdateResponse.series.set(s, { data, fullUpdate: true });
 			});
 
+			if (!this._seriesRowsBySeries.has(series)) {
+				dataUpdateResponse.series.set(series, { data: [], fullUpdate: true });
```
timocov, line 424: would you mind to leave a comment here why do we need this here please?

.gitignore
```diff
@@ -26,3 +26,6 @@ debug.html
 
 # graphics tests out data
 /tests/e2e/graphics/.gendata/
+
+# OS X .DS_Store files
+.DS_Store
```
timocov, line 31: hm, probably it should be system-wide ignore 😂 I'm just wondering why we have an ignore item for `*.pyc` files, but I think that ignoring OS-specific files shouldn't be a project-related thing (in this case we'd need to add some windows-specific files, linux-specific etc)

timocov (changes requested):
The fix looks good for me, a few comments regarding tests/other changed files. btw, would you mind to add graphics tests as well?

tests/unittests/data-layer.spec.ts
```diff
@@ -411,6 +418,56 @@ describe('DataLayer', () => {
...
+		const series1 = createSeriesMock();
+		const series2 = createSeriesMock();
+		const data1 = generateData();
+		const data2 = generateData();
+		dataLayer.setSeriesData(series1, data1);
+		dataLayer.setSeriesData(series2, data2);
+
+		const updateResult1 = dataLayer.setSeriesData(series1, []);
+
+		expect(updateResult1.timeScale.baseIndex).to.be.equal(0, 'expected base index to be 0');
```
timocov, line 444: shouldn't it be 9 here? we had 2 series with 10 data items each and just removed one of them so probably baseIndex value shouldn't be changed after removal?

tests/unittests/data-layer.spec.ts
```diff
@@ -411,6 +418,56 @@ describe('DataLayer', () => {
...
+		const updateResult2 = dataLayer.setSeriesData(series2, []);
+
+		expect(updateResult2.timeScale.baseIndex).to.be.equal(0, 'expected base index to be 0');
+		expect(updateResult2.timeScale.points?.length).to.be.equal(0, 'expected updated time scale points length to equal 0');
+		expect(updateResult2.series.has(series2)).to.be.equal(true, 'expected to contain series2');
+		expect(updateResult2.series.get(series2)?.data.length).to.be.equal(0, 'expected series2 data to be empty');
+
+		const updateResult3 = dataLayer.setSeriesData(series1, data1);
+
+		expect(updateResult3.timeScale.baseIndex).to.be.equal(0, 'expected base index to be 0');
```
timocov, line 458: the same here about baseIndex value

tests/unittests/data-layer.spec.ts
edew: 👍 This is because the test data has invalid `time`s, so the index of all data points is `0`.

timocov (approved):

PR #760, opened 2021-06-11
Fix #750 candlesticks colour options update

### LINKED ISSUE #750: Candlestick upColor and downColor is not changed on applyOptions

**Steps/code to reproduce:**

```ts
import { BarData, createChart } from "lightweight-charts";

const chartRoot = document.getElementById("main-chart") as HTMLDivElement;
const chart = createChart(chartRoot, {
  width: 800,
  height: 600
});

const candlestick = chart.addCandlestickSeries({
  upColor: "rgb(0,255,0)",
  downColor: "rgb(255,0,0)"
});

const data: BarData[] = [
  {
    time: "2021-06-04",
    open: 1000,
    high: 1200,
    low: 800,
    close: 900
  },
  {
    time: "2021-06-05",
    open: 1000,
    high: 1200,
    low: 800,
    close: 1100
  }
];

candlestick.setData(data);

setTimeout(() => {
  candlestick.applyOptions({
    upColor: "rgb(0,0,255)",
    downColor: "rgb(0,255,0)"
  });
}, 2000);

setTimeout(() => {
  candlestick.applyOptions({
    upColor: "rgb(0,255,255)",
    downColor: "rgb(255,255,0)"
  });
}, 4000);
```

**Actual behavior:**

Candlestick color doesn't change at all

**Expected behavior:**

Candlestick color should change after applyOptions

**Screenshots:**

image
After last applyOptions, candlestick color still not changed, it should be yellow down and light blue up

edew:
**Type of PR:** Bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #750 
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Update the series' pane view when options are applied.

This seems to fix the issue but I'm not sure if this will be bad for performance in some way? (or, maybe we can assume that options aren't applied very often and it's OK to do a little more work here?)

### DIFF
--- a/src/views/pane/candlesticks-pane-view.ts
+++ b/src/views/pane/candlesticks-pane-view.ts
@@ -34,6 +34,22 @@ export class SeriesCandlesticksPaneView extends BarsPaneViewBase<'Candlestick', 
 		return this._renderer;
 	}
 
+	protected _updatePoints(): void {
+		super._updatePoints();
+
+		if (this._itemsVisibleRange === null) {
+			return;
+		}
+
+		for (let i = this._itemsVisibleRange.from; i < this._itemsVisibleRange.to; i++) {
+			const x = this._items[i];
+			const style = this._series.barColorer().barStyle(x.time);
+			x.color = style.barColor;
+			x.wickColor = style.barWickColor;
+			x.borderColor = style.barBorderColor;
+		}
+	}
+
 	protected _createRawItem(time: TimePointIndex, bar: SeriesPlotRow, colorer: SeriesBarColorer): CandlestickItem {
 		const style = colorer.barStyle(time);
 		return {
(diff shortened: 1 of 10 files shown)


src/views/pane/candlesticks-pane-view.ts
```diff
@@ -34,6 +34,22 @@ export class SeriesCandlesticksPaneView extends BarsPaneViewBase<'Candlestick',
 		return this._renderer;
 	}
 
+	protected _updatePoints(): void {
+		super._updatePoints();
```
timocov, line 38: this method is calling on every paint event basically (on every non-"light" invalidation), which might affect the performance more than invalidating everything once on change settings (which we had previously). Can't we do that only on options change for all data at once?

timocov (changes requested):

tests/e2e/graphics/test-cases/applying-options/change-candlestick-colors.js
```diff
@@ -0,0 +1,74 @@
...
+
+function runTestCase(container) {
+	const chart = LightweightCharts.createChart(container, {
+		timeScale: {
+			barSpacing: 40,
+			timeVisible: true,
+		},
+	});
+
+	const candlestick = chart.addCandlestickSeries({
```
timocov, line 49: can you add a bars series here as well just to make sure that the code works fine now for it please? I think you can use the same data as for candlestick series

timocov (approved):

PR #765, opened 2021-06-16
Add fixRightEdge time scale option

### LINKED ISSUE #218: Add option to fix right edge

edew:
**Type of PR:** Enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #218 
- [x] Includes tests
- [x] Documentation update

**Overview of change:**

This adds a `fixRightEdge` option. Logically it's the same as `fixLeftEdge` but for the right.

These changes include logic to limit time scale bar spacing to `width / number of points` when **both** `fixRightEdge` and `fixLeftEdge` are enabled. This is so that zooming-out is prevented if it would cause either the first or last point to move too far from an edge.

### DIFF
--- a/src/model/time-scale.ts
+++ b/src/model/time-scale.ts
@@ -843,7 +843,10 @@ export class TimeScale {
 	}
 
 	private _doFixRightEdge(): void {
-		this.setRightOffset(0);
+		if (this._rightOffset > 0) {
+			this.setRightOffset(0);
+		}
+
 		this._correctBarSpacing();
 	}
 }
--- a/src/model/time-scale.ts
+++ b/src/model/time-scale.ts
@@ -843,9 +843,7 @@ export class TimeScale {
 	}
 
 	private _doFixRightEdge(): void {
-		if (this._rightOffset > 0) {
-			this.setRightOffset(0);
-		}
+		this._correctOffset();
 
 		this._correctBarSpacing();
 	}
--- a/src/model/time-scale.ts
+++ b/src/model/time-scale.ts
@@ -69,6 +69,7 @@ export interface TimeScaleOptions {
 	barSpacing: number;
 	minBarSpacing: number;
 	fixLeftEdge: boolean;
+	fixRightEdge: boolean;
 	lockVisibleTimeRangeOnResize: boolean;
 	rightBarStaysOnScroll: boolean;
 	borderVisible: boolean;
@@ -135,6 +136,10 @@ export class TimeScale {
 
 		if (this._options.fixLeftEdge) {
 			this._doFixLeftEdge();
+		}
+
+		if (this._options.fixRightEdge) {
+			this._doFixRightEdge();
 		}
 
 		// note that bar spacing should be applied before right offset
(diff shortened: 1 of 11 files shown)


tests/e2e/graphics/test-cases/applying-options/fix-both-edges.js
```diff
@@ -0,0 +1,29 @@
...
+	const mainSeries = chart.addLineSeries();
+	mainSeries.setData(generateData());
+
+	chart.timeScale().applyOptions({
+		fixLeftEdge: true,
+		fixRightEdge: true,
+	});
+
+	return new Promise(resolve => {
+		setTimeout(() => resolve(), 1000);
```
timocov, line 27: what this test case tests exactly?

btw if you down need to wait a time and use async operations in your test case - you can just skip returning a promise and the test runner will handle it correctly, e.g. https://github.com/tradingview/lightweight-charts/blob/master/tests/e2e/graphics/test-cases/months-chart.js

tests/e2e/graphics/test-cases/applying-options/fix-right-edge.js
```diff
@@ -0,0 +1,31 @@
...
+
+	const mainSeries = chart.addLineSeries();
+	mainSeries.setData(generateData());
+
+	chart.timeScale().applyOptions({
+		fixRightEdge: true,
+	});
+
+	return new Promise(resolve => {
+		setTimeout(() => resolve(), 1000);
```
timocov, line 29: shouldn't we set both fixRightEdge and fixLeftEdge here? just because the name of test case is `applying-options/fix-right-edge` so a bit confusing

timocov (changes requested):
I think we need to have the following test cases here:
1. Applying `fixRightEdge` at the start and make sure that it doesn't allow you to scroll to the future
1. Open chart, scroll to the future, apply `fixRightEdge` in a timeout and make sure that it fixes the current position
1. Open chart, scroll to _the past_, apply `fixRightEdge` in a timeout and make sure that it does nothing
1. Open chart, apply `fixRightEdge` in a timeout and make sure that it doesn't allow you to scroll to the future
1. Open chart with enabled `fixRightEdge`, then disable it after some time and then you should be able to scroll to the future
1. Open chart with enabled both `fixLeftEdge` and `fixRightEdge` and it shouldn't allow you to scroll to the past/future
1. Open chart, enable both `fixLeftEdge` and `fixRightEdge` after some time and it shouldn't allow you to scroll to the past/future (also it should correct the position _and_ barSpacing value)
1. Open the chart, zoom in the data so you have a data from the left and from the right edges, enable `

Probably you cover some of cases in your test cases already.

src/model/time-scale.ts
```diff
@@ -819,5 +838,15 @@ export class TimeScale {
 			const leftEdgeOffset = this._rightOffset - delta - 1;
 			this.setRightOffset(leftEdgeOffset);
 		}
+
+		this._correctBarSpacing();
+	}
+
+	private _doFixRightEdge(): void {
+		if (this._rightOffset > 0) {
+			this.setRightOffset(0);
```
timocov, line 847: why not `_correctOffset` method? it looks like it does what we need to? and we wouldn't need to check current value then

edew: I thought that the `_correctOffset` logic would cause the right offset to the wrong value in some cases... but I've tested it just now and it seems to work fine so I will change this 👍 .

timocov (approved):

Issue #795, opened 2021-07-26
Price line label show on both axis


thanhlmm:
**Lightweight Charts Version:** 3.4.0

**Steps/code to reproduce:**

```js
var series = chart.addLineSeries({
  color: "green",
  lineWidth: 2,
  priceScaleId: "left"
});

series.setData(
  data.map((row) => ({
    time: row.time,
    value: row.value * 10
  }))
);

var series2 = chart.addLineSeries({
  color: "red",
  lineWidth: 2,
  priceScaleId: "right"
});

series2.setData(data);

series2.createPriceLine({
  price: 98,
  color: "blue"
});
```

**Actual behavior:**

Price line label show on both axis

**Expected behavior:**

Price line show only on its axis

**Screenshots:**

<img width="514" alt="Screen Shot 2021-07-26 at 17 20 35" src="https://user-images.githubusercontent.com/9281080/126973739-76325b92-961b-4731-baf7-d41ccde0df1c.png">

**CodeSandbox/JSFiddle/etc link:**

https://codesandbox.io/s/price-line-show-on-both-axis-wo98u?file=/src/index.js:16149-16520

timocov:
The issues is [somewhere here](https://github.com/tradingview/lightweight-charts/blob/f26d7ce1592f8c8c4dec7c2736cf8b6bdee662fb/src/model/series.ts#L375-L378). @thanhlmm would you like to provide a fix for that?

thanhlmm:
@timocov I will

PR #799, opened 2021-07-27
Fix price line show on both axis

thanhlmm:
**Type of PR:** Fix bug

**PR checklist:**

- [x] Addresses an existing issue: #795
- [x] Includes tests
- [ ] Documentation update

### DIFF
--- a/src/model/custom-price-line.ts
+++ b/src/model/custom-price-line.ts
@@ -8,7 +8,6 @@ import { IPriceAxisView } from '../views/price-axis/iprice-axis-view';
 
 import { Coordinate } from './coordinate';
 import { PriceLineOptions } from './price-line-options';
-import { PriceScale } from './price-scale';
 import { Series } from './series';
 
 export class CustomPriceLine {
@@ -45,10 +44,6 @@ export class CustomPriceLine {
 
 	public priceAxisView(): IPriceAxisView {
 		return this._priceAxisView;
-	}
-
-	public priceScale(): PriceScale {
-		return this._series.priceScale();
 	}
 
 	public update(): void {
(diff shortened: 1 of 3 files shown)


src/model/custom-price-line.ts
```diff
@@ -46,6 +47,10 @@ export class CustomPriceLine {
 		return this._priceAxisView;
 	}
 
+	public priceScale(): PriceScale {
+		return this._series.priceScale();
```
timocov, line 51: as soon as price line is attached to series' price scale, do we need to have this method? look like we can check the value in series

src/model/series.ts
```diff
@@ -374,7 +374,10 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
 	public priceAxisViews(pane: Pane, priceScale: PriceScale): readonly IPriceAxisView[] {
 		const result = (priceScale === this._priceScale || this._isOverlay()) ? [...this._priceAxisViews] : [];
 		for (const customPriceLine of this._customPriceLines) {
-			result.push(customPriceLine.priceAxisView());
+			if (customPriceLine.priceScale() === priceScale) {
+				// Only show price on its series axis
+				result.push(customPriceLine.priceAxisView());
+			}
 		}
```
timocov, line 381: Looks like in case of `priceScale !== this._priceScale && !this._isOverlay()` we should return an empty array and the current result otherwise?

timocov (changes requested):
Thanks for your contributing! It seems that we can do it even simpler, see comments below.

src/model/series.ts
thanhlmm: The change is updated. I'm still confused by how things getting connected together

PR #800, opened 2021-07-28
Fix series draw in wrong order

### LINKED ISSUE #794: Overlapped line chart show wrong color order when hover
**Lightweight Charts Version:** 3.4.0

**Steps/code to reproduce:**

```js
  color: "green",
  lineWidth: 2
});

series.setData(data);

var series2 = chart.addLineSeries({
  color: "red",
  lineWidth: 2
});

series2.setData(data);
```

**Actual behavior:**

The crosshair when hover show as red but the line show in the chart is green.

**Expected behavior:**

The crosshair when hover should be in green since the green line is above the red line

**Screenshots:**

Screen Shot 2021-07-26 at 16 48 44

https://codesandbox.io/s/hardcore-frost-3pxjn?file=/src/index.js

thanhlmm:
**Type of PR:** Bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #794
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**
Change the way we sort sources

### DIFF
--- a/src/model/sort-sources.ts
+++ b/src/model/sort-sources.ts
@@ -4,6 +4,6 @@ import { IDataSource } from './idata-source';
 
 export function sortSources<T extends IDataSource>(sources: readonly T[]): T[] {
 	return sources.slice().sort((s1: IDataSource, s2: IDataSource) => {
-		return (ensureNotNull(ensureNotNull(s2.zorder())) - ensureNotNull(s1.zorder()));
+		return (ensureNotNull(s1.zorder()) - ensureNotNull(s2.zorder()));
 	});
 }
--- a/src/model/sort-sources.ts
+++ b/src/model/sort-sources.ts
@@ -4,6 +4,6 @@ import { IDataSource } from './idata-source';
 
 export function sortSources<T extends IDataSource>(sources: readonly T[]): T[] {
 	return sources.slice().sort((s1: IDataSource, s2: IDataSource) => {
-		return (ensureNotNull(s1.zorder()) - ensureNotNull(s2.zorder()));
+		return (ensureNotNull(ensureNotNull(s2.zorder())) - ensureNotNull(s1.zorder()));
 	});
 }
--- a/tests/e2e/graphics/test-cases/series/line-overlap.js
+++ b/tests/e2e/graphics/test-cases/series/line-overlap.js
@@ -1,0 +1,31 @@
+function generateData() {
+	const res = [];
+	const time = new Date(Date.UTC(2018, 0, 1, 0, 0, 0, 0));
+	for (let i = 0; i < 500; ++i) {
+		res.push({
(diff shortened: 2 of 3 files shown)


thanhlmm:
@timocov https://github.com/tradingview/lightweight-charts/blob/master/src/model/pane.ts#L169
The logic is that new added source will get a lower zorder (-1, -2,...). It does not feel like normal standard, `zIndex` in css for examples

src/model/sort-sources.ts
```diff
@@ -4,6 +4,6 @@ import { IDataSource } from './idata-source';
 
 export function sortSources<T extends IDataSource>(sources: readonly T[]): T[] {
 	return sources.slice().sort((s1: IDataSource, s2: IDataSource) => {
-		return (ensureNotNull(s1.zorder()) - ensureNotNull(s2.zorder()));
+		return (ensureNotNull(ensureNotNull(s2.zorder())) - ensureNotNull(s1.zorder()));
```
timocov, line 7: It seems that the order will be different now? I'd say that issue is in the code you mentioned in https://github.com/tradingview/lightweight-charts/pull/800#issuecomment-887996141, since it's strange that the new created series are back of previous ones. This change affects the order so basically our zorder will have different meaning and I don't think that this is what we want to have.

I'd say the new series should have `previous max + 1` zorder just to be at top of all previous ones so it might be expected to others who creates the series.

timocov (changes requested):

src/model/sort-sources.ts
thanhlmm: I will update it as `previous max + 1`. I'm not sure that the previous logic is there for any other reason so I make as little change as possible

timocov (approved):

tests/e2e/graphics/test-cases/series/line-overlap.js
```diff
@@ -0,0 +1,31 @@
...
+			time: time.getTime() / 1000,
+			value: i,
+		});
+
+		time.setUTCDate(time.getUTCDate() + 1);
+	}
+	return res;
+}
+
+export function runTestCase(container) {
```
timocov, line 15:
```suggestion
function runTestCase(container) {
```

thanhlmm: In debug, I can import the test case to live test it by doing
```js
<script type="module">
  import { runTestCase } from './tests/e2e/graphics/test-cases/series/line-color-function.js';
  runTestCase(document.getElementById('container'));
</script>
```

timocov: Oh, didn't know that 👍 Our test system isn't ready for that yet so just removed export keyword for now to make it work in CI

timocov:
@thanhlmm thanks for your work! Merged now.

PR #801, opened 2021-07-28
Add chart export background

### LINKED ISSUE #798: Screenshot output missing piece on bottom right
**Lightweight Charts Version:** 3.4.0

**Steps/code to reproduce:**

```js
  color: "green",
  lineWidth: 2,
  priceScaleId: "left"
});

series.setData(
  data.map((row) => ({
    time: row.time,
    value: row.value * 10
  }))
);

var series2 = chart.addLineSeries({
  color: "red",
  lineWidth: 2,
  priceScaleId: "right"
});

series2.setData(data);

series2.createPriceLine({
  price: 98,
  color: "blue"
});

const dataUrl = window.chart.takeScreenshot().toDataURL();
```

**Actual behavior:**

The output image is missing a pieces on bottom right

**Expected behavior:**

The image is filled all by the chart layout.backgroundColor

**Screenshots:**

Screen Shot 2021-07-27 at 20 00 44
(The yellow is background)

https://codesandbox.io/s/screenshot-bug-yyqck?file=/src/index.js

thanhlmm:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #798 
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**
Add chart background when exporting


timocov:
@thanhlmm can you add a test case for that please? Just to be sure that 1) the change fixes the issue 2) we won't break it in the future. I think you can slightly modify this test case https://github.com/tradingview/lightweight-charts/blob/master/tests/e2e/graphics/test-cases/take-screenshot.js.

timocov:
@thanhlmm it seems that the test case works fine, probably something is missed?

thanhlmm:
I added new test case

timocov:
> it seems that the test case works fine, probably something is missed?

Let me explain a bit. e2e tests are screenshots-based tests which are comparing what the library from merge-base commit and from PR branch are generating. Thus, if you fixed something and added e2e test case for that we expect that the tests to be failed as soon as there should be a difference between master branch and your changes. If tests are passed that means that a test doesn't actually check the issue.

thanhlmm:
Updated. Thanks

src/gui/chart-widget.ts
```diff
@@ -289,6 +289,11 @@ export class ChartWidget implements IDestroyable {
 				}
 			}
 		});
+		// Set chart background
+		ctx.globalCompositeOperation = 'destination-over';
+		ctx.fillStyle = this._options.layout.backgroundColor;
+		ctx.fillRect(0, 0, targetCanvas.width, targetCanvas.height);
+		ctx.globalCompositeOperation = 'source-over';
```
timocov, line 296: I don't think that this code is the issue actually. If you look closer to the screenshot you can see that there are still the issue with screenshots and looks like the issue is in different place.

How it is now (in master branch):

image

How it is in this PR:

image

Differences:
image

- looks like a small dot of the right axis still is not presented on the screenshot
- there is some overlapping of time axis and something else so we don't have a border in some area

It seems that the issue is at 286 line, so instead of `targetX = firstPane.getSize().w;` we should do `targetX += firstPane.getSize().w;` to add left pane's width to current coordinate. If you remove this change and only replace `=` with `+=` it will works correctly 🙂 Single character fix 😄

timocov (changes requested):
It seems that the fix is made in wrong place. Look at my comment above

timocov (approved):

timocov:
Thank you for your contribution!

PR #813, opened 2021-08-05
Area with baseline chart

### LINKED ISSUE #151: Baseline series type
Is it possible to fill Area from predefined level (ex. 0)?
If you draw graph with negative values it currently fills area up from minimum value.
It maybe will be more logical to fill are from maximum to minimum.

https://jsfiddle.net/axtm3wr0/

image

thanhlmm:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #151
- [x] Includes tests
- [x] Documentation update

**Overview of change:**
Add new api to add Area with baseline
```
	const mainSeries = chart.addAreaBaselineSeries({
		lineWidth: 1,
		baseValue: {
			type: 'price',
			price: 0,
		},
	});
```


thanhlmm:
It still missing custom style for line top and bottom now. I think it can be postponed

thanhlmm:
@timocov Any thought on this PR?

timocov:
Note for others who might be concerned: we discussed with @thanhlmm in PM this PR and agreed that I'll push some changes to this PR and provide an update what is changed. The main reason of this is that we already have a code for baseline chart on tradingview.com so it would be easy for us to support similar code rather than have several implementation of the same things. We scheduled this feature for 3.7 version.

timocov:
Hi there @thanhlmm,

I just pushed changes to your branch as we discussed previously and here is a list of main changes against yours (luckily no breaking changes in API):

- resolved conflicts with master branch
- updated default values of colors for baseline series
- updated docs
- refactored pane renderers and pane views since baseline looks pretty similar to area with the only change of fill/stroke style, so I extracted that to the base class
- corrected test cases
- removed adding 0.01 to the gradient step for the bottom part (it seems it works fine now, did you have any issue with that?)

Let me know what do you think.

Also, I should say that I didn't change a lot actually so you did really good job. Thanks for your contribution!

timocov (approved):

thanhlmm:
Awesome

PR #815, opened 2021-08-05
Fix fitContent with lockVisibleTimeRangeOnResize and fixed edge

### LINKED ISSUE #814: Fixed both timescale edges  make lockVisibleTimeRangeOnResize turn wrong
**Lightweight Charts Version:** 3.5.0

**Steps/code to reproduce:**

```js
var chart = createChart(container, {
  leftPriceScale: {
    visible: true,
    borderColor: "#EFF2F5",
    drawTicks: false
  },
  timeScale: {
    borderVisible: false,
    fixLeftEdge: true,
    fixRightEdge: true,
    timeVisible: true,
    lockVisibleTimeRangeOnResize: true
  },
  layout: {
    backgroundColor: "#ffffff",
    textColor: "#333"
  },
});

  color: "green",
  lineWidth: 2,
  priceScaleId: "left"
});

series.setData(
  data.map((row) => ({
    time: row.time,
    value: row.value * 10
  }))
);

chart.timeScale().subscribeVisibleLogicalRangeChange((range) => {
  console.log(range);
});

```

**Actual behavior:**
When run `chart.timeScale().fitContent();` => The chart change visible range 2 times, the first time is `{from: 0, to: 299}` and the last one is `{from: 0, to: 297.5623003194888}`

It happend due to: `fitContent` => `visibleRange` change => after change `visibleRange` => width of axis get updated => `visibleRange` change again which not cover the full range
**Expected behavior:**
The final `visibleRange` should still show the full chart, which is `{from: 0, to: 299}`

**Screenshots:**

https://codesandbox.io/s/fitcontent-with-fixed-left-and-right-50n1k?file=/src/index.js:15599-16542

thanhlmm:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #814
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

### DIFF
--- a/src/model/time-scale.ts
+++ b/src/model/time-scale.ts
@@ -261,7 +261,7 @@ export class TimeScale {
 		if (this._options.lockVisibleTimeRangeOnResize && this._width) {
 			// recalculate bar spacing
 			const newBarSpacing = this._barSpacing * width / this._width;
-			this._setBarSpacing(newBarSpacing);
+			this._barSpacing = newBarSpacing;
 		}
 
 		// if time scale is scrolled to the end of data and we have fixed right edge
--- a/tests/e2e/graphics/test-cases/fit-content-with-lot-data.js
+++ b/tests/e2e/graphics/test-cases/fit-content-with-lot-data.js
@@ -1,0 +1,40 @@
+function generateData(valueOffset, count) {
+	const res = [];
+	const time = new Date(Date.UTC(2018, 0, 1, 0, 0, 0, 0));
+	for (let i = 0; i < count; ++i) {
+		res.push({
+			time: time.getTime() / 1000,
+			value: i + valueOffset,
+		});
+
+		time.setUTCDate(time.getUTCDate() + 1);
+	}
+
+	return res;
+}
+
+function runTestCase(container) {
+	const chart = LightweightCharts.createChart(container, {
+		leftPriceScale: {
+			visible: true,
+			drawTicks: false,
+		},
+		timeScale: {
+			borderVisible: false,
+			fixLeftEdge: true,
+			fixRightEdge: true,
+			timeVisible: true,
+			lockVisibleTimeRangeOnResize: true,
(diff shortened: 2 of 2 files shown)


src/model/time-scale.ts
```diff
@@ -261,7 +261,7 @@ export class TimeScale {
 		if (this._options.lockVisibleTimeRangeOnResize && this._width) {
 			// recalculate bar spacing
 			const newBarSpacing = this._barSpacing * width / this._width;
-			this._setBarSpacing(newBarSpacing);
+			this._barSpacing = newBarSpacing;
```
timocov, line 264: Just curious how it helps, because afaik when the full update is happened we do the following:
1. Recreate widgets, *update their sizes* - https://github.com/tradingview/lightweight-charts/blob/263ff2299dbdc363087957f0f5b14917490d986b/src/gui/chart-widget.ts#L445
2. Apply time scale invalidations like `fitContent` - https://github.com/tradingview/lightweight-charts/blob/263ff2299dbdc363087957f0f5b14917490d986b/src/gui/chart-widget.ts#L462

Thus it looks like `fitContent` should be applied after the price scale's widget is changed. Did you look into that?

timocov: @thanhlmm any updates?

thanhlmm: Hi, let me brief what I found:
The flow when we `fitContent()` is
`fitContent()` => chart re-render and the axis render => the axis then check the space and try make the chart re-render again (due to its width is changed) => the chart re-render but the logical range is changed as I describe in https://github.com/tradingview/lightweight-charts/issues/814

So I figured out that `this._setBarSpacing(newBarSpacing);` flow has **some condition and it make the logical range smaller**. So I just set it and let the 2nd update do the job

Let me know if there is a better way to do it

timocov (approved):

PR #829, opened 2021-09-01
Set visible range with percent right margin

kirchet:
**Type of PR:** enhancement

**PR checklist:**

- [x] Includes tests
- [x] Documentation update

**Overview of change:**

```js
chart.timeScale().setVisibleRange({ from, to }, { percentRightMargin: 10 });
```

### DIFF
--- a/docs/time-scale.md
+++ b/docs/time-scale.md
@@ -102,13 +102,24 @@ chart.timeScale().getVisibleRange();
 
 ### setVisibleRange()
 
-Sets visible time range of the chart. The argument is an object with the first and last time points of a desired time range.
+Sets visible time range of the chart. It has the following arguments:
+
+`range` - an object with the first and last time points of a desired time range.
+
+`options` - an optional object that for now has only one field:
+
+* `percentRightMargin`: indicates whether the library should apply the percent right margin to the right border if it points on the last bar.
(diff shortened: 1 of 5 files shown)


timocov:
@DTupalov do you really need this API? It seems that for logical ranges it might be done externally from your code so I just curious whether this API is needed for someone.

DTupalov:
> @DTupalov do you really need this API? It seems that for logical ranges it might be done externally from your code so I just curious whether this API is needed for someone.

@timocov I've realised such ability with logical range, but this need some calculations on my side, so if this would be just a part of library, would be more declarative in my code.

Why then you implement this API? :)

timocov:
> Why then you implement this API? :)

Ha-ha, fair enough 😄 

> I've realised such ability with logical range, but this need some calculations on my side, so if this would be just a part of library, would be more declarative in my code.

We realised exactly the same, that's why I'm a bit confused with having this in the library (just to avoid having anything what could be done outside in 2-3 lines).

maxtrem271991:
chart.timeScale().setVisibleRange({ from, to }, { percentRightMargin: 10 });

SlicedSilver:
Agreed internally that it is no longer required.

PR #849, opened 2021-09-20
Add methods for getting bar info by logical index

tv-ksi:
**Type of PR:** enhancement

**PR checklist:**

- [x] Includes tests
- [x] Documentation update

**Overview of change:**

Add method `getBarInfoByLogicalIndex` to series API  for getting bar price with change value
Add optional `logiacalIndex` param to mouse event handlers

### DIFF
--- a/docs/constants.md
+++ b/docs/constants.md
@@ -101,6 +101,6 @@ It has the following values:
 
 `PlotRowSearchMode` enums is used to specify plot by index searching mode
 
-- `PlotRowSearchMode.Exact` - search a plot exact at index
-- `PlotRowSearchMode.NearestLeft` - search the nearest plot from left side
-- `PlotRowSearchMode.NearestRight` - search the nearest plot from right side
+- `PlotRowSearchMode.Exact` - search a plot at the index
+- `PlotRowSearchMode.NearestLeft` - search a plot at the index, if it is not exist search for the nearest plot from left side from the index
+- `PlotRowSearchMode.NearestRight` - search a plot at the index, if it is not exist search for the nearest plot from right side from the index
--- a/docs/constants.md
+++ b/docs/constants.md
@@ -96,3 +96,11 @@ It has the following values:
 
 - `ColorType.Solid` - solid color.
 - `ColorType.VerticalGradient` - gradient color.
+
+## PlotRowSearchMode
+
+`PlotRowSearchMode` enums is used to specify plot by index searching mode
+
+- `PlotRowSearchMode.Exact` - search a plot exact at index
+- `PlotRowSearchMode.NearestLeft` - search the nearest plot from left side
+- `PlotRowSearchMode.NearestRight` - search the nearest plot from right side
--- a/docs/events.md
+++ b/docs/events.md
@@ -67,7 +67,7 @@ export type MouseEventHandler = (param: MouseEventParams) => void;
 - `time` (`Time`, optional) - time
 - `point`: (`{ x: number, y: number }`, optional) - coordinate
 - `seriesPrices`: (`Map<ISeriesApi, number | OHLC>`) - series prices
-- `logicalIndex`: (`Logical`, optional) - logical index
+- `index`: (`Logical`, optional) - logical index
 
 `time` is not defined if an event was fired outside of data range (e.g. right/left of all data points).
 
--- a/docs/events.md
+++ b/docs/events.md
@@ -66,11 +66,11 @@ export type MouseEventHandler = (param: MouseEventParams) => void;
 
 - `time` (`Time`, optional) - time
 - `point`: (`{ x: number, y: number }`, optional) - coordinate
-- `seriesPrices`: (`Map<ISeriesApi, number | OHLC>`) - series prices
+- `seriesPrices`: (`Map<ISeriesApi, number | BarPrices>`) - series prices
 - `index`: (`Logical`, optional) - logical index
 
 `time` is not defined if an event was fired outside of data range (e.g. right/left of all data points).
 
 `point` is not defined if an event was fired outside of the chart (for example on a mouse leave event).
 
-`seriesPrices` is an object with prices of all series corresponding to the event point. The object keys are series APIs, values are prices. Each price value is a number for single-value series types (line, area, histogram) or OHLC structure for candlestick and bar series.
+`seriesPrices` is an object with prices of all series corresponding to the event point. The object keys are series APIs, values are prices. Each price value is a number for single-value series types (line, area, histogram) or [BarPrices](./types.md#BarPrices) structure for candlestick and bar series.
--- a/docs/events.md
+++ b/docs/events.md
@@ -67,10 +67,12 @@ export type MouseEventHandler = (param: MouseEventParams) => void;
 - `time` (`Time`, optional) - time
 - `point`: (`{ x: number, y: number }`, optional) - coordinate
 - `seriesPrices`: (`Map<ISeriesApi, number | BarPrices>`) - series prices
-- `index`: (`Logical`, optional) - logical index
+- `index`: (`Logical | null`) - logical index
 
 `time` is not defined if an event was fired outside of data range (e.g. right/left of all data points).
 
 `point` is not defined if an event was fired outside of the chart (for example on a mouse leave event).
 
 `seriesPrices` is an object with prices of all series corresponding to the event point. The object keys are series APIs, values are prices. Each price value is a number for single-value series types (line, area, histogram) or [BarPrices](./types.md#BarPrices) structure for candlestick and bar series.
+
+`index` is `null` if an event was fired outside of data range or outside of the chart (e.g. right/left of all data points or mouse point leave event).
--- a/docs/events.md
+++ b/docs/events.md
@@ -66,13 +66,13 @@ export type MouseEventHandler = (param: MouseEventParams) => void;
 
 - `time` (`Time`, optional) - time
 - `point`: (`{ x: number, y: number }`, optional) - coordinate
-- `seriesPrices`: (`Map<ISeriesApi, number | BarPrices>`) - series prices
+- `seriesPrices`: (`Map<ISeriesApi, number | OHLC>`) - series prices
 - `index`: (`Logical | null`) - logical index
 
 `time` is not defined if an event was fired outside of data range (e.g. right/left of all data points).
 
 `point` is not defined if an event was fired outside of the chart (for example on a mouse leave event).
 
-`seriesPrices` is an object with prices of all series corresponding to the event point. The object keys are series APIs, values are prices. Each price value is a number for single-value series types (line, area, histogram) or [BarPrices](./types.md#BarPrices) structure for candlestick and bar series.
+`seriesPrices` is an object with prices of all series corresponding to the event point. The object keys are series APIs, values are prices. Each price value is a number for single-value series types (line, area, histogram) or [OHLC](./types.md#OHLC) structure for candlestick and bar series.
 
 `index` is `null` if an event was fired outside of data range or outside of the chart (e.g. right/left of all data points or mouse point leave event).
--- a/docs/events.md
+++ b/docs/events.md
@@ -67,6 +67,7 @@ export type MouseEventHandler = (param: MouseEventParams) => void;
 - `time` (`Time`, optional) - time
 - `point`: (`{ x: number, y: number }`, optional) - coordinate
 - `seriesPrices`: (`Map<ISeriesApi, number | OHLC>`) - series prices
+- `logicalIndex`: (`Logical`, optional) - logical index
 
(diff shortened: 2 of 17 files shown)


src/api/series-api.ts
```diff
@@ -173,4 +170,34 @@ export class SeriesApi<TSeriesType extends SeriesType> implements ISeriesApi<TSe
 	public seriesType(): TSeriesType {
 		return this._series.seriesType();
 	}
+
+	public getBarInfoByLogicalIndex(logicalIndex: Logical): BarPriceInfoAtTypeMap[TSeriesType] | null {
+		const series = this._series;
+		const bar = series.dataAt(logicalIndex as unknown as TimePointIndex) as SeriesDataAtTypeMap[TSeriesType];
```
kirchet, line 176: We should pass `PlotRowSearchMode` as an argument and use it directly:

```
public getBarInfoByLogicalIndex(logicalIndex: Logical, searchMode: PlotRowSearchMode = PlotRowSearchMode.Exact): BarPriceInfoAtTypeMap[TSeriesType] | null {
    return series.bars().search(logicalIndex as TimePointIndex, searchMode);
}
```

src/index.ts
```diff
@@ -7,6 +7,7 @@ export { PriceScaleMode } from './model/price-scale';
 export { PriceLineSource, LastPriceAnimationMode, LasPriceAnimationMode } from './model/series-options';
 export { TickMarkType } from './model/time-scale';
 export { ColorType } from './model/layout-options';
+export { Logical } from './model/time-data';
```
kirchet, line 10: Here we re-export only const enums

src/model/bar.ts
```diff
@@ -17,3 +17,20 @@ export interface BarCoordinates {
 	lowY: Coordinate;
 	closeY: Coordinate;
 }
+
+interface BarPriceInfo {
+	value: BarPrice;
+	change: number | undefined;
```
kirchet, line 23: We don't need `value` and `change`. We need only `BarPrice` | `BarPrices`

kirchet (changes requested):

docs/events.md
```diff
@@ -67,6 +67,7 @@ export type MouseEventHandler = (param: MouseEventParams) => void;
 - `time` (`Time`, optional) - time
 - `point`: (`{ x: number, y: number }`, optional) - coordinate
 - `seriesPrices`: (`Map<ISeriesApi, number | OHLC>`) - series prices
+- `logicalIndex`: (`Logical`, optional) - logical index
```
timocov, line 70: why is it optional? it seems that a logical index always exists?

src/api/iseries-api.ts
```diff
@@ -61,6 +57,13 @@ export interface ISeriesApi<TSeriesType extends SeriesType> {
 	 */
 	barsInLogicalRange(range: Range<number>): BarsInfo | null;
 
+	/**
+	 * Get bar price info
+	 *
+	 * @param logicalIndex - the logical index to retrieve bar info
+	 */
+	getBarInfoByLogicalIndex(logicalIndex: Logical): BarPriceInfoAtTypeMap[TSeriesType] | null;
```
timocov, line 65: we need to add this function to the docs

timocov: btw, it seems that we don't use `get` prefixes for getters (see `barsInLogicalRange` or `options`)

src/api/ichart-api.ts
```diff
@@ -25,6 +25,7 @@ export interface MouseEventParams {
 	seriesPrices: Map<ISeriesApi<SeriesType>, BarPrice | BarPrices>;
 	hoveredSeries?: ISeriesApi<SeriesType>;
 	hoveredMarkerId?: SeriesMarker<Time>['id'];
+	logicalIndex?: Logical;
```
timocov, line 28: why not just `index`? I don't think that we have different indexes...

src/api/series-api.ts
```diff
@@ -173,4 +170,34 @@ export class SeriesApi<TSeriesType extends SeriesType> implements ISeriesApi<TSe
...
+		// eslint-disable-next-line @typescript-eslint/tslint/config
+		if (typeof bar === 'number') {
+			result = {
+				value: bar,
+				change: prevBar ? (bar as number) - prevBar.value[PlotRowValueIndex.Close] : undefined,
+			};
+		} else {
+			result = {
+				...bar,
+				change: prevBar ? bar.close - prevBar.value[PlotRowValueIndex.Close] : undefined,
```
timocov, line 197: I don't think that this is a good idea to return any kind of "change". Why did you choose close field over any other fields? Could you calculate a "change" value based on 2 values outside of lightweight-charts code?

src/api/series-api.ts
```diff
@@ -173,4 +170,34 @@ export class SeriesApi<TSeriesType extends SeriesType> implements ISeriesApi<TSe
...
+		const series = this._series;
+		const bar = series.dataAt(logicalIndex as unknown as TimePointIndex) as SeriesDataAtTypeMap[TSeriesType];
+		let result = null;
+
+		if (!bar) {
+			return result;
+		}
+
+		const prevBar = series.bars().search(
+			(logicalIndex - 1) as unknown as TimePointIndex,
```
timocov, line 184: why do you use `as unknown` here? I think it should work without it (just `as TimePointIndex`)

src/gui/chart-widget.ts
```diff
@@ -604,12 +605,17 @@ export class ChartWidget implements IDestroyable {
 			? hoveredSource.object.externalId
 			: undefined;
 
+		const logicalIndex = point !== null
+			? this._model.timeScale().coordinateToIndex(point?.x) as unknown as Logical
```
timocov, line 609: you don't need to convert `number` to `Logical` via `unknown`

timocov (changes requested):

docs/events.md
kirchet: I think for the same reason as `time`

timocov: We might not have a `time` because a crosshair might be outside of the data and we won't extrapolate the time in this case, but for index the situation is different, so I'm just curious in what cases `index` might be omitted.

src/gui/chart-widget.ts
tv-ksi:
```
TS2352: Conversion of type 'TimePointIndex' to type 'Logical' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.   Type 'TimePointIndex' is not comparable to type '{ [Symbol.species]: "Logical"; }'.     Types of property '[Symbol.species]' are incompatible.       Type '"TimePointIndex"' is not comparable to type '"Logical"'.
```

docs/events.md
kirchet: But what logical index should we provide if a crosshair is out of a chart?

src/model/series.ts
```diff
@@ -638,3 +631,20 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
...
+			high: plot.value[PlotRowValueIndex.High] as BarPrice,
+			low: plot.value[PlotRowValueIndex.Low] as BarPrice,
+			close: plot.value[PlotRowValueIndex.Close] as BarPrice,
+		};
+	} else {
+		result = plot.value[PlotRowValueIndex.Close] as BarPrice;
+	}
+
+	return result as SeriesDataAtTypeMap[T];
+}
```
kirchet, line 650: 1. Let's move this code before `Series` class
2. There is no need to have a result variable, we can just return a resul in both cases

docs/series-basics.md
```diff
@@ -533,6 +533,29 @@ function onVisibleLogicalRangeChanged(newVisibleLogicalRange) {
 chart.timeScale().subscribeVisibleLogicalRangeChange(onVisibleLogicalRangeChanged);
 ```
 
+### barByIndex
+
+Returns a bar price for the series by the provided logical index. If no one bar satisfy the index, `null` is returned.
```
kirchet, line 538: Maybe bar prices? Also if the structure of bar prices is already described we need to add a link to this description here

docs/series-basics.md
```diff
@@ -533,6 +533,29 @@ function onVisibleLogicalRangeChanged(newVisibleLogicalRange) {
...
+
+- `open`(`BarPrice`) - an item open
+- `high`(`BarPrice`) - an item high
+- `low`(`BarPrice`) - an item low
+- `close`(`BarPrice`) - an item close
+
+The method takes the following arguments:
+
+- `index`(`Logical`) - a logical index to retrieve bar price
+- `searchMode`(`PlotRowSearchMode`) - a search mode
```
kirchet, line 550: Seems like it's the same structure as in `subscribeCrosshairMove` callback argument `seriesPrices`. It's already described here: https://github.com/tradingview/lightweight-charts/blob/master/docs/events.md#types. You can just add a link to this description here and modify the existing description if you want

docs/series-basics.md
```diff
@@ -533,6 +533,29 @@ function onVisibleLogicalRangeChanged(newVisibleLogicalRange) {
...
+### barByIndex
+
+Returns a bar price for the series by the provided logical index. If no one bar satisfy the index, `null` is returned.
+
+The returned value can be a `number` or an `object` with the following properties:
+
+- `open`(`BarPrice`) - an item open
+- `high`(`BarPrice`) - an item high
+- `low`(`BarPrice`) - an item low
+- `close`(`BarPrice`) - an item close
```
kirchet, line 545: Seems like it's the same structure as in `subscribeCrosshairMove` callback argument `seriesPrices`. It's already described here: https://github.com/tradingview/lightweight-charts/blob/master/docs/events.md#types. You can just add a link to this description here and modify the existing description if you want

tv-ksi: It's not, here it's an object or a number, the seriesPrices it is a number or a OHLC (tuple)

docs/series-basics.md
```diff
@@ -547,13 +547,16 @@ The returned value can be a `number` or an `object` with the following propertie
...
 
-```javascript
-const lineSeriesApi = chartApi.addLineSeries();
-const barPrice = lineSeriesApi.barByIndex(logicalIndex, PlotRowSearch.NearestLeft);
+```typescript
+const lineSeriesApi: ISeriesApi<'Line'> = chartApi.addLineSeries();
+const barPrice: number | null = lineSeriesApi.barByIndex(logicalIndex, PlotRowSearch.NearestLeft);
+
+const barSeriesApi: ISeriesApi<'Bar'> = chartApi.addBarSeries();
+const barPrices: object | null = barSeriesApi.barByIndex(logicalIndex, PlotRowSearch.NearestRight);
```
kirchet, line 559: Why do you use TS? All other examples are on JS

docs/series-basics.md
kirchet: > The object keys are series APIs, values are prices

Seems like these prices are the same as here. I suggest to move there the description of this type and add a link to it.

docs/constants.md
```diff
@@ -96,3 +96,11 @@ It has the following values:
...
 - `ColorType.Solid` - solid color.
 - `ColorType.VerticalGradient` - gradient color.
+
+## PlotRowSearchMode
+
+`PlotRowSearchMode` enums is used to specify plot by index searching mode
+
+- `PlotRowSearchMode.Exact` - search a plot exact at index
+- `PlotRowSearchMode.NearestLeft` - search the nearest plot from left side
+- `PlotRowSearchMode.NearestRight` - search the nearest plot from right side
```
kirchet, line 106: Actually we search at exact index or nearest non-empty left/nearest non-empty righ bar prices (if at exact indes nothing found) correspondingly for `PlotRowSearchMode.NearestLeft` / `PlotRowSearchMode.NearestRight`

docs/series-basics.md
tv-ksi: OHLC it's a tuple [open: BarPrice, high: BarPrice, low: BarPrice, close: BarPrice], it's not the same as object

docs/series-basics.md
```diff
@@ -551,12 +551,9 @@ The method takes the following arguments:
...
 
-```typescript
-const lineSeriesApi: ISeriesApi<'Line'> = chartApi.addLineSeries();
-const barPrice: number | null = lineSeriesApi.barByIndex(logicalIndex, PlotRowSearch.NearestLeft);
-
-const barSeriesApi: ISeriesApi<'Bar'> = chartApi.addBarSeries();
-const barPrices: object | null = barSeriesApi.barByIndex(logicalIndex, PlotRowSearch.NearestRight);
+```javascript
+const lineSeriesApi = chartApi.addLineSeries();
+const barPrice = lineSeriesApi.barByIndex(logicalIndex, PlotRowSearch.NearestLeft);
```
kirchet, line 556:
```suggestion
const barPrice = lineSeriesApi.barByIndex(logicalIndex, LightweightCharts.PlotRowSearch.NearestLeft);
```

PR #854, opened 2021-09-27
Added methods to get time axis size and subscribe on size change

### LINKED ISSUE #853: Add additional methods to API for positioning custom controls over chart

For now we cannot position custom controls over the chart because there is no corresponding API to get time axis size and subscribe to its size change.

Add additional methods to the time scale API: `width`, `height`, `subscribeSizeChange`

kirchet:
**Type of PR:**  enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #853
- [ ] Includes tests
- [x] Documentation update

**Overview of change:**

```
chart.timeScale().height();
chart.timeScale().width();
const handler = (width, height) => console.log(width, height));
chart.timeScale().subscribeSizeChange(handler);
chart.timeScale().unsubscribeSizeChange(handler);
```

### DIFF
--- a/docs/time-scale.md
+++ b/docs/time-scale.md
@@ -169,6 +169,22 @@ Returns [Time](./time.md) of a bar that is located on the passed coordinate or `
 chart.timeScale().coordinateToTime(42);
 ```
 
+### width
+
+Returns a current width of the time scale.
+
+```js
+console.log(chart.timeScale().width());
+```
+
+### height
+
+Returns a current height of the time scale.
+
+```js
+console.log(chart.timeScale().height());
+```
+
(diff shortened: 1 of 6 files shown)


timocov (changes requested):
The changes look good for me, Would you mind to add some tests please? It might be:
- if the time axis is shown/hidden, then the size should be changed (and when it is hidden the size should be 0x0)
- if the chart's size is changed the the time axis' size should be changed as well
- if price scale is shown/hidden the size should be changed as well
- (unnecessary to add, just as an idea) if the new data item is added which extends visible price range most likely the size of time axis will be changed as well

PR #876, opened 2021-10-28
Documentation about time zones support

### LINKED ISSUE #781: How to work/add support with different time zones

**Steps/code to reproduce:**
Visit https://jsfiddle.net/b5v8L9et/

**Actual behavior:**
Day TickMark is displayed at 2:00 instead of 0:00

**Expected behavior:**
Allow to configure chart to display tickmark in position valid for specific locale - at 0:00 for day, month, year

**Screenshots:**
obraz

https://jsfiddle.net/b5v8L9et/

I provided custom tickMarkFormatter (and timeFormatter for easier readability) but there is no way to customize where specific tickMark should be displayed

timocov:
**Type of PR:** enhancement, docs

**PR checklist:**

- [x] Addresses an existing issue: fixes #781
- [ ] Includes tests
- [x] Documentation update

Related to #643, https://stackoverflow.com/questions/65785045/how-to-set-a-custom-time-zone-for-lightweight-charts

### DIFF
--- a/docs/time-zones.md
+++ b/docs/time-zones.md
@@ -16,7 +16,7 @@ Out of the box we could rely on 2 APIs:
 - [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl)
 
 And even if to format a date we could (and we do) use `Date` object with its `toLocaleString` method (and we could even pass a `timeZone` field as an option),
-but what's about getting a date/time fields?
+but how about date/time field?
 
 All to solve this it seems that the only solution we have is `Date`'s getters, e.g. `getHours`. Here we could use 2 APIs:
 
--- a/docs/time-zones.md
+++ b/docs/time-zones.md
@@ -94,7 +94,7 @@ If you have lots of data items and the performance of other solutions doesn't fi
 
 The better performance could be achieved with this approach because:
 
-- you don't need to parse dates every time you want to get an offset so you could use lowerbound algorithm (which is `O(log N)`) to find an offset of very first data point quickly
+- you don't need to parse dates every time you want to get an offset so you could use [lowerbound algorithm](https://en.wikipedia.org/wiki/Upper_and_lower_bounds) (which is `O(log N)`) to find an offset of very first data point quickly
 - after you found an offset, you go through all data items and check whether an offset should be changed or not to the next one (based on a time of the next time shift)
 
 ## Why we didn't implement it in the library
--- a/docs/time-zones.md
+++ b/docs/time-zones.md
@@ -4,7 +4,7 @@ This doc describes what do you need to do if you want to add a time zone support
 
 ## Background
 
-By default, `lightweight-charts` doesn't support time zones in any kind, just because JavaScript doesn't an API to do that.
+By default, `lightweight-charts` doesn't support time zones in any kind, just because JavaScript doesn't have an API to do that.
 Things that the library uses internally includes an API to:
 
 - Format a date
--- a/docs/time-zones.md
+++ b/docs/time-zones.md
@@ -1,0 +1,111 @@
+# Working with time zones
+
+This doc describes what do you need to do if you want to add a time zone support to your chart.
+
+## Background
+
+By default, `lightweight-charts` doesn't support time zones in any kind, just because JavaScript doesn't an API to do that.
+Things that the library uses internally includes an API to:
+
+- Format a date
+- Get a date and/or time parts of a date object (year, month, day, hours, etc)
+
+Out of the box we could rely on 2 APIs:
+
+- [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)
+- [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl)
+
+And even if to format a date we could (and we do) use `Date` object with its `toLocaleString` method (and we could even pass a `timeZone` field as an option),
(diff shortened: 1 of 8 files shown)


docs/time-zones.md
```diff
@@ -0,0 +1,111 @@
+# Working with time zones
+
+This doc describes what do you need to do if you want to add a time zone support to your chart.
+
+## Background
+
+By default, `lightweight-charts` doesn't support time zones in any kind, just because JavaScript doesn't an API to do that.
```
romfrancois, line 7: > doesn't an API to do that
Missing one word: have

docs/time-zones.md
```diff
@@ -0,0 +1,111 @@
...
+- Format a date
+- Get a date and/or time parts of a date object (year, month, day, hours, etc)
+
+Out of the box we could rely on 2 APIs:
+
+- [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)
+- [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl)
+
+And even if to format a date we could (and we do) use `Date` object with its `toLocaleString` method (and we could even pass a `timeZone` field as an option),
+but what's about getting a date/time fields?
```
romfrancois, line 19: > what's about getting a date/time fields?

what about getting a date/time field?

docs/time-zones.md
timocov:
```suggestion
By default, `lightweight-charts` doesn't support time zones in any kind, just because JavaScript doesn't have an API to do that.
```

docs/time-zones.md
timocov: I'd say this is rhetorical question. I was trying to say that JS doesn't have powerful API to work with dates, only to format them, but not work with date & time parts individually. Do you have any idea how we could rephrase it?

romfrancois: Maybe with `..but how about date/time field?`?

timocov:
```suggestion
but how about date/time field?
```

docs/time-zones.md
```diff
@@ -0,0 +1,111 @@
...
+Lets consider the example.
+
+Lets say you have a bar with time `2021-01-01T10:00:00.000Z` (a string representation is just for better readability).
+And you want to display your chart in `Europe/Moscow` time zone.
+
+According to tz database, for `Europe/Moscow` time zone a time offset at this time is `UTC+03:00`, i.e. +3 hours (pay attention that you cannot use the same offset all the time, because of DST and many other things!).
+
+By this means, the time for `Europe/Moscow` is `2021-01-01 13:00:00.000` (so basically you want to display this time over the UTC one).
+
+So to display your chart in `Europe/Moscow` time zone you need to correct the time of your bar by time zone offset, i.e. by +3 hours, so it will be `2021-01-01T13:00:00.000Z` (keep in mind that the library will use it as UTC time, but it is not correct time actually).
```
edew, line 49: _"(keep in mind that the library will use it as UTC time, but it is not correct time actually)."_ 

I think this makes it unclear what I should be doing (i.e. what is the correct way to adjust for my time zone)?

_"To display your chart in the `Europe/Moscow` time zone you would need to adjust the time of your data by +3 hours. So `2021-01-01T10:00:00.000Z` would become `2021-01-01T13:00:00.000Z` [...]"_

docs/time-zones.md
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+### `tzdata` solution
+
+If you have lots of data items and the performance of other solutions doesn't fit your requirements you could try to implement more complex solution by using raw [`tzdata`](https://www.npmjs.com/package/tzdata).
+
+The better performance could be achieved with this approach because:
+
+- you don't need to parse dates every time you want to get an offset so you could use lowerbound algorithm (which is `O(log N)`) to find an offset of very first data point quickly
```
edew, line 97: _" [...] lowerbound algorithm [...]"_

Like in `src/helpers/algorithms.ts` for example?

docs/time-zones.md
```diff
@@ -0,0 +1,111 @@
+# Working with time zones
+
+This doc describes what do you need to do if you want to add a time zone support to your chart.
+
+## Background
+
+By default, `lightweight-charts` doesn't support time zones in any kind, just because JavaScript doesn't have an API to do that.
```
edew, line 7: _"in any kind" -> "of any kind"_ I think.

docs/time-zones.md
```diff
@@ -0,0 +1,111 @@
+# Working with time zones
+
+This doc describes what do you need to do if you want to add a time zone support to your chart.
```
edew, line 3: _"a time zone support" -> "time zone support"_ I think.

docs/time-zones.md
timocov: I thought it is a common name https://en.wikipedia.org/wiki/Upper_and_lower_bounds, https://www.cplusplus.com/reference/algorithm/lower_bound/. Probably we could add a link to wiki/code

timocov:
```suggestion
- you don't need to parse dates every time you want to get an offset so you could use [lowerbound algorithm](https://en.wikipedia.org/wiki/Upper_and_lower_bounds) (which is `O(log N)`) to find an offset of very first data point quickly
```

edew: I can't accept suggestions but yes I think a link is good. 

I think the reason I added this comment is I predict that we may get users reading this and then asking _"how do I implement a lower bound algorithm"_?

PR #879, opened 2021-11-02
Use lowerbound in TimeScale timeToIndex search

### LINKED ISSUE #767: Chart with a lot of markers lags when chart.update of a single candle is called
**Lightweight Charts Version:** 3.3 

**Steps/code to reproduce:**

Create a chart with a LOT of markers / chart data, the chart I am viewing has a total of 5747 markers, example image can be seen below.

Everything runs incredibly smooth when I don't call following method (once a second or let's say every 5 seconds), but when I do call update on the final candle to repaint, cursor, scrolling and dragging all hangs and it becomes very choppy.

**Actual behavior:**

It repaints every single marker, which takes around 300ms and the chart becomes unusable.

**Expected behavior:**

I would assume that it shouldn't touch any of the existing markers, unless there is a marker on the candle that was updated, then maybe it should update the marker position?

**Screenshots:**

chart with a lot of markers:
image

Devtools output when calling .update() with a single new candle:
image

edew:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #767 
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Use the existing lowerbound algorithm to reduce the amount of work done in `_recalculateMarkers`.

Profiling on my machine shows a reduction in execution time of `_recalculateMarkers` for the series update case shown here https://github.com/tradingview/lightweight-charts/issues/767#issuecomment-864123626 from ~580ms to ~3ms. In practice this means that the chart remains responsive while recalculating the 12660 markers in the example.


src/model/series.ts
```diff
@@ -255,10 +255,12 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
 		this._paneView.update('options');
 	}
 
-	public setData(data: readonly SeriesPlotRow<T>[]): void {
+	public setData(data: readonly SeriesPlotRow<T>[], firstChangedPointIndex?: TimePointIndex): void {
 		this._data.setData(data);
 
-		this._recalculateMarkers();
+		if (firstChangedPointIndex !== undefined) {
+			this._recalculateMarkers(firstChangedPointIndex);
```
timocov, line 262: I'm afraid this cannot work in common case since `firstChangedPointIndex` means the first change point on the time scale, but if data is updated it doesn't necessary mean that time scale was changed. You can try the following example (you can add it to the tests as well to make sure that this situation is handled correctly):

```js
function runTestCase(container) {
	const chart = LightweightCharts.createChart(container);

	const candlestickSeries = chart.addCandlestickSeries();

	candlestickSeries.setData([
		{ time: '2021-10-01', open: 100, high: 110, low: 100, close: 110 },
		{ time: '2021-10-02', open: 110, high: 110, low: 100, close: 100 },
		{ time: '2021-10-03', open: 100, high: 110, low: 100, close: 110 },
		{ time: '2021-10-04', open: 120, high: 130, low: 120, close: 125 },
	]);


	const lineSeries = chart.addLineSeries();

	lineSeries.setData([
		{ time: '2021-10-01', value: 153 },
		{ time: '2021-10-03', value: 152 },
	]);

	lineSeries.setMarkers([
		{ time: '2021-10-05', position: 'aboveBar', shape: 'circle' },
	]);

	lineSeries.setData([
		{ time: '2021-10-01', value: 153 },
		{ time: '2021-10-02', value: 152 },
	]);

	// uncomment this call to see how it should look like (note that we're setting _the same data_)
	// lineSeries.setMarkers([
	// 	{ time: '2021-10-05', position: 'aboveBar', shape: 'circle' },
	// ]);

}
```

timocov (changes requested):
See comment

timocov (approved):

PR #892, opened 2021-11-12
Fixed last price animation on setting new data

### LINKED ISSUE #886: Last price animation is active when no data added to the right (but to the left)
**Lightweight Charts Version:**  3.7.0

**Steps/code to reproduce:**

* Enable last price animation on data update
* Add more history bars via `setData`

**Actual behavior:**

Animation is active

**Expected behavior:**

No animation. Animation should start only if a series has data and we're adding new bars to the right side or updating the last bar

**Screenshots:**

animation

https://jsfiddle.net/6hq2vopz/2/

kirchet:
**Type of PR:** 

**PR checklist:**

- [x] Addresses an existing issue: fixes #886
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Fixed incorrect last price animation on first `setData` and adding new data to the left side

### DIFF
--- a/src/api/data-layer.ts
+++ b/src/api/data-layer.ts
@@ -4,7 +4,7 @@ import { lowerbound } from '../helpers/algorithms';
 import { ensureNotNull } from '../helpers/assertions';
 import { isString } from '../helpers/strict-type-checks';
 
-import { Series } from '../model/series';
+import { Series, SeriesUpdateInfo } from '../model/series';
 import { SeriesPlotRow } from '../model/series-data';
 import { SeriesType } from '../model/series-options';
 import { BusinessDay, TimePoint, TimePointIndex, TimeScalePoint, UTCTimestamp } from '../model/time-data';
@@ -124,6 +124,10 @@ export interface SeriesChanges {
 	 * Data to be merged into series' plot list
 	 */
 	data: readonly SeriesPlotRow[];
+	/**
+	 * Additional info about this change
+	 */
+	info?: SeriesUpdateInfo;
 }
 
 export interface DataUpdateResponse {
@@ -155,6 +159,26 @@ function createEmptyTimePointData(timePoint: TimePoint): TimePointData {
 	return { index: 0 as TimePointIndex, mapping: new Map(), timePoint };
 }
 
+function seriesRowsLastTime(seriesRows: SeriesPlotRow[] | undefined): UTCTimestamp | undefined {
+	if (seriesRows === undefined) {
+		return undefined;
+	}
+
+	return seriesRows.length > 0 ? seriesRows[seriesRows.length - 1].time.timestamp : undefined;
+}
+
+function seriesUpdateInfo(seriesRows: SeriesPlotRow[] | undefined, prevSeriesRows: SeriesPlotRow[] | undefined): SeriesUpdateInfo | undefined {
+	const lastTime = seriesRowsLastTime(seriesRows);
+	if (lastTime !== undefined) {
+		const prevLastTime = seriesRowsLastTime(prevSeriesRows);
+		if (prevLastTime !== undefined && lastTime >= prevLastTime) {
+			return { lastBarUpdatedOrNewBarsAddedToTheRight: true };
+		}
+	}
+
+	return undefined;
+}
+
 export class DataLayer {
 	// note that _pointDataByTimePoint and _seriesRowsBySeries shares THE SAME objects in their values between each other
 	// it's just different kind of maps to make usages/perf better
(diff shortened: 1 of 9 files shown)


src/model/series.ts
```diff
@@ -240,13 +240,24 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
 	}
 
 	public setData(data: readonly SeriesPlotRow<T>[]): void {
+		const lastIndex = this._data.lastIndex();
+		const newRealtimeDataReceived = lastIndex !== null && data.length > 0 && data[data.length - 1].index >= lastIndex;
```
timocov, line 244: it is not really true - you could extend time scale in the middle, indexes will be changed, but no new data received.

kirchet: Yes, and in this case we should display last price animation for this series. BTW, maybe name is not so clear and correct

timocov: > and in this case we should display last price animation for this series

Why? Nothing is changed...

kirchet: Hm, so we need to compare times? Not indexes?

timocov: Yeah, I think so. Btw we already did this [in updating time scale](https://github.com/tradingview/lightweight-charts/blob/ebd3586cd7334b9a251cb7ea2e16d5308166d3c2/src/model/chart-model.ts#L560-L589).

But most likely we need to provide an index of the first changed point for every series (or for being updated one) and rely on this index. https://github.com/tradingview/lightweight-charts/issues/886#issuecomment-966274225

Also, please keep in mind that it is possible that an array _could be_ shared between data layer and series due perf optimizations (that's why a data array is `readonly`).

kirchet: We can just change the current condition to compare times instead of indexes, I think that tha's more easiest way to do the same work.

kirchet: https://github.com/tradingview/lightweight-charts/pull/892/commits/12c227b2cfb18b84801853f3f50b20aefe180f0c

tests/e2e/graphics/test-cases/series/no-last-price-animation-on-first-set-data-after-reseting-data.js
```diff
@@ -0,0 +1,23 @@
...
+	series.setData([]);
+
+	series.setData([
+		{ time: '1990-04-24', value: 0 },
+		{ time: '1990-04-25', value: 1 },
+		{ time: '1990-04-26', value: 2 },
+		{ time: '1990-04-28', value: 3 },
+	]);
+
+	return new Promise(resolve => setTimeout(resolve, 500));
```
timocov, line 22: what value should be here? 500 means we want to handle after animation time or somewhere in the middle?

src/api/data-layer.ts
```diff
@@ -277,9 +307,11 @@ export class DataLayer {
 
 		this._updateLastSeriesRow(series, plotRow);
 
+		const info: SeriesUpdateInfo = { lastBarUpdatedOrNewBarsAddedToTheRight: true };
```
timocov, line 310: it is not really true. what if you replace the last item with whitespace? time scale won't be affected, no new data added, but animation will be started?

src/api/data-layer.ts
```diff
@@ -155,6 +159,26 @@ function createEmptyTimePointData(timePoint: TimePoint): TimePointData {
...
+	}
+
+	return seriesRows.length > 0 ? seriesRows[seriesRows.length - 1].time.timestamp : undefined;
+}
+
+function seriesUpdateInfo(seriesRows: SeriesPlotRow[] | undefined, prevSeriesRows: SeriesPlotRow[] | undefined): SeriesUpdateInfo | undefined {
+	const lastTime = seriesRowsLastTime(seriesRows);
+	if (lastTime !== undefined) {
+		const prevLastTime = seriesRowsLastTime(prevSeriesRows);
+		if (prevLastTime !== undefined && lastTime >= prevLastTime) {
```
timocov, line 174: `lastTime >= prevLastTime` does this mean that if you scroll to the left (to the past) and you'll set data for the whole period this animation will be run as well even nothing is changed for the last bar?

tests/e2e/graphics/test-cases/series/no-last-price-animation-on-first-set-data-after-reseting-data.js
kirchet: In the middle

src/api/data-layer.ts
kirchet: https://github.com/tradingview/lightweight-charts/pull/892/commits/7fabd9d7c16aee4573caeec1b4294cb5293171f5

src/api/data-layer.ts
kirchet: https://github.com/tradingview/lightweight-charts/pull/892/commits/7fabd9d7c16aee4573caeec1b4294cb5293171f5

timocov (approved):

PR #900, opened 2021-11-17
Fixing creating several baseline series with different base value

### LINKED ISSUE #898: When rendering multiple charts with baseline series, baseValue of the last element is used on all charts series.
**Lightweight Charts 3.7.0**
**React 16.14.0** used in my project, **React 17.0.2** in codesandbox with same results
**Codesandbox example :** https://codesandbox.io/s/pensive-blackwell-cuz1j?file=/src/Baseline.jsx

Hello,

While trying to render multiple charts, each with a baseline series, setting the baseValue on each chart independently doesn't seem to work as expected, instead it applies the value of the last series to each element. 

In my example, the grey price line is to indicate where the baseValue price should be set for each chart series.

image

I have tried everything I could think of as far as React goes. To name a few, setting baseValue in either addBaselineSeries or applyOptions doesn't affect anything. Thinking it might be because each series is not unique, I tried keeping a list of refs of all my series in state, outside of my loop, and once the charts rendered, iterating through the list calling applyOptions on each series to set baseValue, without success.

If I play around with the data and put the first item at the end of my array, you notice that now it works for that particular chart, but only because it's now last. And the second to last (previously last) shows all red because its values are lower than the new baseValue price.

image

I'm scratching my head over this because all other series options seem to work fine. I have read the docs without finding anything about this. I'm still thinking it could be user error, my apologies if so. That is the reason I did not file this under bugs, just in case.

Thank you for this great library!

timocov:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #898
- [X] Includes tests
- [ ] Documentation update

### DIFF
--- a/src/api/chart-api.ts
+++ b/src/api/chart-api.ts
@@ -227,7 +227,8 @@ export class ChartApi implements IChartApi, DataUpdatesConsumer<SeriesType> {
 		options = migrateOptions(options);
 		patchPriceFormat(options.priceFormat);
 
-		const strictOptions = merge(clone(seriesOptionsDefaults), baselineStyleDefaults, options) as BaselineSeriesOptions;
+		// baseline style defaults has an object in it so we have to clone it as well to avoid assigning fields to defaults
+		const strictOptions = merge(clone(seriesOptionsDefaults), clone(baselineStyleDefaults), options) as BaselineSeriesOptions;
 		const series = this._chartWidget.model().createSeries('Baseline', strictOptions);
 
 		const res = new SeriesApi<'Baseline'>(series, this, this);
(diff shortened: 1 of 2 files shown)


src/api/chart-api.ts
```diff
@@ -227,7 +227,8 @@ export class ChartApi implements IChartApi, DataUpdatesConsumer<SeriesType> {
 		options = migrateOptions(options);
 		patchPriceFormat(options.priceFormat);
 
-		const strictOptions = merge(clone(seriesOptionsDefaults), baselineStyleDefaults, options) as BaselineSeriesOptions;
+		// baseline style defaults has an object in it so we have to clone it as well to avoid assigning fields to defaults
```
kirchet, line 230:
```suggestion
		// to avoid assigning fields to defaults we have to clone them
```

kirchet (approved):

PR #929, opened 2021-12-08
fix: overlay align to the right when no left/right price scale defined

### LINKED ISSUE #926: Overlay series title is not rendered when no series use right price scale
**Lightweight Charts Version:** 3.7.0

**Steps/code to reproduce:**

```ts
import { createChart, LineData, UTCTimestamp } from "lightweight-charts";

const chartRoot = document.getElementById("main-chart") as HTMLDivElement;
const chart = createChart(chartRoot, {
  width: 800,
  height: 600,
});

const line1 = chart.addLineSeries({
  priceScaleId: "overlay",
  color: "rgb(255, 0, 0)",
  title: "SIN",
});
const line2 = chart.addLineSeries({
  priceScaleId: "overlay",
  color: "rgb(0, 255, 0)",
  title: "COS",
});

const data1: LineData[] = [];
const data2: LineData[] = [];

const now = new Date();
function dateToUTC(date: Date): UTCTimestamp {
  return Math.floor(date.getTime() / 1000) as UTCTimestamp;
}

for (let i = 0; i < 90; i++) {
  data1.push({
    time: dateToUTC(now),
    value: Math.sin(i * 4 * Math.PI / 180),
  });

  data2.push({
    time: dateToUTC(now),
    value: Math.cos(i * 4 * Math.PI / 180),
  });

  now.setDate(now.getDate() + 1);
}

line1.setData(data1);
line2.setData(data2);
```

**Actual behavior:**

Series title is not rendered

**Expected behavior:**

Series title should rendered

When another series is added that use right priceScale, even though it's not used, will fixed issue above. However if it used only left priceScale, it won't render title properly again
```ts
const line3 = chart.addLineSeries();
```

**Screenshots:**

image

with line3 added
image

Ffloriel:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #926
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

The price scale should have an indication about where it should be align (left or right).
The default price scale is an overlay if there is no left/right price scale.
Changes essentially make overlay price scale align to the right by default.

Related issue: the option is not reflected in `this._model.options().leftPriceScale.visible` when changing with
`chart.pricescale('left').applyOptions({ visible: true })`.

### DIFF
--- a/src/model/chart-model.ts
+++ b/src/model/chart-model.ts
@@ -385,6 +385,18 @@ export class ChartModel implements IDestroyable {
 	}
 
 	public applyPriceScaleOptions(priceScaleId: string, options: DeepPartial<PriceScaleOptions>): void {
+		if (priceScaleId === DefaultPriceScaleId.Left) {
+			this.applyOptions({
+				leftPriceScale: options,
+			});
+			return;
+		} else if (priceScaleId === DefaultPriceScaleId.Right) {
+			this.applyOptions({
+				rightPriceScale: options,
+			});
+			return;
+		}
+
 		const res = this.findPriceScale(priceScaleId);
 
 		if (res === null) {
@@ -394,18 +406,9 @@ export class ChartModel implements IDestroyable {
 
 			return;
 		}
-		if (priceScaleId === 'left') {
-			this.applyOptions({
-				leftPriceScale: options,
-			});
-		} else if (priceScaleId === 'right') {
-			this.applyOptions({
-				rightPriceScale: options,
-			});
-		} else {
-			res.priceScale.applyOptions(options);
-			this._priceScalesOptionsChanged.fire();
-		}
+
+		res.priceScale.applyOptions(options);
+		this._priceScalesOptionsChanged.fire();
 	}
 
 	public findPriceScale(priceScaleId: string): PriceScaleOnPane | null {
--- a/src/model/chart-model.ts
+++ b/src/model/chart-model.ts
@@ -394,9 +394,18 @@ export class ChartModel implements IDestroyable {
 
 			return;
 		}
-
-		res.priceScale.applyOptions(options);
-		this._priceScalesOptionsChanged.fire();
+		if (priceScaleId === 'left') {
+			this.applyOptions({
+				leftPriceScale: options,
+			});
+		} else if (priceScaleId === 'right') {
+			this.applyOptions({
+				rightPriceScale: options,
+			});
+		} else {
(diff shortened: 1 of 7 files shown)


src/model/chart-model.ts
```diff
@@ -394,9 +394,18 @@ export class ChartModel implements IDestroyable {
 
 			return;
 		}
-
-		res.priceScale.applyOptions(options);
-		this._priceScalesOptionsChanged.fire();
+		if (priceScaleId === 'left') {
```
timocov, line 397: I think we can use `DefaultPriceScaleId` enum instead of hardcoded values here

src/model/chart-model.ts
```diff
@@ -394,9 +394,18 @@ export class ChartModel implements IDestroyable {
...
+			this.applyOptions({
+				leftPriceScale: options,
+			});
+		} else if (priceScaleId === 'right') {
+			this.applyOptions({
+				rightPriceScale: options,
+			});
+		} else {
+			res.priceScale.applyOptions(options);
```
timocov, line 406: it seems that the code above with finding a price scale by id is needed only here so we can move it in `else` statement. if even write it in the following way:

```
if (price scale is right) {
  apply for right();
  return;
}

if (price scale is left) {
  apply for left();
  return;
}

all previous code will be here;
```

src/views/pane/pane-price-axis-view.ts
```diff
@@ -69,17 +69,16 @@ export class PanePriceAxisView implements IPaneView {
 		}
 
 		const position = pane.priceScalePosition(priceScale);
-		if (position === 'overlay') {
-			return null;
-		}
+		// with no left and right price scale, overlay will be align to the right
+		const align = position === 'overlay' ? 'right' : position;
```
timocov, line 73: will it work if the visible scale is left, not right?

timocov (changes requested):

src/views/pane/pane-price-axis-view.ts
Ffloriel: will not work. We can display by default in the price scale that is visible. If the left and right are not visible, I'm guessing it means we shouldn't display the overlay?

timocov: > I'm guessing it means we shouldn't display the overlay?

Yeah, I'd expect this.

timocov (approved):

Issue #200, opened 2019-09-07
React Component - integration


jurepetrovic:
How can I integrate this with React? 
Do I have to create custom react component?

Regards, 
Jure

AoDev:
You need to render an element with a [ref](https://reactjs.org/docs/refs-and-the-dom.html) that you can use to get a reference to the dom element and mount the chart.

jurepetrovic:
I did a custom react component for now. Just for test, but it seems ok for now. 
Obviously, I will need to add some props and state. 

`import * as React from 'react';
import { createChart } from 'lightweight-charts';


export class LightweightChart extends React.PureComponent {
    
    static defaultProps = {
		containerId: 'lightweight_chart_container',
	};

    chart = null;

    componentDidMount() {

        const chart = createChart(this.props.containerId, { width: 800, height: 600 });
        this.chart = chart;

        const lineSeries = chart.addLineSeries();
        
        lineSeries.setData([
            { time: '2019-04-10', value: 60.01 },
            { time: '2019-04-11', value: 80.01 },
        ]);

        const barSeries = chart.addBarSeries({
            thinBars: false,
        });

        // set the data
        barSeries.setData([
            { time: "2019-04-10", open: 141.77, high: 170.39, low: 120.25, close: 145.72 },
            { time: "2019-04-11", open: 145.72, high: 147.99, low: 100.11, close: 108.19 },
                 ]);
    }

	componentWillUnmount() {
		if (this.chart !== null) {
			this.chart.remove();
			this.chart = null;
		}
	}

	render() {
		return (
			<div
				id={ this.props.containerId }
				className={ 'LightweightChart' }
			/>
		);
	}
}
`
Do you see any problems with that?

AoDev:
Personally I don't code React apps that way. I always keep the state outside of the component. That said, your approach is fine, except that personally I would use a ref instead of a static id.

jurepetrovic:
Yes, I noticed that while reading your "refs" link. Since I don't need more than one for now, I'll keep it simple ;)

How do you mean state outside component? You make extra component for the state or you keep it in parent component? 

Thanks for your help!

ackimwilliams:
@jurepetrovic were you able to resolve this issue? Still having this issue: lightweight-charts.esm.production.js:7 Uncaught (in promise) Error: Value is null

jurepetrovic:
Yes, it worked with code above. Then I just included <LightweightChart /> component where needed.

[...]

PR #958, opened 2021-12-31
Docusaurus react tutorial

romfrancois:
**Type of PR:** enhancement

**PR checklist:**
- [x] Documentation update

**Overview of change:**
Adds a React tutorial to integrate LWC to a project from scratch.


website/docs/tutorials/react/tutorials.mdx
```diff
@@ -0,0 +1,14 @@
+---
+id: tutorials
+title: Examples
+---
+
+If you've successfully followed all the steps on the [setup page](./howTo) section, you can now check some examples on how to use LWC
+
+### series-type
```
edew, line 8: Can we move these examples for each series to the corresponding section of series-types.md and replace the images there with interactive examples?

website/src/pages/_react/react-components.tsx
```diff
@@ -0,0 +1,194 @@
...
+			const series = chart.${seriesType}();
+			series.setData(data);
+		}
+
+		// When used as a component, you could 'import' your component into another one
+		import { ChartComponent } from './ChartComponent';
+
+		...
+		export const MyNewComponent = () => {
+			// Careful as for some type data would have to be OHLC formatted as in
```
edew, line 42: I think we should show the correct usage in our examples, so can we change this so `myData` is correct for OHLC data series?

edew (changes requested):

.circleci/config.yml
```diff
@@ -79,6 +79,7 @@ jobs:
     executor: node12-executor
     steps:
       - checkout-with-deps
+      - run: cd website && npm install && cd -
```
edew, line 82: Adding this step here won't do anything unfortunately. It's installing the `website` dependencies for the `install-deps` job and then throwing it away once the job ends.

This `install-deps` job would need to save the `website/node_modules` to the cache (see https://circleci.com/docs/2.0/caching/) so that it can use reused in later jobs (for example the lint job).

I'd move this step to the `lint-ts` job for now to quickly see if it fixes your issue. If it does then we can do it properly and take advantage for the CircleCI cache.

edew (changes requested):

.circleci/config.yml
romfrancois: And I thought it would be clever :( 
Thanks, will try that.

website/sidebars.js
```diff
@@ -16,6 +16,22 @@ const sidebars = {
 				},
 			],
 		},
+		{
+			Tutorials: [
+				{
+					type: 'autogenerated',
+					dirName: 'tutorials',
+				},
```
edew, line 24: I don't think we need this?

Also if we only have one react tutorial file we can remove the react directory and just have a `tutorials/react.mdx` file (for example).

website/docs/series-types.mdx
```diff
@@ -4,6 +4,9 @@ sidebar_position: 1
 
 # Series types
 
+import { ChartComponent } from "../src/pages/_react/react-components.tsx"
```
edew, line 7: Like me mentioned this morning we can keep the images here for now and replace them with interactive charts in another PR. 💪

.circleci/config.yml
```diff
@@ -118,6 +118,7 @@ jobs:
...
```
edew, line 121: Now we know that this solves the error on CI we can create aliases for saving and restoring the `website/node_modules` directory to and from the cache.

Check out `restore-node-modules-cache` and `save-node-modules-cache` in this file for examples of how we do it for the root `node_modules`

.eslintrc.js
```diff
@@ -2,7 +2,7 @@ function getNamingConventionRules(additionalDefaultFormats = []) {
 	return [
 		{ selector: 'default', format: ['camelCase', ...additionalDefaultFormats], leadingUnderscore: 'forbid', trailingUnderscore: 'forbid' },
 
-		{ selector: 'variable', format: ['camelCase', 'UPPER_CASE'] },
+		{ selector: 'variable', format: ['camelCase', 'UPPER_CASE', ...additionalDefaultFormats] },
```
edew, line 5: Why did we need to add this? 🤔

edew (changes requested):

website/sidebars.js
romfrancois: Agreed with the second part.

As for the first one, if omitted sidebar will disappear when going into Tutorials/react.md

.eslintrc.js
romfrancois: Otherwise I get errors lie

> react-components.tsx:24:14: Variable name `ChartComponent` must match one of the following formats: camelCase, UPPER_CASE [Error/@typescript-eslint/naming-convention]

Happy to check that one with you if the issue comes from my environment

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,152 @@
...
+		},
+		[data]
+	);
+
+	return (
+		<div
+			ref={chartContainerRef}
+		/>
+	);
+};
```
edew, line 115: I think we can simplify this further by

1. showing the example in Javascript (so removing the TypeScript syntax).
2. Combining this code block and the one below.

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,152 @@
...
+			window.addEventListener('resize', handleResize);
+
+			return () => window.removeEventListener('resize', handleResize);
+		},
+		[mainChart]
+	);
+
+	useEffect(
+		() => {
+			mainChart = createChart(chartContainerRef.current, {
```
edew, line 99: `mainChart` will be redeclared each time `ChartComponent` is called (i.e. on each render).

From eslint:
```
Assignments to the 'mainChart' variable from inside React Hook useEffect will be lost after each render.

To preserve the value over time, store it in a useRef Hook and keep the mutable value in the '.current' property. 

Otherwise, you can move this variable directly inside useEffect. (react-hooks/exhaustive-deps)eslint
```

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,152 @@
...
+	time: string;
+	open: number;
+	high: number;
+	low: number;
+	close: number;
+}
+
+export type ChartDataType = NonOHLCDataType | OHLCDataType;
+
+export interface ChartProperties {
```
edew, line 74: Can we include an example that shows how users can `update` series data as well as setting it? I imagine that would be one of the first questions we are asked.

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,152 @@
+---
+id: howTo
```
edew, line 2:
```suggestion
id: react
```

Currently the path for this page is `/lightweight-charts/docs/tutorials/howTo`

edew (changes requested):

website/docs/tutorials/react.mdx
romfrancois: As discuss let's revisit it later.

timocov:
It seems that this tutorial is addressing to #200

justin-pierce:
Would love if the tutorial touched on this: 

https://github.com/tradingview/lightweight-charts/issues/200#issuecomment-951563309

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,136 @@
...
+			chart.current = createChart(chartContainerRef.current, {
+				width: 400,
+				height: 300
+			});
+
+			const newSeries = chart.current.addAreaSeries();
+			setSeries(newSeries);
+
+			window.addEventListener('resize', handleResize);
+			return () => window.removeEventListener('resize', handleResize);
```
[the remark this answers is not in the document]
romfrancois: Yes, that's what I was mentioning this morning.

website/src/pages/_react/react-components.tsx
```diff
@@ -0,0 +1,115 @@
+import { createChart, IChartApi, ISeriesApi, SeriesType } from 'lightweight-charts';
```
romfrancois, line 1: I'll remove this file as it's not used anymore

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,136 @@
...
+	);
+}
+
+```
+
+and you'll have a reusable component that could then be enhanced, tweaked to meet your needs, adding properties and even functionalities.
+
+If you've successfully followed all the steps you should see something similar to
+
+import { SetupComponent } from "../../src/pages/_react/setup-component.tsx"
```
romfrancois, line 134: Rather than using internal code, I'd vote in favour of using a jsfiddle/codepen embedded example here. It would avoid having to maintain the code within this file vs possibility to do it directly on the example without impacting the doc.
+ it would go along the next phase which would be to replace all static images by a real example.

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,133 @@
...
+	const [series, setSeries] = useState();
+
+	const handleResize = () => {
+		chart.current.resize(400, 300);
+	};
+
+	useEffect(
+		() => {
+			if (chart.current !== undefined && series !== undefined) {
+				series.setData(data);
```
timocov, line 71: it seems that this component is designed to be area component rather than a chart one since you're unable to change the type of series in runtime.

I as a consumer of this component might want to do the following from _the outside_ of this component so I might want to have an ability to have this in API:
- change the type of the series in runtime
- add new more series to the chart (or remove them)
- change visible range
- apply options to the chart

How it could be done? I think that this is the first question we will be asked about (and we were in #200 I believe). This example is quite simple and there is a place for this, but it won't work in real-world applications where you might want to have something more.

Probably we can leave this simple example as is, but we need to provide more complex example as well IMO.

timocov (changes requested):

.circleci/config.yml
```diff
@@ -291,6 +318,7 @@ workflows:
           filters: *default-filters
           requires:
             - install-deps
+            - install-deps-website
```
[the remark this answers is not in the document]
edew: It works currently but we could clean it up

1. add the `install-deps` and `install-reps-website` [jobs](https://github.com/romfrancois/lightweight-charts/blob/docusaurus-react-tutorial/.circleci/config.yml#L94) to to the [`website` workflow](https://github.com/romfrancois/lightweight-charts/blob/docusaurus-react-tutorial/.circleci/config.yml#L368) like they are in the [`build-lint-test` workflow](https://github.com/romfrancois/lightweight-charts/blob/docusaurus-react-tutorial/.circleci/config.yml#L300).

2. combine the `checkout-with-deps-website` and `checkout-with-deps` [commands](https://github.com/romfrancois/lightweight-charts/blob/docusaurus-react-tutorial/.circleci/config.yml#L55). (Remove the website one and  add the `restore_cache: *restore-node-modules-cache-website` step to the original.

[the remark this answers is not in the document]
timocov: yeah, I thought about that it seems that we do the same things a bit differently in several places so we could refactor this part a bit

website/docs/tutorials/react-complex-example.mdx
```diff
@@ -0,0 +1,214 @@
...
+
+The following describes a more complex scenario where a user could imagine splitting the responsibilities of the chart between components.
+
+If you want to consult a simpler approach please consult this [example](./react).
+
+:::
+
+:::warning
+
+By following the steps below we assume you're familiar with Lightweight Chart, how to set up a project using it and how to render a chart.
```
timocov, line 17:
```suggestion
By following the steps below we assume you're familiar with Lightweight Charts, how to set up a project using it and how to render a chart.
```

website/docs/tutorials/react-complex-example.mdx
```diff
@@ -0,0 +1,214 @@
...
+
+If not, please follow this [guide](./react).
+
+:::
+
+If you're familiar with Lightweight Chart you probably already know that a _Chart_ is a container that can contain one or more [_Series_](../series-types.md).
```
timocov, line 23:
```suggestion
If you're familiar with Lightweight Charts you probably already know that a _Chart_ is a container that can contain one or more [_Series_](../series-types.md).
```

website/docs/tutorials/react-complex-example.mdx
```diff
@@ -0,0 +1,214 @@
...
+Each _Series_ has its own options (for instance [AreaStyleOptions](/api/interfaces/AreaStyleOptions.md), [LineStyleOptions](/api/interfaces/LineStyleOptions.md), etc) in addition to [Price](../price-scale.md) and/or [Time](../time-scale.md) scale.
```
timocov, line 24:
```suggestion
Each _Series_ has its own options (for instance [AreaStyleOptions](/api/interfaces/AreaStyleOptions.md), [LineStyleOptions](/api/interfaces/LineStyleOptions.md), etc) in addition to [price](../price-scale.md) and/or [time](../time-scale.md) scale.
```

the time scale is common for all series though. Probably we could rephrase it? What did you want to say here?

website/docs/tutorials/react-complex-example.mdx
```diff
@@ -0,0 +1,214 @@
...
+	}, []);
+
+	useEffect(() => {
+		// this effect will be triggered in position 4
+
+		return () => {
+			// this clean up will be triggered in position 2
+		};
+	}, []);
+}
```
timocov, line 65: shouldn't we return a ChildComponent here for better understanding why it is so?

website/docs/tutorials/react-complex-example.mdx
```diff
@@ -0,0 +1,214 @@
...
+		// this effect will be triggered in position 2
+
+		return () => {
+			// this clean up will be triggered in position 4
+		};
+	}, []);
+}
+```
+
+In essence, taking the example above, it means that a ChildComponent (aka Series) would be created first whilst requiring a ParentComponent (aka Chart).
```
timocov, line 86:
```suggestion
In essence, taking the example above, it means that a `ChildComponent` (aka Series) would be created first whilst requiring a `ParentComponent` (aka Chart).
```

?

website/docs/tutorials/react-complex-example.mdx
```diff
@@ -0,0 +1,214 @@
...
+		</Series component 1>
+		<Series component n>
+			<child component />
+		</Series component n>
+	</ChartContainer>
+</Chart component>
+```
+
+where the ChartContainer's role would be needed to attach a DOMElement on which the chart will render.
+ChartContainer will be responsible for creating a **ref**[erence] that will hold functions to handle the lifecycle of the chart.
```
timocov, line 106: > **ref**[erence]

is this a special syntax for something? 🤔

website/docs/tutorials/react-complex-example.mdx
```diff
@@ -0,0 +1,214 @@
...
+			{props.children}
+		</Context.Provider>
+	);
+}
+```
+
+By considering all the above you could end up with Chart/Series components looking like the following
+
+import CodeBlock from '@theme/CodeBlock';
+import Code from '!!raw-loader!./../../src/pages/_react/complex-example.jsx';
```
timocov, line 205: btw do need to specify an extension for imports in mdx files? it seems that we don't this in ts/tsx files not js?

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,71 @@
+---
+id: react
+title: How to use Lightweight Charts with React
+sidebar_position: 0
+---
+
+:::info
+
+The following only describes a relatively simple example that only renders an [areaSeries](../series-types.md#area) that could be tweaked to render any other type of [series](../series-types.md).
```
timocov, line 9:
```suggestion
The following only describes a relatively simple example that only renders an [area series](../series-types.md#area) that could be tweaked to render any other type of [series](../series-types.md).
```

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+For a more complete scenario where you could wrap Lightweight Chart into a component having sub components, please consult this [example](./react-complex-example).
```
timocov, line 11:
```suggestion
For a more complete scenario where you could wrap Lightweight Charts into a component having sub components, please consult this [example](./react-complex-example).
```

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+:::
+
+On this page you will learn how to easily use Lightweight Charts with React.
+
+The below procedure will use [Parcel](https://github.com/brandiqa/react-parcel-starter) but the dependencies used in package.json will also be provided to start from scratch an empty project.
+
+### Setup your React project - using Parcel
```
timocov, line 19: is there a reason why it starts from 3rd level of heading instead of 1st (or 2nd at least)?

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,71 @@
...
+// package.json
+...
+"dependencies": {
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1"
+},
+"devDependencies": {
+	"@babel/core": "^7.10.2",
+	"@babel/preset-env": "^7.10.2",
+	"@babel/preset-react": "^7.10.1",
```
timocov, line 48:
```suggestion
    "@babel/core": "^7.10.2",
    "@babel/preset-env": "^7.10.2",
    "@babel/preset-react": "^7.10.1",
```

tabs vs spaces

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,71 @@
...
+}
+```
+and then
+```
+npm install
+```
+
+### How to create your own Charting React component
+
+The example _React component_ on this page may not fit your requirements completely. Creating a general purpose declarative wrapper for Lightweight Chart's imperative API is a challenge, but hopefully you can adapt this example to your use case.
```
timocov, line 58:
```suggestion
The example _React component_ on this page may not fit your requirements completely. Creating a general purpose declarative wrapper for Lightweight Charts' imperative API is a challenge, but hopefully you can adapt this example to your use case.
```

website/src/pages/_react/chart-component.jsx
```diff
@@ -0,0 +1,69 @@
...
+import React, { useEffect, useRef, useState } from 'react';
+
+export const ChartComponent = () => {
+	const chart = useRef();
+	const chartContainerRef = useRef();
+
+	const [series, setSeries] = useState();
+
+	const handleResize = () => {
+		chart.current.resize(400, 300);
```
timocov, line 11: hm, should we handle it at all if we use static size all the time? 🤔

website/src/pages/_react/chart-component.jsx
```diff
@@ -0,0 +1,69 @@
+import { createChart } from 'lightweight-charts';
+import React, { useEffect, useRef, useState } from 'react';
+
+export const ChartComponent = () => {
+	const chart = useRef();
```
timocov, line 5: do we need to use a ref for this simple case? or it is not react-way to create local variables?

website/src/pages/_react/complex-example.jsx
```diff
@@ -0,0 +1,128 @@
...
+			{container && <ChartContainer {...props} container={container} />}
+		</div>
+	);
+}
+
+export const ChartContainer = forwardRef((props, ref) => {
+	const context = useRef({
+		api() {
+			if (!this._api) {
+				// eslint-disable-next-line no-unused-vars
```
timocov, line 50: will this line be displayed on the page as well?

website/src/pages/_react/complex-example.jsx
```diff
@@ -0,0 +1,128 @@
...
+				const { children, container, ...rest } = props;
+				this._api = createChart(container, rest);
+				this._api.timeScale().fitContent();
```
timocov, line 53: probably we could omit this here just to make it simpler?

website/src/pages/_react/complex-example.jsx
```diff
@@ -0,0 +1,128 @@
...
+
+	useImperativeHandle(ref, () => context.current.api(), []);
+
+	return (
+		<Context.Provider value={context.current}>
+			{props.children}
+		</Context.Provider>
+	);
+});
+ChartContainer.displayName = 'ChartContainer';
```
timocov, line 85: is this react thing?

website/src/pages/_react/chart-component.jsx
```diff
@@ -0,0 +1,69 @@
...
+	{ time: '2018-12-27', value: 28.89 },
+	{ time: '2018-12-28', value: 25.46 },
+	{ time: '2018-12-29', value: 23.92 },
+	{ time: '2018-12-30', value: 22.68 },
+	{ time: '2018-12-31', value: 22.67 },
+];
+
+export function App() {
+	return (
+		<ChartComponent data={data}></ChartComponent>
```
tv-aponomarev, line 67: The data is passed to the component, but it is not used in the component:
```
export const ChartComponent = () => {
```
Maybe `ChartComponent` could be declared like this:
```
export function ChartComponent({ data }) {

}
```

website/src/pages/_react/chart-component.jsx
```diff
@@ -0,0 +1,69 @@
...
+			chart.current = createChart(chartContainerRef.current, {
+				width: 400,
+				height: 300,
+			});
+
+			const newSeries = chart.current.addAreaSeries();
+			newSeries.setData(data);
+			setSeries(newSeries);
+
+			window.addEventListener('resize', handleResize);
```
tv-aponomarev, line 34: `handleResize` can be declared inside the effect as it is not used outside. It's better to leave as much as possible inside the effect to make it easier to find dependencies to re-run the effect.

website/src/pages/_react/chart-component.jsx
```diff
@@ -0,0 +1,69 @@
...
+	useEffect(
+		() => {
+			chart.current = createChart(chartContainerRef.current, {
+				width: 400,
+				height: 300,
+			});
+
+			const newSeries = chart.current.addAreaSeries();
+			newSeries.setData(data);
+			setSeries(newSeries);
```
tv-aponomarev, line 32: I think there is no need to keep the series in state. The data will be set at initialization and there is no need to keep track of the series as an effect dependency. Only the data can be tracked as a dependency to react to data changes and update the series. Therefore, the series can be stored in ref and not trigger a re-render.

website/src/pages/_react/complex-example.jsx
```diff
@@ -0,0 +1,128 @@
...
+		currentRef.api();
+	}, []);
+
+	useLayoutEffect(() => {
+		const currentRef = context.current;
+
+		// eslint-disable-next-line no-unused-vars
+		const { children, container, ...rest } = props;
+		currentRef.api().applyOptions(rest);
+	}, [props]);
```
tv-aponomarev, line 75: I think there is no need to use props as a dependency. This is close to "updating on every render" and the dependency array can be omitted entirely anywhere where properties are used as a dependency.

website/src/pages/_react/complex-example.jsx
```diff
@@ -0,0 +1,128 @@
...
+	useImperativeHandle,
+	useLayoutEffect,
+	useRef,
+	useState,
+} from 'react';
+
+const Context = createContext();
+
+export const App = () => {
+	const series1 = useRef(null);
```
tv-aponomarev, line 16: Maybe we can add code to demonstrate how this ref can be used.

website/docs/tutorials/react-complex-example.mdx
romfrancois: I wanted to highlight that you could specify options when creating your serie but that you could also interact/change some options for both the price & time scale.

website/docs/tutorials/react-complex-example.mdx
romfrancois: No I just tried to mix both the React wording "ref" by making it sound like a "reference" to something.
Just specifying "ref" sounded maybe too technical.

website/docs/tutorials/react-complex-example.mdx
romfrancois: Good catch will remove.

website/docs/tutorials/react.mdx
romfrancois: I didn't use 1 level because it didn't build the TOC on the right hand side of the page.
Will at least change it from level 3 to 2.

Screenshot 2022-02-04 at 11 40 44

website/docs/tutorials/react.mdx
```diff
@@ -0,0 +1,71 @@
...
+```
+git clone git@github.com:brandiqa/react-parcel-starter.git lwc-react
+cd lwc-react
+npm install
+```
+
+To run your project simply
+
+```
+// The following will create a web page accessible by default on http://localhost:1234/
```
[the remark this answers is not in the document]
romfrancois: Looks better indeed.

website/src/pages/_react/complex-example.jsx
romfrancois: It will yes.

website/src/pages/_react/complex-example.jsx
romfrancois: I'll remove it; I added it to make it more realistic but will remove it.

website/src/pages/_react/complex-example.jsx
romfrancois: I didn't know this one either but linter was complaining about it
```
Component definition is missing display nameeslint[react/display-name](https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/display-name.md)
```

website/src/pages/_react/chart-component.jsx
romfrancois: Good catch!

website/src/pages/_react/chart-component.jsx
romfrancois: Removed as unnecessary.

website/src/pages/_react/advanced-react-example.jsx
```diff
@@ -0,0 +1,162 @@
...
+
+		return () => currentRef.free();
+	}, []);
+
+	useLayoutEffect(() => {
+		const currentRef = context.current;
+		// eslint-disable-next-line no-unused-vars
+		const { chilren, data, ...rest } = props;
+		currentRef.api().applyOptions(rest);
+	}, []);
```
tv-aponomarev, line 152: An empty array means this will only be called on mount. The array needs to be completely excluded to update the parameters on every render as we don't want to detect real changes in the properties.

timocov:
Things to fix:

- scrollbars of the chart on windows
- borders inside the table on the chart
- add handling resize in advanced example
- (probably) fill the tutorials index page with some content
- Not for this PR but to fix later #1001

romfrancois:
@timocov would you mind checking both
> scrollbars of the chart on windows
> borders inside the table on the chart

website/src/pages/_react/simple-react-example.jsx
```diff
@@ -0,0 +1,58 @@
...
+			};
+
+			const chart = createChart(chartContainerRef.current, {
+				width: chartContainerRef.current.clientWidth,
+				height: 300,
+			});
+			chart.timeScale().fitContent();
+
+			const newSeries = chart.addAreaSeries();
+			newSeries.setData(data);
```
timocov, line 21: it seems that `data` here is a global object rather than what you're getting in `props`.

website/src/pages/_react/simple-react-example.jsx
```diff
@@ -0,0 +1,58 @@
+import { createChart } from 'lightweight-charts';
+import React, { useEffect, useRef } from 'react';
+import styles from '../chart.module.css';
```
timocov, line 3: Is this related to the example? Couldn't we set it from the outside of this example code? Otherwise for consumers it will make no sense what is that

website/src/pages/_react/advanced-react-example.jsx
```diff
@@ -0,0 +1,179 @@
...
+	forwardRef,
+	useCallback,
+	useContext,
+	useEffect,
+	useImperativeHandle,
+	useLayoutEffect,
+	useRef,
+	useState,
+} from 'react';
+import styles from '../chart.module.css';
```
timocov, line 13: the same for styles

website/src/pages/chart.module.css
```diff
@@ -0,0 +1,8 @@
+.ChartContainer * {
```
timocov, line 1: do we re-use these styles on the landing page?

timocov (changes requested):

website/docs/tutorials/react/02-advanced.mdx
```diff
@@ -0,0 +1,231 @@
...
+import Code from '!!raw-loader!@site/src/pages/_react/advanced-react-example';
+
+<CodeBlock className="language-jsx">{Code}</CodeBlock>
+
+The code above will produce a Series of type `line` in a 600x400 Chart.
+Given a `series1` reference is created to be passed to the Series component you could reuse that object via `series1.current.[any function applicable on Series]`.
+
+For instance and as shown below `series1.current.update(new data)` is used upon clicking on the button.
+
+import { App } from "@site/src/pages/_react/advanced-react-example"
```
timocov, line 224:
```suggestion
import { App } from '@site/src/pages/_react/advanced-react-example';
```

website/docs/tutorials/react/01-simple.mdx
```diff
@@ -46,11 +46,9 @@ and you'll have a reusable component that could then be enhanced, tweaked to mee
 
 If you've successfully followed all the steps you should see something similar to
 
-import { App } from "@site/src/pages/_react/simple-react-example"
-import styles from '../chart.module.css';
+import { App } from '@site/src/pages/_react/simple-react-example'
+import styles from '../../../src/pages/chart.module.css';
```
timocov, line 50:
```suggestion
import styles from '@site/src/pages/chart.module.css';
```

website/docs/tutorials/react/01-simple.mdx
```diff
@@ -46,11 +46,9 @@ and you'll have a reusable component that could then be enhanced, tweaked to mee
 
 If you've successfully followed all the steps you should see something similar to
 
-import { App } from "@site/src/pages/_react/simple-react-example"
-import styles from '../chart.module.css';
+import { App } from '@site/src/pages/_react/simple-react-example'
```
timocov, line 49:
```suggestion
import { App } from '@site/src/pages/_react/simple-react-example';
```

website/docs/tutorials/react/02-advanced.mdx
```diff
@@ -216,16 +217,14 @@ import Code from '!!raw-loader!@site/src/pages/_react/advanced-react-example';
 
 <CodeBlock className="language-jsx">{Code}</CodeBlock>
 
-The code above will produce a Series of type `line` in a 600x400 Chart.
+The code above will produce a Series of type `line` Chart.
```
timocov, line 220:
```suggestion
The code above will produce a line series.
```

website/docs/tutorials/react/02-advanced.mdx
```diff
@@ -216,16 +217,14 @@ import Code from '!!raw-loader!@site/src/pages/_react/advanced-react-example';
...
 Given a `series1` reference is created to be passed to the Series component you could reuse that object via `series1.current.[any function applicable on Series]`.
 
 For instance and as shown below `series1.current.update(new data)` is used upon clicking on the button.
 
-import { App } from "@site/src/pages/_react/advanced-react-example"
-import styles from '../chart.module.css';
+import { App } from '@site/src/pages/_react/advanced-react-example'
+import styles from '../../../src/pages/chart.module.css';
```
timocov, line 226:
```suggestion
import { App } from '@site/src/pages/_react/advanced-react-example';
import styles from '@site/src/pages/chart.module.css';
```

website/src/pages/index.tsx
```diff
@@ -45,7 +46,7 @@ function isScreenMinimumWidthForDisplayingChart(): boolean {
 function HeroChart(): JSX.Element {
 	const ref = React.useRef<HTMLDivElement>(null);
 	const layout = useThemeAwareLayoutOptions();
-	const [chartContainerClassName, setChartContainerClassName] = React.useState<string>(styles.HeroChartContainer);
+	const [chartContainerClassName, setChartContainerClassName] = React.useState<string>(`${styles.HeroChartContainer}`);
```
timocov, line 49:
```suggestion
	const [chartContainerClassName, setChartContainerClassName] = React.useState<string>(styles.HeroChartContainer);
```

timocov (approved):

PR #972, opened 2022-01-12
feat: add option to activate scroll on touch end event

### LINKED ISSUE #106: Get rid of detecting touch/non-touch devices
Currently `MouseEventHandler` and widgets like `PaneWidget` [use detection of the device type](https://github.com/tradingview/lightweight-charts/blob/6f132374ed82b31f3b1e66837b2956470057a50d/src/gui/mouse-event-handler.ts#L4) to determine what type of event subscribe for and what to do in several cases (long tap event, dragging, tracking mode).

In a lot of cases we don't need to detect it, we can subscribe on all needed event with separate handles and there we'll know what's happened (touch event or mouse one). It requires some refactoring in MouseEventHandler and detecting that an event is emulated from original one (mouse event from touch for instance).

Don't forget about [this](https://github.com/tradingview/lightweight-charts/blob/6f132374ed82b31f3b1e66837b2956470057a50d/src/gui/mouse-event-handler.ts#L41-L42).

I believe in the PR for this issue we can remove https://github.com/tradingview/lightweight-charts/blob/6f132374ed82b31f3b1e66837b2956470057a50d/src/gui/support-touch.ts module at all.

Related links:

- https://developer.mozilla.org/en-US/docs/Web/API/InputDeviceCapabilities/firesTouchEvents

### LINKED ISSUE #830: Quick tracking mode
At the moment the tracking mode (on mobile when if long press on the chart to "select" bar and price) might be enabled by long press and then you need to touch end and then start to touch move to move the crosshair. 

Sometimes it might be useful to use kind of "fast tracking mode" so you can enable the tracking mode by long press and then you should be able to touch move to select bar. After you touch end the tracking mode should be finished.

Ffloriel:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #106, fixes #830
- [ ] Includes tests
- [x] Documentation update // in type

**Overview of change:**

- Update MouseEventHandler, separating the mouse and touch events (related to #106)
- Add _quick tracking option_, to activate scroll on touch end event instead of having to click after to reactivate the scroll.


src/gui/support-touch.ts
```diff
@@ -27,20 +27,3 @@ function getMobileTouch(): boolean {
 }
 
 export const mobileTouch = getMobileTouch();
```
timocov, line 29: it seems that the only place where we use this variable as-is is `PaneWidget.mouseEnterEvent` method, do we still need to have it there since we migrated to event-based detection rather than browser-feature detection. probably we could remove it from there and rename this variable and the value?

tests/e2e/coverage/coverage-config.ts
```diff
@@ -1,2 +1,2 @@
-export const expectedCoverage = 89.5;
```
timocov, line 1: that's sad but understandable 🙂

src/typings/_global-types/index.d.ts
```diff
@@ -23,3 +23,18 @@ declare const enum MouseEventButton {
...
+interface UIEvent {
+	/** https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/sourceCapabilities */
+	sourceCapabilities?: InputDeviceCapabilities;
+}
+
+interface Window {
+	chrome?: {
+		runtime: unknown;
+	};
+}
```
timocov, line 40: I'd suggest to extract these types to a different folder like `src/types/dom-not-standarted/index.d.ts` since this file is more about global helpers but not related to the existent API so I'd prefer to extract this DOM extensions from it.

src/gui/support-touch.ts
timocov: another please where this variable is still using is `src/gui/last-touch-or-mouse-event.ts`, but it seems that we shouldn't rely on a mobile touch there. it should be more whether a touch is supported or not

src/gui/last-touch-or-mouse-event.ts
```diff
@@ -0,0 +1,19 @@
+import { mobileTouch } from './support-touch';
+
+let lastEventIsTouchValue = mobileTouch;
+
+export function lastEventIsTouch(): boolean {
+	return lastEventIsTouchValue;
+}
+
+export function lastEventIsMouse(): boolean {
+	return lastEventIsTouchValue;
```
timocov, line 10: shouldn't it be `!lastEventIsTouchValue` instead? but it seems that this function is not used anywhere so we can remove it 😄

src/helpers/events.ts
```diff
@@ -0,0 +1,17 @@
+import { isChrome } from './browsers';
+
+export function preventScrollByWheelClick(el: HTMLElement): void {
+	if (!isChrome()) {
+		return;
+	}
+
+	el.addEventListener('mousedown', (e: MouseEvent) => {
+		if (e.button === MouseEventButton.Middle) {
+			// prevent incorrect scrolling event
```
timocov, line 10: @kirchet do you remember what is the reason why we need to ignore middle-mouse-button scroll on chrome only?

src/model/chart-model.ts
```diff
@@ -63,6 +63,18 @@ export interface HandleScrollOptions {
...
+	 * Activate scroll on touch end event.
+	 *
+	 * By default, mobile users will long press to deactivate the scroll and have the ability to check values and dates.
+	 * Another press is required to activate the scroll, be able to move left/right, zoom, etc.
+	 *
+	 * When enabled, the chart will activate the scroll on touch end event.
+	 *
+	 * @defaultValue `false`
+	 */
+	activateScrollOnTouchEnd: boolean;
```
timocov, line 77: Hmm, for me it looks a bit confusing. Whether every touch end should activate scroll? What does "activate scroll" mean? Probably it shouldn't be a part of scroll options but another option on the root layer describing the behaviour. Maybe we can add something like `touchTrackingMode` and there 2 options: Quick, Long (not sure about names though). There we can also describe what does "tracking mode" mean to users. @kirchet what do you think?

src/gui/mouse-event-handler.ts
```diff
@@ -203,75 +299,200 @@ export class MouseEventHandler implements IDestroyable {
...
+			yOffset: yOffset,
+			manhattanDistance: manhattanDistance,
+		};
+	}
+
+	/**
+		* In Firefox mouse events dont't fire if the mouse position is outside of the browser's border.
+		* To prevent the mouse from hanging while pressed we're subscribing on the mouseleave event of the document element.
+		* We're subscribing on mouseleave, but this event is actually fired on mouseup outside of the browser's border.
+		*/
```
timocov, line 360:
```suggestion
	/**
	 * In Firefox mouse events dont't fire if the mouse position is outside of the browser's border.
	 * To prevent the mouse from hanging while pressed we're subscribing on the mouseleave event of the document element.
	 * We're subscribing on mouseleave, but this event is actually fired on mouseup outside of the browser's border.
	 */
```

looks like tabs issues here?

src/model/chart-model.ts
timocov: or probably it might be even inclusive like "allowQuickTrackingMode" so if you activate tracking mode but didn't move the finger then it shouldn't disabled after touchend event

timocov (changes requested):

src/gui/support-touch.ts
kirchet: Hm, not sure that it can be removed. Maybe in some cases it's needed to have "correct" initial `lastEventIsTouch()` value

src/model/chart-model.ts
kirchet: I think we should add a new field like `trackingMode: TrackingModeOptions`. So we can extend it in the future without breaking changes

```
export const enum ExitTrackingModeBehaviour {
	OnTouchEnd,
	OnNextTap,
}

export interface TrackingModeOptions {
	exitMode: ExitTrackingModeBehaviour;
}
```

kirchet (changes requested):

src/helpers/events.ts
kirchet: Because there was a bug in the Chrome only. After zooming a page and middle button clicking the cursor mode incorrectly changed to "native scroll by midde button mode"

src/model/chart-model.ts
kirchet: I think we should add a new field like `trackingMode`. So we can extend in in the future.

```
export const enum TrackingExitMode {
    OnTouchEnd,
    OnNextTap,
}
export interface TrackingModeOptions {
    exitMode: TrackingExitMode;
}
```

src/helpers/events.ts
kirchet: Not sure if it's needed in lightweight-charts

timocov: @Ffloriel could you check the case above please?

src/gui/support-touch.ts
kirchet: Ah, yes, I agree

kirchet: Seems that it's not used now anywhere, so I'm resolving this thread

src/gui/last-touch-or-mouse-event.ts
kirchet: The wnole module is removed, so this thread is resolved

src/helpers/events.ts
Ffloriel: Behavior is the same on Chrome and Firefox. I don't see the issue with the middle button.

Ffloriel: nevermind, I can see the issue on windows.

src/gui/pane-widget.ts
```diff
@@ -270,6 +270,7 @@ export class PaneWidget implements IDestroyable, MouseEventHandlers {
 	}
 
 	public pressedMouseMoveEvent(event: MouseEventHandlerMouseEvent): void {
+		this._startTrackPoint = null;
```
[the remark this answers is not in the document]
kirchet: Or from all mouse handlers because its name is `_onMouseEvent` 🙂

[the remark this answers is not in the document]
Ffloriel: :thinking_face: ok, I see what you mean. We only preventing the scroll for pressedMouseMoveEvent so that's why I originally only reset it here. Other events should not be affected but the logic makes sense to do it also for the other ones

src/helpers/events.ts
kirchet: I think it's not directly related to this PR, so if there is an issue we can fix it in another PR

kirchet (approved):

src/model/chart-model.ts
```diff
@@ -175,6 +175,39 @@ export type VisiblePriceScaleOptions = PriceScaleOptions;
...
+ */
+export const enum TrackingExitMode {
+	/**
+	 * Tracking Mode will be deactivated on touch end event.
+	 */
+    OnTouchEnd,
+	/**
+	 * Tracking Mode will be deactivated on the next tap event.
+	 */
+    OnNextTap,
```
timocov, line 192:
```suggestion
	OnTouchEnd,
	/**
	 * Tracking Mode will be deactivated on the next tap event.
	 */
	OnNextTap,
```

looks like tabs vs spaces issue

src/model/chart-model.ts
```diff
@@ -175,6 +175,39 @@ export type VisiblePriceScaleOptions = PriceScaleOptions;
...
+/**
+ * Represent options for the tracking mode's behavior.
+ *
+ * Mobile users will not have the ability to see the values/dates like they do on desktop.
+ * To see it, they should enter the tracking mode. The tracking mode will deactivate the scrolling
+ * and make it possible to check values and dates.
+ */
+export interface TrackingModeOptions {
+	// eslint-disable-next-line tsdoc/syntax
+	/** @inheritdoc TrackingExitMode
```
timocov, line 204: what does this thing for the docs? should we apply it for other options as well (in different PR of course)?

src/model/chart-model.ts
```diff
@@ -175,6 +175,39 @@ export type VisiblePriceScaleOptions = PriceScaleOptions;
...
+	 *
+	 * @defaultValue `TrackingExitMode.OnNextTap`
```
timocov, line 206:
```suggestion
	 * @defaultValue {@link TrackingExitMode.OnNextTap}
```

shouldn't it be like that instead?

src/model/chart-model.ts
```diff
@@ -175,6 +175,39 @@ export type VisiblePriceScaleOptions = PriceScaleOptions;
  */
 export type OverlayPriceScaleOptions = Omit<PriceScaleOptions, 'visible' | 'autoScale'>;
 
+/**
+ * Determine how to exit the tracking mode.
+ *
+ * By default, mobile users will long press to deactivate the scroll and have the ability to check values and dates.
+ * Another press is required to activate the scroll, be able to move left/right, zoom, etc.
+ */
+export const enum TrackingExitMode {
```
timocov, line 184: shouldn't it be TrackingModeExitMode? 😂

src/model/chart-model.ts
Ffloriel: yes, it's better if it's a link

src/model/chart-model.ts
Ffloriel: it basically copies the description. Avoid having to duplicate it. There's an issue with the eslint plugin though (https://github.com/microsoft/tsdoc/issues/220), because typedoc and tsdoc have a different syntax for it.

timocov: Ah, I see, thanks! At least we can format it properly, or it is formatted so because we need to disable linter for this line? 😂

Ffloriel: ahah yep, unfortunately it's to disable the linter

timocov (approved):

PR #976, opened 2022-01-14
Fix partially hidden timescale marks when scrolling and scaling are disabled

### LINKED ISSUE #947: Labels might be cut off when disabling scale and scroll
**Lightweight Charts Version:** 3.7.0

**Steps/code to reproduce:** refer to this codesandbox https://codesandbox.io/s/unruffled-benz-z9f5p

**Actual behavior:** The last label on the time axis will be hidden behind a non-transparent canvas, if it happens to overflow the canvas for the time scale

**Expected behavior:** The time scale labels would be overflowing their canvas, so they would display their whole content

**Screenshots:**

**CodeSandbox/JSFiddle/etc link:** https://codesandbox.io/s/unruffled-benz-z9f5p

With reference to 
#226 and #835

edew:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #947
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

When the time scale is deciding if a mark's coordinate needs to be realigned we need to handle the case where all scrolling and scaling has been disabled because it is not possible for a user to move a partially hidden mark into view.

I couldn't see a nicer api to check all the conditions in `_isAllScalingAndScrollingDisabled`, but maybe there is one I missed?


src/model/time-scale.ts
```diff
@@ -727,6 +731,18 @@ export class TimeScale {
...
+	private _isAllScalingAndScrollingDisabled(): boolean {
+		return !this._model.options().handleScroll.horzTouchDrag
+			&& !this._model.options().handleScroll.mouseWheel
+			&& !this._model.options().handleScroll.pressedMouseMove
+			&& !this._model.options().handleScroll.vertTouchDrag
+			&& !this._model.options().handleScale.axisDoubleClickReset
+			&& !this._model.options().handleScale.axisPressedMouseMove.price
+			&& !this._model.options().handleScale.axisPressedMouseMove.time
+			&& !this._model.options().handleScale.mouseWheel
+			&& !this._model.options().handleScale.pinch;
```
timocov, line 743: it seems that it might be easy to forget to add anything to this list. @kirchet maybe you might have an idea whether we can handle it differently.

```suggestion
		const { handleScroll, handleScale } = this._model.options();
		return !handleScroll.horzTouchDrag
			&& !handleScroll.mouseWheel
			&& !handleScroll.pressedMouseMove
			&& !handleScroll.vertTouchDrag
			&& !handleScale.axisDoubleClickReset
			&& !handleScale.axisPressedMouseMove.time
			&& !handleScale.mouseWheel
			&& !handleScale.pinch;
```

I'd change it to this but it doesn't change the situation a lot anyway (please note that I removed checking for `axisPressedMouseMove.price` since it seems that that's OK if you're able to change price range so why should we disable this feature in this case)

kirchet: > maybe you might have an idea whether we can handle it differently

I don't know, maybe we can add a helper that walks through all nestend properties and checks them for a provided predicate.

timocov: I thought about that but it looks overcomplicated for me. Ok, let's apply the changes above and keep as is for now

timocov (approved):

PR #977, opened 2022-01-17
Fixed incorrectly applied fitting content if a price axis changed its width

### LINKED ISSUE #966: timeScale.fitContent is not working correctly
**Lightweight Charts Version:** 3.7.0 (but 3.5.0 is working OK)

**Code to reproduce:** https://jsfiddle.net/iksent/8akv3x59/

**Actual behavior:** executing `.timeScale().fitContent()` (after changing series data) sometimes not fitting the chart fully.

**Screen recording:**
https://user-images.githubusercontent.com/32956521/148347730-4f0dbccd-5e5a-4104-8db8-30fc4b64f5f3.mov

**Expected behavior:** `fitContent` must fit the content correctly.

Ffloriel:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #966
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

`fitContent()` might cause the data of the price axis to change, causing optimal width of the price axis widget to change, causing time scale to adjust its width. It will make the bar spacing previously set to fit the content incorrect.
Add a check to call fitContent again.


timocov (approved):

timocov:
LGTM but some of tests are failed, which is not expected so let's try to find out what happened there.

Ffloriel:
For the failing tests, most of them are related to the price-scale width flickering.

_update-removed-series-data_
In master branch, we ask for a full update to be trigger 100ms after. The test is changing the data every 50ms.
In the master branch, we are adjusting the price scale width three times. The last time adjusts it back to 46px.
In the new branch, we are adjusting the price scale width twice. The last time adjusts it to 52px.

In the master, we trigger the calculation of the price scale optimal width 100ms after, so it’s calculated with the data that determined the optimal width to be 46.

In the new branch, it only sees the current data, which correspond to the optimal width of 52px. It doesn’t try to adjust the size of the optimal width after because there have been no full update after that (contrary to the master branch).

A way to see that is to set the change in data to 150ms instead of 50ms. Then, the behaviour of master is the same as our new branch.
Because it will not use the data+1 in the current full update.

_fit-content-pricescale-changes_
The test I added, should be failing.

_scroll-to-position-after-set-visible-range_
Independent of the scrollToPosition, it has a difference in price scale width when you add the data. Like fit-content-pricescale-changes. You can see a flicker btw on the master branch test, the price-scale width changes quickly.

_set-visible-range_
Fail like fit-content-pricescale-changes. On local, the price scale is larger for the new branch. Price scale  width flickers on the master branch.

_no-autoscale_
The first paint will show the price scale with a smaller width, so the data on the price scale is a bit lower than in the new branch (ax+b function on the graph). Then, the price scale width adjusts correctly.

timocov:
Many thanks @Ffloriel for this detailed explanation. Looks like all changes are expected and for especially for `no-autoscale` it is even better since we applied autoscale for the latest price axis width.

As for the issue with 100ms I think we could consider to do the following: when a light+ update is happened and there is an autoscale invalidation _and_ no time scale invalidation we need to save time scale logical range before applying auto scale and then check whether a width of price axis is changed and if it is so then set the visible logical range back to the previous value. It might allow us to avoid flickering and infinite changing widths of time & price scales (since they might affect each other).

PR #983, opened 2022-01-21
Add documentation website landing page

edew:
**Type of PR:** enhancement

**PR checklist:**

- [ ] Includes tests
- [x] Documentation update

**Overview of change:**

This PR adds a fresh landing page for the documentation website.

* I'm styling the svg icons in the page with CSS so they change correctly for light/dark mode, but the logo in the navbar can't be styled like this as Docusaurus loads it in an `img` tag so the navbar loads two different static svg files.
* The chart data doesn't match the design perfectly. Do we think that's OK?


package.json
```diff
@@ -59,6 +59,7 @@
     "eslint-plugin-prefer-arrow": "~1.2.1",
     "eslint-plugin-tsdoc": "~0.2.14",
     "eslint-plugin-unicorn": "~39.0.0",
+    "eslint-plugin-react": "^7.28.0",
```
timocov, line 62:
```suggestion
    "eslint-plugin-react": "~7.28.0",
```

website/docusaurus.config.js
```diff
@@ -140,14 +140,10 @@ const config = {
 				title: 'Lightweight Charts',
 				logo: {
 					alt: 'Lightweight Charts Logo',
-					// please note that there is no such file in static/img folder in git
-					// but this file will be copied by this config (see above fs.copyFileSync call)
-					// this is just to avoid duplicating the same files in the repo
-					// and at the same time to avoid loading files from other domains (like github)
-					// (loading from other domains might take some time and the page might blink especially while loading images)
-					src: 'img/lightweight-charts-logo.svg',
```
timocov, line 148: so we need to remove a code above which copies this file, see a statement after comment `copy logo file to static folder so we can refer to it in the config` above

website/package.json
```diff
@@ -13,11 +13,13 @@
   },
   "dependencies": {
     "@docusaurus/core": "2.0.0-beta.13",
+    "@docusaurus/module-type-aliases": "^2.0.0-beta.14",
```
timocov, line 16:
```suggestion
    "@docusaurus/module-type-aliases": "2.0.0-beta.13",
```

website/package.json
```diff
@@ -13,11 +13,13 @@
...
     "@docusaurus/preset-classic": "2.0.0-beta.13",
     "@docusaurus/theme-search-algolia": "2.0.0-beta.13",
     "@tsconfig/docusaurus": "~1.0.4",
     "cross-env": "~7.0.3",
     "docusaurus-plugin-typedoc": "0.16.6",
+    "lightweight-charts": "^3.7.0",
```
timocov, line 22:
```suggestion
    "lightweight-charts": "~3.7.0",
```

website/docusaurus.config.js
```diff
@@ -140,14 +140,10 @@ const config = {
...
-					width: 48,
-					height: 32,
+					src: 'img/logo-dark.png',
+					srcDark: 'img/logo-light.png',
```
timocov, line 144: maybe we can use svg icon instead? not too sure that use raster images as logos with predefined static (quite small) size is a good idea though

website/src/pages/index.module.css
```diff
@@ -0,0 +1,300 @@
...
+	flex-direction: column;
+	align-items: center;
+}
+
+.HeroChartContainer {
+	display: none;
+}
+
+.HeroContainer svg path {
+	stroke: var(--LargeIconSvgFillColor);
```
timocov, line 12: shouldn't it be defined in an svg file instead? 🤔

website/static/img/logo.svg
```diff
@@ -0,0 +1,8 @@
+<svg class="logo-svg" width="92" height="56" viewBox="0 0 92 56" fill="none" xmlns="http://www.w3.org/2000/svg">
```
timocov, line 1: please apply svgo for this file (or internal svg optimization tool - see `tools/svg-tools`)

website/src/pages/index.module.css
```diff
@@ -0,0 +1,300 @@
...
+		display: unset;
+		/* width: 100%; */
+		height: 40vh;
+		width: 100vw;
+	}
+
+	.HeroTextContainer {
+		position: absolute;
+		z-index: 2;
+		background: var(--HeroTextBackgroundColor);
```
timocov, line 220: is it expected that bg color is changed when a width is changed? I'm just curios, I'm not big expert in that 😄

website/src/pages/index.module.css
```diff
@@ -0,0 +1,300 @@
+.HeroContainer {
```
timocov, line 1: since we're going to add more css files to about pages, probably we need to add stylelint to the repo? what do you think?

website/src/pages/index.tsx
```diff
@@ -0,0 +1,243 @@
...
..., time: 1631029368 as UTCTimestamp }, { value: 32.13494929971301, time: 1631634168 as UTCTimestamp }, { value: 32.24176261787218, time: 1632238968 as UTCTimestamp }, { value: 32.23416002963385, time: 1632843768 as UTCTimestamp }, { value: 33.155464487194166, time: 1633448568 as UTCTimestamp }, { value: 32.30858974959849, time: 1634053368 as UTCTimestamp }, { value: 33.240351713622175, time: 1634658168 as UTCTimestamp }, { value: 32.476820033756304, time: 1635262968 as UTCTimestamp }, { value: 33.38322664774607, time: 1635867768 as UTCTimestamp }, { value: 33.112040840762795, time: 1636472568 as UTCTimestamp }, { value: 33.21270434841732, time: 1637077368 as UTCTimestamp }, { value: 32.34952853888793, time: 1637682168 as UTCTimestamp }, { value: 33.309373031504336, time: 1638286968 as UTCTimestamp }, { value: 32.68788509068168, time: 1638891768 as UTCTimestamp }, { value: 32.991148534675084, time: 1639496568 as UTCTimestamp }, { value: 32.019560141931144, time: 1640101368 as UTCTimestamp }, { value: 32.6775781486036, time: 1640706168 as UTCTimestamp }, { value: 31.739487272423506, time: 1641310968 as UTCTimestamp }, { value: 31.957098637385883, time: 1641915768 as UTCTimestamp }] };
```
timocov, line 11: would you mind to format it in several lines or/and even extract it to a separate file please?

website/src/pages/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+	React.useLayoutEffect(
+		() => {
+			if (!chart) {
+				return;
+			}
+
+			const layout = isDarkTheme
+				? { background: { color: '#000000' }, textColor: 'rgba(248, 249, 253, 1)' }
+				: { background: { color: 'rgba(248, 249, 253, 1)' }, textColor: '#000000' };
```
timocov, line 93: is seems that this is defined twice, maybe we can extract it to a function?

website/src/pages/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+	);
+
+	return (
+		<div className={styles.HeroChartContainer} ref={ref}></div>
+	);
+}
+
+function PerformanceIcon(): JSX.Element {
+	return (
+		<svg width="115" height="103" viewBox="0 0 115 104" fill="none" xmlns="http://www.w3.org/2000/svg">
```
timocov, line 107: is there a reason why we don't extract all these svgs to separate files instead?

website/src/pages/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+						<h3>Flexible styling</h3>
+						<p>Change the standard look & feel to match your style with perfection. There are many premade examples that you can copy & paste.</p>
+					</div>
+				</div>
+			</div>
+		</Layout>
+	);
+}
+
+// eslint-disable-next-line import/no-default-export
```
timocov, line 242: probably for jsx/tsx (btw we we don't use tsx for components?) we need to disable this rule in eslint? we can use either overrides or local eslint config files for that (to specify it only for pages for example)

timocov (changes requested):

website/src/css/custom.css
```diff
@@ -26,3 +50,32 @@
...
+  fill-rule: evenodd;
+}
+
+.main-wrapper {
+  display: flex;
+  justify-content: center;
+  gap: 2rem;
+}
+
+.logo-svg path {
```
subzey, line 75: The inline SVGs can be recolored much easier! Just mark the necessary `fill`s and `stroke`s as `currentcolor` and now the entire svg color can be controlled using a single `color` property.

https://jsfiddle.net/subzey/mwj30uy9/

website/src/pages/index.module.css
```diff
@@ -0,0 +1,300 @@
...
+		flex: 1 1 50%;
+		padding-left: 2rem;
+		padding-right: 2rem;
+	}
+
+	.LargeCard {
+		flex: 0 1 47%;
+	}
+
+	.LargeCard:last-of-type {
```
subzey, line 176: Not really an issue, rather a reminder:

This selector may behave not as you'd expect. `.LargeCard:last-of-type` doesn't mean "The last child of those who have the class LargeCard". It's rather "The last child with the <unspecified> tagname if by chance is has the class LargeCard"

website/src/pages/index.module.css
```diff
@@ -0,0 +1,300 @@
...
+		border-bottom: 1px solid var(--SmallCardBorderColor);
+	}
+
+	.SmallCard:nth-of-type(4) {
+		border-bottom-left-radius: unset;
+		border-left: unset;
+	}
+}
+
+@media (min-width: 1280px) {
```
subzey, line 204: @timocov Could this page be shown inside an `iframe`?

Firefox is known to have fractional viewport sizes on non-integer DPRs. See as a demo: https://bl.ocks.org/subzey/ffa9cfda0efc8e09bb3c19a155d610ca

The workaround is really simple:

```css
@media (min-width: 1279.5px) {
```

(also it helps those poor souls who cannot remember is `min-width` a `<` or `<=`)

timocov: > Could this page be shown inside an iframe?

No, I don't think so. I mean, maybe someone could put it into an iframe, but we aren't going to do this.

website/src/pages/index.module.css
```diff
@@ -0,0 +1,300 @@
...
+		border-top-right-radius: 1rem;
+		border-left: unset;
+	}
+
+	.SmallCard:nth-of-type(3) {
+		border-bottom-left-radius: 1rem;
+		border-bottom: 1px solid var(--SmallCardBorderColor);
+	}
+
+	.SmallCard:nth-of-type(4) {
```
subzey, line 198: The stylization based on the element order may be really painful to maintain. If something is changed in the future the entire block of rules should be checked and fixed.

Also, there can be a stray `span` or `div` elements inserted by libraries and browser extensions.

> Here's an example of such behavior: https://bl.ocks.org/subzey/752f24df13784e4f3517e7237aad33a1. Open it up in Google Chrome and the translate the page into English. The bang sign is red after the translation instead of the username.

website/src/pages/index.module.css
edew: Good point - this can be moved to the definition outside the media query I think.

website/src/pages/index.tsx
edew: No I think they can be extracted. I will do that.

website/src/pages/index.module.css
edew: Yes that's a good idea. Shall we do it in this PR or in another one?

website/src/css/custom.css
```diff
@@ -6,14 +6,40 @@
...
+  --ifm-color-primary: rgb(41, 98, 255);
+  --ifm-color-primary-dark: rgb(30, 83, 229);
+  --ifm-color-primary-darker: rgb(24, 72, 204);
+  --ifm-color-primary-darkest: rgb(20, 62, 179);
+  --ifm-color-primary-light: rgb(49, 121, 245);
+  --ifm-color-primary-lighter: rgb(91, 156, 246);
+  --ifm-color-primary-lightest: rgb(144, 191, 249);
   --ifm-code-font-size: 95%;
+  --ifm-footer-background-color: rgb(255, 255, 255);
+	--PrimaryButtonColor: rgba(41, 98, 255, 1);
```
timocov, line 18: looks like spaces/tabs issue

website/src/css/custom.css
timocov: it looks like this is still valid?

website/src/pages/index.module.css
timocov: we can do it as a separate PR

website/src/pages/index.module.css
```diff
@@ -0,0 +1,297 @@
...
+}
+
+.HeroButtonPrimary {
+	color: #FFFFFF;
+	background: #2962FF;
+}
+
+.HeroButtonPrimary:hover {
+	color: #FFFFFF;
+	background: rgba(30, 83, 229, 1);
```
timocov, line 59: shouldn't colors be a part of the theme? 🤔 I mean color values

website/src/pages/index.module.css
```diff
@@ -0,0 +1,297 @@
...
+
+.RootContainer {
+	max-width: 98%;
+	display: flex;
+	flex-direction: column;
+	gap: 2rem;
+	align-items: center;
+}
+
+@media (min-width: 768px) {
```
timocov, line 162: shouldn't this be the same as below with 0.5?

.eslintrc.js
```diff
@@ -56,6 +56,7 @@ module.exports = {
 	},
 	extends: [
 		'eslint:recommended',
+		'plugin:react/recommended',
```
timocov, line 59: will it work for tsx files as well right?

website/docusaurus.config.js
```diff
@@ -140,14 +140,8 @@ const config = {
...
-					// please note that there is no such file in static/img folder in git
-					// but this file will be copied by this config (see above fs.copyFileSync call)
-					// this is just to avoid duplicating the same files in the repo
-					// and at the same time to avoid loading files from other domains (like github)
-					// (loading from other domains might take some time and the page might blink especially while loading images)
-					src: 'img/lightweight-charts-logo.svg',
-					width: 48,
-					height: 32,
+					src: 'img/logo-light.svg',
+					srcDark: 'img/logo-dark.svg',
```
timocov, line 144: do you know the difference between light and dark icons? is it the color only?

website/docusaurus.config.js
timocov: looks like it is still valid

website/src/pages/index.tsx
timocov: Looks like we need to use https://www.npmjs.com/package/react-svg-loader for svg to make them inlined?

website/src/pages/index.module.css
edew: Good point. I will replace these selectors with specific classes. I think that will be easier to maintain/harder to break accidentally.

website/src/pages/index.tsx
timocov: Oh cool, looks like they have built-in webpack loader https://react-svgr.com/docs/webpack/

website/docusaurus.config.js
edew: Yes, the colours are different.

.eslintrc.js
edew: Yep 👍

website/src/pages/index.module.css
edew: I only extracted vars for colors that are used in multiple rules for now

website/src/pages/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+			orangeSeries.setData(data.orangeData as LineData[]);
+			blueSeries.setData(data.blueData as LineData[]);
+
+			c.timeScale().setVisibleLogicalRange({ from: 1, to: data.orangeData.length - 2 });
+
+			const resizeListener = () => {
+				const { width, height } = container.getBoundingClientRect();
+				c.resize(width, height);
+				c.timeScale().setVisibleLogicalRange({ from: 1, to: data.orangeData.length - 2 });
```
timocov, line 90: See https://tradingview.github.io/lightweight-charts/docs/api/interfaces/TimeScaleOptions#lockvisibletimerangeonresize, it seems that this is what you need instead 🙂

website/docusaurus.config.js
edew: Docusaurus handles the loading of the logo image in the navbar. Because it loads it into a `` we can't style the svg color with css.

timocov (approved):

PR #992, opened 2022-01-28
Fix landing page chart overflow

edew:
**Type of PR:** bugfix

**PR checklist:**

- [ ] Includes tests
- [x] Documentation update

**Overview of change:**

* Fixes overflow of the chart container caused by a built-in Docusaurus style that allows overflow on table elements.

* Fixes chart container itself overflowing out of the available visual width.

* Changes an unnecessary `useLayoutEffect` to `useEffect`

### DIFF
--- a/website/src/pages/index.module.css
+++ b/website/src/pages/index.module.css
@@ -8,6 +8,10 @@
 	display: none;
 }
 
+.HeroChartContainer * {
+	overflow: hidden;
+}
+
 .HeroTextContainer {
 	display: flex;
 	flex-direction: column;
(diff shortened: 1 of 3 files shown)


website/src/pages/index.module.css
```diff
@@ -8,6 +8,10 @@
 	display: none;
 }
 
+.HeroChartContainer * {
```
timocov, line 11:
```suggestion
.HeroChartContainer table
```

shouldn't we do this instead? 🤔 Or there are another nodes where docusaurus' overflow rules are applied?

PR #999, opened 2022-02-02
Update landing page design and fix visual bugs

edew:
**Type of PR:** enhancement

**PR checklist:**

- [ ] Includes tests
- [x] Documentation update

**Overview of change:**

* Switch small cards "Ultra lightweight..." with "Integrating & connecting...".
* Fix navbar and footer borders, fix spacing, fix card borders.
* Use updated logos.
* Exit early from resize events if the chart is not displayed.
* Use a fade in animation to hide the fact that the chart flickers on first render.
* Parse largest size limit and use that on the landing page ("At just 45.5 kilobytes, ...").


romfrancois:
Not being an expert on the subject but I've noticed a few inconsistencies on the rendering for both dark/light themes.

- With the dark mode on submenu's items are really hard to distinguish and it's even harder to get which item is currently being hovered.

bw_submenu

Dont know if accessibility is also part of this ticket but even on my monitor it's hard to see.

- Also, on light theme, there's an inconsistency between the landing page and the other pages while navigating.

wt_submenu
Colours don't seem to be of the same naming.

- Maybe for the design team but some texts are centred while others are aligned to the left.

- Nit pick: should all the rendering in the cards (small & large) be 1-1 relation?

- Probably not linked to that PR and more for design but on dark theme it's kinda to distinguish the separation between small cards

edew:
> Not being an expert on the subject but I've noticed a few inconsistencies on the rendering for both dark/light themes.
> 
> * With the dark mode on submenu's items are really hard to distinguish and it's even harder to get which item is currently being hovered.
> 
> bw_submenu
> 
> Dont know if accessibility is also part of this ticket but even on my monitor it's hard to see.
> 

Yeah I also find the dark theme quite hard to use on my monitor - but I never use dark themes so I thought that might just be me. We can ask for the design to be changed if we all find it hard to use.

> * Also, on light theme, there's an inconsistency between the landing page and the other pages while navigating.
> 
> wt_submenu
> 
> Colours don't seem to be of the same naming.

I can update the sidebar to match the style of the navigation menu.

> * Nit pick: should all the rendering in the cards (small & large) be 1-1 relation?

I don't understand - can you clarify what you mean?

romfrancois:
> I don't understand - can you clarify what you mean?

I meant given the same screen size as shown on Figma, should the texts be aligned similarly with the same letter space and everything like shown on the screenshot (captures are not of the same size but you'll get the idea with the text).
comparison

PR #1006, opened 2022-02-11
The landing page improvements

timocov:


website/src/pages/index.tsx
```diff
@@ -30,7 +30,7 @@ const visibleLogicalRange = { from: 0.5, to: data.orangeData.length - 1.5 };
 
 function getLayoutOptionsForTheme(isDarkTheme: boolean): DeepPartial<LayoutOptions> {
 	return isDarkTheme
-		? { background: { color: '#0C0E15' }, textColor: 'rgba(248, 249, 253, 1)' }
+		? { background: { color: '#131722' }, textColor: 'rgba(248, 249, 253, 1)' }
```
romfrancois, line 33: Maybe we could replace with `--color-cold-gray-900`?

edew: I don't know if it will work but you can try `color: 'var(--color-cold-gray-900)'` seeing as the css variables will be defined by the time the JS runs?

timocov: The library doesn't support this kind of stuff and will fail to run I guess

edew: I just tried and setting `fillStyle` to a defined CSS variable didn't work, so it seems this isn't supported by the canvas context.

edew (approved):

PR #1010, opened 2022-02-14
Changed rendering of baseline and area series when using scaleMargins

### LINKED ISSUE #1005: Baseline series should use min/max price range in visible range to calculate gradient stops
3.7.0

**Steps/code to reproduce:**
- Add BaselineSeries
- Set a custom price scale
- Assign scaleMargins

```js
var series = chart.addBaselineSeries({
  priceScaleId: 'my-custom-scale',
  scaleMargins: {
    top: 0.8,
    bottom: 0,
  },
  bottomFillColor2: 'rgba(0,0,0,1)',
  bottomFillColor1: 'rgba(0,0,0,0)',
  topFillColor2: 'rgba(0,0,0,0)',
  topFillColor1: 'rgba(0,0,0,1)',
  baseValue: { type: 'price', price: 180 },
});
```

**Actual behavior:**
The baseline gradient does not fully render the assigned color stop from the configuration. 

**Expected behavior:**
The baseline gradient should use the scaleMargins to compute the color stops. 

```js
// src/renderers/baseline-renderer.ts:21
const gradient = ctx.createLinearGradient(0, 0, 0, data.bottom);
```

`data.bottom` should not be the height of the current chart, but the height of the series with the current scaleMargins. 

**Screenshots:**
Expected here would be a full-black (`#000000`) top and bottom gradient. 

Without scale margins:
image

Scale margins moved to top: 
image

Scale margins moved to bottom:
image

https://jsfiddle.net/0r3cLmjd/51/

EastingAndNorthing:
**Type of PR:** 
Enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #1005
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**
This change will make the rendering of color stops in the baseline and area series gradients more consistent when using scaleMargins. 

The gradient will now always reach the full color whether it is scaled to the top or bottom of the view (as I'd expect from setting its color options). 

The gradient is only changed when using a _custom_ price scale, since it seems that these cannot be moved manually. Is this correct? Otherwise, the gradient will go into negative values, giving unwanted results. 

Furthermore, I'm using the code below to check whether or not we are on a custom price scale: 
```js 
const isCustomScale = priceScale.id() !== 'right' && priceScale.id() !== 'left';
```
Not sure if this is the best method, open for suggestions. Maybe use the `DefaultPriceScaleId` enum?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@ debug.html
 
 # graphics tests out data
 /tests/e2e/graphics/.gendata/
+
+# MacOS
+.DS_Store
--- a/src/views/pane/baseline-pane-view.ts
+++ b/src/views/pane/baseline-pane-view.ts
@@ -32,15 +32,22 @@ export class SeriesBaselinePaneView extends LinePaneViewBase<'Baseline', LineIte
 		}
 
 		const baselineProps = this._series.options();
-		const priceScaleProps = this._series.priceScale().options();
+		const priceScale = this._series.priceScale();
+		const priceScaleProps = priceScale.options();
+		const isCustomScale = priceScale.id() !== 'right' && priceScale.id() !== 'left';
 
 		this._makeValid();
 
-		const baseLevelCoordinate = this._series.priceScale().priceToCoordinate(baselineProps.baseValue.price, firstValue.value);
+		const baseLevelCoordinate = priceScale.priceToCoordinate(baselineProps.baseValue.price, firstValue.value);
 		const barWidth = this._model.timeScale().barSpacing();
 
-		const bottom = priceScaleProps.scaleMargins ? (1 - priceScaleProps.scaleMargins.bottom) * height : height;
-		const top = priceScaleProps.scaleMargins ? priceScaleProps.scaleMargins.top * height : 0;
+		let top = 0;
+		let bottom = height;
+
+		if (baselineProps.baseValue.type === 'price' && isCustomScale && priceScaleProps.scaleMargins) {
+			bottom = height * (1 - priceScaleProps.scaleMargins.bottom);
+			top = height * priceScaleProps.scaleMargins.top;
+		}
 
 		this._baselineAreaRenderer.setData({
 			items: this._items,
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -32,3 +32,6 @@ debug.html
 
 # graphics tests out data
 /tests/e2e/graphics/.gendata/
+
+# MacOS
+.DS_Store
```
timocov, line 37: If you don't mind let's revert this. This folder should in your global gitignore since it has no in common with the project itself.

src/views/pane/baseline-pane-view.ts
```diff
@@ -32,12 +32,23 @@ export class SeriesBaselinePaneView extends LinePaneViewBase<'Baseline', LineIte
...
-		const baseLevelCoordinate = this._series.priceScale().priceToCoordinate(baselineProps.baseValue.price, firstValue.value);
+		const baseLevelCoordinate = priceScale.priceToCoordinate(baselineProps.baseValue.price, firstValue.value);
 		const barWidth = this._model.timeScale().barSpacing();
 
+		let top = 0;
+		let bottom = height;
+
+		if (baselineProps.baseValue.type === 'price' && isCustomScale && priceScaleProps.scaleMargins) {
+			bottom = height * (1 - priceScaleProps.scaleMargins.bottom);
+			top = height * priceScaleProps.scaleMargins.top;
```
timocov, line 49: well, I don't think that it should be like that. Instead, we need to get min/max coordinates for a visible range of this series. If we'll do this we won't need to check whether it is a "custom" scale or not because it will work automatically for any series. See my comment here https://github.com/tradingview/lightweight-charts/issues/1005#issuecomment-1038986474

timocov: we have `this._items` where we can get all coordinates for each item, but please make sure that you use `this._itemsVisibleRange` to iterate over that array properly.

timocov: also, you could take a look at `_convertToCoordinates` method, which is used to update coordinates, probably we could change override it here and calculate the data only once and save it somewhere.

timocov (changes requested):
See comments. Also, could you add tests for this feature please? Let me know if you need any help or additional information.

0220tiezo1323 (changes requested):
No

PR #1045, opened 2022-03-15
Docs replace images with charts

edew:
**Type of PR:** enhancement

**PR checklist:**

- [x] Documentation update

**Overview of change:**

Replaces the images of charts with real charts!

This PR adds a plugin to Docusaurus that will render a chart within a iframe using the content of a code block.

Basic support for being "version aware" is included. The plugin component will dynamically import the correct version of `lightweight-charts` based on the currently selected version of the docs.

Currently it does require changes to the plugin code block component every time we want to support a new version. It should be possible to remove this limitation if we can configure Webpack to correctly resolve dynamic imports for paths like `lightweight-charts-${version}`.


romfrancois:
Looks like the first rendering on iOS devices (I've checked on both iPad & iPhone) isn't displayed.
When changing page & coming back it's there.

IMG_0432

timocov (approved):

BUILDING.md
```diff
@@ -34,6 +34,7 @@ The minimal supported version of [NodeJS](https://nodejs.org/) for development i
 1. Run `npm run docusaurus docs:version MAJ:MIN` in `website` folder to create new versioned docs.
   Note that there is not patch version in docs, only major and minor parts.
 1. (optional) Remove docs for the oldest version (see <https://docusaurus.io/docs/versioning#deleting-an-existing-version>).
+1. Handle the new version in `import-lightweight-charts-version.ts`: add a package reference for that version should to `website/package.json` (e.g. `"lightweight-charts-MAJ.MIN": "npm:lightweight-charts@~MAJ.MIN.0"`) and a import of that package in a matching case statement.
```
romfrancois, line 37: I believe `should` should be removed.

romfrancois (approved):

PR #1052, opened 2022-04-04
feature/Allow set crosshair programmatically

matiasmolleja:
PR to allow set crosshair programmatically.
Read the README file for more info.


hskang9:
Can someone please add this feature? I am using version 4.0.0 and it sill shows  `setCrossHairXY is not a function`.

Marbemoreno:
Do you have something to solve this problem? or it is posible to restored the feature/set-crosshair and merge it?

Issue #1142, opened 2022-08-08
Unable to run `npm install` with Node v18+


SlicedSilver:
**Note:** This affects local development, not the published Lightweight charts package.

It is not possible to `npm install` if you are running node v18, unless you set `engine-strict=false` within the `.npmrc` file. 

## Issue
The outdated version of  `eslint-plugin-jsdoc` is causing the issue because one of it’s dependencies is requiring the following versions of node: `Required: {“node":"^12 || ^14 || ^16 || ^17"}`.

Our documentation is currently listing our minimal supported version as v16.13 thus we should be supporting the use of Node v18.
https://github.com/tradingview/lightweight-charts/blob/4daa902167fcac8341d0bb4f45a5648f56eec2af/BUILDING.md?plain=1#L3
https://github.com/tradingview/lightweight-charts/blob/4daa902167fcac8341d0bb4f45a5648f56eec2af/package.json#L32

PR #1143, opened 2022-08-09
Chore: Updating devDependencies

SlicedSilver:
**Type of PR:** enhancement

* Fixes: #1142 (ability to build project with node v18)
* Updated typescript to match the current Charting Library version of `4.7.3`
* Chore: updated the majority of the remaining `devDependencies` to the latest versions.

### Can’t be updated:
* `eslint`and `eslint-plugin-unicorn` need to stay at the current versions because updating to v8 of eslint  results in `jsdoc/check-examples` rule no longer running due to this issue: https://github.com/eslint/eslint/issues/14745
* `eslint-plugin-mdx` can’t be updated to v2 or later because it results in the linting task running indefinitely. The project makes heavy use of @example jsdocs thus it is important that these checks remain.

### Checks completed
- [x] `npm run verify` passes
- [x] Graphics tests pass
- [x] Checked built files
	* Development builds remain identical except for the typescript inserted comments.
	* Production builds remain identical except for the generated build numbers (which should always be different).
	* `typings.d.ts` files remain identical.


romfrancois:
Dont see any reasons not to merge this PR.
@eugene-korobko, @kirchet any objections?

package.json
```diff
@@ -35,55 +35,55 @@
...
-    "@typescript-eslint/eslint-plugin": "~4.32.0",
-    "@typescript-eslint/eslint-plugin-tslint": "~4.32.0",
-    "@typescript-eslint/parser": "~4.32.0",
-    "bytes": "~3.1.1",
+    "@rollup/plugin-node-resolve": "~13.3.0",
+    "@rollup/plugin-replace": "~4.0.0",
+    "@size-limit/file": "~8.0.0",
+    "@types/chai": "~4.3.3",
+    "@types/mocha": "~9.1.1",
+    "@types/node": "~18.6.5",
```
kirchet, line 43: I think we should sync `@types/node` version and docker images version here:

https://github.com/tradingview/lightweight-charts/blob/4daa902167fcac8341d0bb4f45a5648f56eec2af/.circleci/config.yml#L43-L49

SlicedSilver: Agreed, the `@types/node` dependency should be kept in sync with 
* the circleci docker image, and 
* the `engines` property within the `package.json`.

I'll fix that.

SlicedSilver:
FYI: The existing CircleCI docker image `circleci/node:16` was deprecated earlier this year (Feb) hence the reason Node was being loaded as `16.13`. I've updated the config file to rather use the new `cimg/node` set of images. This allows the use of the latest LTS version of `16.16.0`.

For reference:

- https://circleci.com/docs/next-gen-migration-guide
- https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic/

kirchet (approved):

romfrancois (approved):

Issue #1135, opened 2022-08-02
Documentation: How To (Snippets / Examples)


SlicedSilver:
The lightweight charts website ([Lightweight Financial Charting Library — TradingView](https://www.tradingview.com/lightweight-charts/)) has a great selection of examples showcasing how to implement various features, such as: tooltips, range selectors, legends.

A section could be added to the documentation site containing short explanations of how to implement these features. This makes the examples: 
1. easier to find for more users (who may not visit the website but go straight to the documentation site), and 
2. easier to find via a search engine.

These pages can be short and sweet depending on each example. For example,  `How to invert a price scale` would be very short because it just needs to reference the `invertScale` property on the priceScale and show a small code sample. Where as, `How to add a legend` would require a bit more explanation.

**Each page would include:**
* The example embedded with a JSFiddle link to the full source code.
* Code blocks containing the key parts of the example.
* Links to APIs and Options relevant to the current example.

**A tentative list of pages to be added:**
How to…
- [ ]  ...add a custom legend,
- [ ]  ...adjust the range (implement a range selector),
- [ ]  ...add a tooltip,
- [ ]  ...add a watermark,
- [ ]  ...invert a price scale,
- [ ]  ...add a price line (with title),
- [ ]  ...add series markers,
- [ ]  ...show two price scales,
- [ ]  ...show price and volume on a single chart.

Issue #1134, opened 2022-08-02
Documentation: Guided Tutorial - Customisation


SlicedSilver:
**Purpose:** Provide a simple walkthrough which showcases how to customise Lightweight Charts to the user’s requirements.  

Essentially, a step by step guide which explains to the reader where and how to apply the customisations options. The user should follow along on their IDE and end up with the same finished output (building up as they go along). (Each section would carry on from the previous section and progress towards the final result.)

_Note:_ The reader would be expect to first read the `Getting Started` section of the documentation site. Thus it wouldn't need to cover / repeat everything included there.

The walkthrough will start with a standard line chart and describe how to customise the following (simple changes so the tutorial is quick and easy to follow):
- [x]  Colours
- [x]  Price Scale
- [x]  Time Scale
- [x]  Price Line

Page contents:
* Each page will have code blocks, and an embedded chart of the result.
* Each section will have a link to a complete run-able code example (on JSFiddle or something similar).
* Links to the referenced APIs

PR #1144, opened 2022-08-09
Updating Docusaurus version for documentation website

### LINKED ISSUE #1141: Update Docusaurus dependency for documentation website
The documentation website is currently using a beta version of Docusaurus (v2). V2 of Docusaurus has now exited beta and been released. https://docusaurus.io/blog/2022/08/01/announcing-docusaurus-2.0

There is a new feature (https://docusaurus.io/docs/markdown-features/code-blocks#custom-magic-comments) which will be useful for #1134 so it would be a good time to update the dependency to the latest release.

SlicedSilver:
**Type of PR:** enhancement

**Updating Docusaurus version for the documentation website to the latest version.**

The documentation website was using a beta version of Docusaurus (v2.0.0-beta.17) and this pull request brings it up to the release version of 2.0.1.

Updating Docusaurus unlocks a few new features which will be useful for #1134 and #1135

Closes #1141


romfrancois:
Dont see any objections in approving & merging this one.
@eugene-korobko, @kirchet any comments?

kirchet:
> Dont see any objections in approving & merging this one. @eugene-korobko, @kirchet any comments?

Looks good for me, but we need to fix the pipeline 🙂

SlicedSilver:
**To fix the pipeline:**
1. Make the `build` job require the completion of the `install-deps-website` job (in addition to `install-deps`).
2. Bump required node version from `16.3.1` to `16.14`.
---

**Item 1)** The `tsc-verify` task is expecting the docusaurus tsconfig file to be installed already. Thus we need to only run the `build` job after both the `install-deps` and `install-reps-website` jobs have run.
https://github.com/tradingview/lightweight-charts/blob/4daa902167fcac8341d0bb4f45a5648f56eec2af/.circleci/config.yml#L293-L296

**Item 2)** Docusaurus build complains if the node version is less than `16.14`. Can we bump the Node version up to `16.14`?
FYI: Release notes for Node v16.14: https://nodejs.org/en/blog/release/v16.14.0/

Would need to change the following:
https://github.com/tradingview/lightweight-charts/blob/4daa902167fcac8341d0bb4f45a5648f56eec2af/.circleci/config.yml#L43-L45
to:
`- image: circleci/node:16.14` 

which would install node 16.14.2 as mentioned here: https://circleci.com/developer/images/image/cimg/node

Additionally we would need to update the version numbers mentioned in `package.json`, `building.md`

@kirchet Do these changes look correct?

kirchet:
> Do these changes look correct?

Yes. But probably we need to merge #1143 first and then bump version to `18.6.5`. Or even bump this version in #1143.

SlicedSilver:
> Yes. But probably we need to merge #1143 first and then bump version to `18.6.5`. Or even bump this version in #1143.

Agreed, that it would be better to first merge #1144. 

Are we happy to jump to `18.6.5` rather than updating to the minimum required version of `16.14`? If so, the latest 'current' node version is `18.7.0`. Should we use `18.7.0`?

If we are happy to proceed to update Node, then I'll make the changes to the #1143 pull request. Including: updating circleCI version, `@types/node`, `package.json`, and `building.md`.

kirchet:
> Are we happy to jump to 18.6.5 rather than updating to the minimum required version of 16.14?

Yes

> If so, the latest 'current' node version is 18.7.0. Should we use 18.7.0?

I don't think that there are a lot of differences, but it's better to keep in sync the version of Node everywhere where it possible. So, if we're using `@types/node": "~18.6.5` we should use `18.6.5` everywhere.

SlicedSilver:
@romfrancois @kirchet
Tested this PR again with Node 16.16 and the recent merge. Should be good to go.

The `dts-changes` CI test fails but is expected because there was a typo fix for the jsdoc in this PR. 
Diff for your reference: https://output.circle-artifacts.com/output/job/42bd57fb-27e0-49cf-b8fe-f827f43995ab/artifacts/0/dts-changes/changes.diff

romfrancois (approved):

Issue #1002, opened 2022-02-10
Documentation website improves


timocov:
- [x] ~on website deploy, instead of reading size-limit config to get a size it's better to use [bundlephobia api](https://bundlephobia.com/api/size?package=lightweight-charts@3.7.0&record=true) because ppl will use this service to check the size~ Fixed in #1009
- [ ] when a screen is small enough we hide a text from the logo, but it seems that in some cases we could display it for a bit longer
  ![image](https://user-images.githubusercontent.com/3112183/153471099-47dea8f6-999f-4baa-a8a7-f6f20c06ef63.png)
- [ ] when a screen is small the header might look strange sometimes:
  ![image](https://user-images.githubusercontent.com/3112183/154325707-83dba49c-9019-40a2-b03f-4cdeccfaa085.png)
- [ ] #1001 
- [ ] migrate to postcss from css
- [ ] introduce stylelint
- [ ] try to use client-side js-based adaptive logo rather than have 4 icons and load them as images
- tbc

igoralves1:
The documentation needs improvement. It is not clear how to customize charts.

Issue #774, opened 2021-06-21
Watermark icon


djaeger-inercomp:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
It would be nice if it was possible to add an icon next to the watermark text (left or right). It would be generally nice to be able to add an image as watermark instead of just text.

**Additional context**

thanhlmm:
Yeah, this should be great to have it because we need to add logo as trademark

dubeychandan996:
Is there any alternate method to implement to use the image as background color?
Putting image via CSS I (in canvas) not working because it apply on both canvas and has no power (position: back/front) for z-index.

SlicedSilver:
Example for how to add a logo as a watermark: https://tradingview.github.io/lightweight-charts/tutorials/how_to/watermark

PR #1153, opened 2022-08-18
Docs: Customisation Guide and ‘How To’ Examples

### LINKED ISSUE #1134: Documentation: Guided Tutorial - Customisation

### LINKED ISSUE #1135: Documentation: How To (Snippets / Examples)

SlicedSilver:
**Type of PR:** enhancement

**Overview of changes:**
* Add customisation guide to tutorials section.
* Add ‘how to’ examples to the tutorials section.
* Improved code block component to support: 
	* hiding and revealing lines, 
	* fading specific lines, and 
	* removing lines (but keeping in the copy clipboard).
* Added a warning banner for the tutorials page which only appears when user isn’t browsing the latest published version (since the tutorials section isn’t versioned yet).

**Preview Link**
https://gorgeous-florentine-dd67a1.netlify.app/lightweight-charts/tutorials

**Issues addressed:**
* Closes #1135
* Closes #1134

**Related issues:**
* We could potentially close this issue: #1002
* Watermark tutorial address this question: #774


romfrancois:
LGTM ... didn't spot anything obvious so far.
All the example tested make sense and work as expected.

romfrancois (approved):

edew:
Looks good - a couple of questions though @SlicedSilver:

1. When I run the site locally the examples for the tutorial don't load. Is there an extra step I need to run to get them working?

image

2. The expand/collapse animation is quite slow, at least on my machine. What do you think about removing the animation?

expandcollapse

SlicedSilver:
> Looks good - a couple of questions though @SlicedSilver:
> 
> 1. When I run the site locally the examples for the tutorial don't load. Is there an extra step I need to run to get them working?
> 
> 2. The expand/collapse animation is quite slow, at least on my machine. What do you think about removing the animation?
> 

1. There are no extra steps required to build the site. It's an issue which only seems to occur when using `docusaurus serve`. The files are correctly generated, and the site works when hosted on a site or using a local web server like `python -m http.server`. I suspect that the docusaurus implementation of a server is doing something extra that's not on a typical server. Also it works in normal development mode (`npm run start`). I'll do a bit more digging into the docusaurus code to find the reason.

2. The animation time is very long because the revealed code block is very large (and the animation length is determined by the height). Currently using the built-in `details` component. I can swizzle the component to change the animation behaviour (or probably just apply a css rule).

SlicedSilver:
> When I run the site locally the examples for the tutorial don't load. Is there an extra step I need to run to get them working?
>> ...
>> I suspect that the Docusaurus implementation of a server is doing something extra that's not on a typical server. 
>> ...

Answer: The docusaurus serve command is setting this option: https://github.com/vercel/serve-handler#cleanurls-booleanarray

https://github.com/facebook/docusaurus/blob/94067ce9305653a4534f23c96406ae7fe9e35837/packages/docusaurus/src/commands/serve.ts#L77

Which automatically strips `.html` from the end of any url. GH pages doesn't have this behaviour, so it works as expected on GH pages amongst others. 

I'm going to add a feature request to Docusaurus to enable this as an options flag.

SlicedSilver:
Updated the PR to include an 'instant' details element which instantly reveals the complete code blocks instead of using the default animation.

Preview: https://gorgeous-florentine-dd67a1.netlify.app/lightweight-charts/tutorials/customization/chart-colors#complete-code

romfrancois:
> Updated the PR to include an 'instant' details element which instantly reveals the complete code blocks instead of using the default animation.
> 
> Preview: https://gorgeous-florentine-dd67a1.netlify.app/lightweight-charts/tutorials/customization/chart-colors#complete-code

It's way better this way!

edew (approved):

Issue #1154, opened 2022-08-19
Failing graphics tests on the CI pipeline


SlicedSilver:
The automated graphics tests on the CI pipeline are failing at the higher resolutions. 

The only difference between the test cases is the position of the crosshair on the charts. There is currently a programmed delay between moving the mouse and taking the screenshot. The delay appears to be too short when running the test on the CI server.
https://github.com/tradingview/lightweight-charts/blob/42ab0d95b404d70698d971df774d8a3aa9213c04/tests/e2e/graphics/helpers/screenshoter.ts#L74-L85

We can replicate the issue locally by throttling the CPU within puppeteer.
```js
// within generateScreenshot (screenshoter.ts L47)
await page.emulateCPUThrottling(8);
```
and setting the test retries to 1.

PR #1158, opened 2022-08-25
Improve tests

### LINKED ISSUE #1155: Unreliable memory leak test
The memory leak test is inconsistent between repeated executions. 

Running the test repeatedly will achieve different results, where sometimes it passes and sometimes it fails. 
It generally passes on the first try so it isn’t directly affecting the CI pipeline at the moment.

We need to improve the robustness of the test to account for the variable times until the garbage is collected by puppeteer instance.

SlicedSilver:
**Type of PR:** enhancement / fix

**PR checklist:**

- [x] Addresses an existing issue: (see below)
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**

This PR includes a few changes to improve the current e2e tests of the library.

**Issues:**
- Fixes #1155
- Improves #1154 - This is a temporary fix to improve the situation but the test still remains flaky.

**Changes:**
* Added further documentation regarding the tests. 
* Improved the reliability of the memory leaks test. Instead of using a fixed delay, we now rather poll the references count until the expected change has occurred. Additionally, the test now ensures references are created in the first place.
* Coverage test can now produce a ‘covered’ file if evoked with the `GENERATE_COVERAGE_FILE` environment variable set to true. This covered file can be diffed against the input file to determine which parts of the library aren’t being covered in the test.
* Increased the delay before taking a screenshot within the graphics e2e tests. This is a temporary fix as the test is still flaky.
* Improved the score of the coverage test by 5.5%. Test now includes touch gestures: swipe, pinch, zoom, tap, and long tap.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ debug.html
 
 # graphics tests out data
 /tests/e2e/graphics/.gendata/
+tests/e2e/coverage/.gendata
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,6 +3,7 @@
         "dist/**": true,
         "lib/**": true,
         "tests/e2e/graphics/.gendata/**": true,
+        "tests/e2e/coverage/.gendata/**": true,
         "website/.docusaurus/**": true,
         "website/build/**": true,
         "website/docs/api/**": true,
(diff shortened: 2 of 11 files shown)


romfrancois (dismissed):
LGTM

SlicedSilver:
https://github.com/tradingview/lightweight-charts/pull/1158/commits/ac4bd5b82a63c409fce034b2a38ba6a988e75441
Thought of one more thing to add to the coverage test which increases the score by another 0.2%. Brings the score up to 91.4%

kirchet (approved):

PR #1171, opened 2022-09-23
fix type for createPriceLine options parameter

### LINKED ISSUE #1091: More accurate TS types for `createPriceLine` within `ISeriesApi`
**Lightweight Charts Version:** 3.8.0

**Actual behavior:**
When creating a new price line using the [createPriceLine](https://github.com/tradingview/lightweight-charts/blob/e45ed5b89cb9212641a7806b721f5b614d912ee6/src/api/series-api.ts#L173) method on the `ISeriesApi`, the typescript compiler expects the options parameter to be of type `PriceLineOptions` and that ALL the properties are required.

An error message such as this appears:
`Argument of type '{ price: number; }' is not assignable to parameter of type 'PriceLineOptions'.`

However, the actual implementation only requires that the `price` property be set and thus the other properties should be optional. 

https://github.com/tradingview/lightweight-charts/blob/e45ed5b89cb9212641a7806b721f5b614d912ee6/src/api/series-api.ts#L173-L179

* The `checkPriceLineOptions` function ensures that the `price` property is present and is a number. 
* Next, the user supplied options are merged with the default options
* Resulting in a fully formed options object (type `PriceLineOptions`) even if the other properties aren't included

**Expected behavior:**
A typescript error shouldn't occur when leaving out any of the properties (except `price` of course).

**Code to reproduce:**

The following shouldn't trigger a typescript error.
```ts
  // @ts-expect-error 
  const newPriceLine = line.createPriceLine({
    price: 1234,
  });
  // ^ where line is of type: ISeriesApi<"Line">
```

**Proposed Solution:**
Update the typings for the `createPriceLine` options object to something like this:
```ts
// One-liner
type CreatePriceLineOptions = Partial<PriceLineOptions> & Pick<PriceLineOptions, 'price'>;

// Or using a new utility type if required in multiple places
type PartialExcept<T, K extends keyof T> = Partial<T> & Pick<T, K>;
type CreatePriceLineOptions =  PartialExcept<PriceLineOptions, 'price'>;
```

SlicedSilver:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #1091
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**

Some change to fix the type for the `createPriceLine` options parameter which should be a `Partial`

### DIFF
--- a/src/api/data-validators.ts
+++ b/src/api/data-validators.ts
@@ -8,12 +8,11 @@ import { Time } from '../model/time-data';
 import { isFulfilledData, SeriesDataItemTypeMap } from './data-consumer';
 import { convertTime } from './data-layer';
 
-export function checkPriceLineOptions(options: PriceLineOptions): void {
+export function checkPriceLineOptions(options: Partial<PriceLineOptions>): void {
 	if (process.env.NODE_ENV === 'production') {
 		return;
 	}
 
-	// eslint-disable-next-line @typescript-eslint/tslint/config
 	assert(typeof options.price === 'number', `the type of 'price' price line's property must be a number, got '${typeof options.price}'`);
 }
 
--- a/src/api/data-validators.ts
+++ b/src/api/data-validators.ts
@@ -1,6 +1,6 @@
 import { assert } from '../helpers/assertions';
 
-import { PriceLineOptions } from '../model/price-line-options';
+import { CreatePriceLineOptions } from '../model/price-line-options';
 import { SeriesMarker } from '../model/series-markers';
 import { SeriesType } from '../model/series-options';
 import { Time } from '../model/time-data';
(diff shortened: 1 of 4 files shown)


romfrancois (approved):

src/api/data-validators.ts
```diff
@@ -8,12 +8,11 @@ import { Time } from '../model/time-data';
...
 import { convertTime } from './data-layer';
 
-export function checkPriceLineOptions(options: PriceLineOptions): void {
+export function checkPriceLineOptions(options: Partial<PriceLineOptions>): void {
 	if (process.env.NODE_ENV === 'production') {
 		return;
 	}
 
-	// eslint-disable-next-line @typescript-eslint/tslint/config
 	assert(typeof options.price === 'number', `the type of 'price' price line's property must be a number, got '${typeof options.price}'`);
```
kirchet, line 16: Accordingly to this assert the type of options should be `Partial<PriceLineOptions> & Pick<PriceLineOptions, 'price'>`

kirchet (changes requested):

src/api/data-validators.ts
SlicedSilver: Thanks @kirchet. I've updated the typing to ensure that `price` is always provided.

kirchet (approved):

PR #1174, opened 2022-10-03
Improve reliability of graphics tests

### LINKED ISSUE #1154: Failing graphics tests on the CI pipeline

SlicedSilver:

The graphics tests are flaky when run on the CI server. This PR improves the reliability of the tests such false negatives should be greatly reduced.

Fixes #1154 

**Overview of changes:**
* When a test fails and is retried, it will then run in series instead of parallel.
* A `subscribeCrosshairMove` event handler is added to the chart such that the test case waits until the chart has reacted to the mouse movement before taking the screenshot
	* Test cases have been updated to have a `getChartInstance` function which is used to add the above crosshairMove handler. 
	* Setting `window.IGNORE_MOUSE_MOVE = true;` within the test case will disable the waiting for cursor (in cases where the chart is small, there are multiple charts, or the chart isn’t able to have mouse events (`take-screenshot.js`)
* Added the ability to pass Mocha options via terminal commands. (See `tests/e2e/graphics/README.md`). Useful for selecting subsets of tests to run using `--grep`.

### DIFF
--- a/tests/e2e/graphics/graphics-test-cases.ts
+++ b/tests/e2e/graphics/graphics-test-cases.ts
@@ -114,8 +114,16 @@ describe(`Graphics tests with devicePixelRatio=${devicePixelRatioStr} (${buildMo
 });
 
 function registerTestCases(testCases: TestCase[], screenshoter: Screenshoter, outDir: string): void {
+	const attempts: Record<string, number> = {};
+	testCases.forEach((testCase: TestCase) => {
+		attempts[testCase.name] = 0;
+	});
+
 	for (const testCase of testCases) {
 		it(testCase.name, async () => {
+			const previousAttempts = attempts[testCase.name];
+			attempts[testCase.name] += 1;
+
 			const testCaseOutDir = path.join(outDir, testCase.name);
 			rmRf(testCaseOutDir);
 			fs.mkdirSync(testCaseOutDir, { recursive: true });
@@ -132,6 +140,12 @@ function registerTestCases(testCases: TestCase[], screenshoter: Screenshoter, ou
 
 			// run in parallel to increase speed
 			const goldenScreenshotPromise = screenshoter.generateScreenshot(goldenPageContent);
+
+			if (previousAttempts) {
+				// If a test has previously failed then attempt to run the tests in series (one at a time).
+				await goldenScreenshotPromise;
+			}
+
 			const testScreenshotPromise = screenshoter.generateScreenshot(testPageContent);
 
 			const errors: string[] = [];
--- a/tests/e2e/graphics/helpers/screenshoter.ts
+++ b/tests/e2e/graphics/helpers/screenshoter.ts
@@ -71,8 +71,27 @@ export class Screenshoter {
 				return (window as any).testCaseReady;
 			});
 
+			// move mouse to top-left corner
+			await page.mouse.move(0, 0);
+
+			const waitForMouseMove = page.evaluate(() => {
+				/* eslint-disable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-explicit-any */
+				if ((window as any).IGNORE_MOUSE_MOVE) { return Promise.resolve(); }
+				return new Promise<number[]>((resolve: (value: number[]) => void) => {
+					const chart = (window as any).getChartInstance();
+					chart.subscribeCrosshairMove((param: any) => {
+						if (param.point.x > 0 && param.point.y > 0) {
+							requestAnimationFrame(() => resolve([param.point.x, param.point.y] as number[]));
+						}
+					});
+				});
+				/* eslint-enable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-explicit-any */
+			});
+
 			// to avoid random cursor position
 			await page.mouse.move(viewportWidth / 2, viewportHeight / 2);
+
+			await waitForMouseMove;
 
 			// let's wait until the next af to make sure that everything is repainted
 			await page.evaluate(() => {
--- a/package.json
+++ b/package.json
@@ -58,9 +58,9 @@
     "eslint-plugin-markdown": "~3.0.0",
     "eslint-plugin-mdx": "~1.17.0",
     "eslint-plugin-prefer-arrow": "~1.2.3",
+    "eslint-plugin-react": "~7.30.1",
     "eslint-plugin-tsdoc": "~0.2.16",
     "eslint-plugin-unicorn": "~40.1.0",
-    "eslint-plugin-react": "~7.30.1",
     "express": "~4.18.1",
     "glob": "~8.0.3",
     "markdown-it": "~13.0.1",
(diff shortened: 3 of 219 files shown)


tests/e2e/graphics/graphics-test-cases.ts
```diff
@@ -132,6 +140,12 @@ function registerTestCases(testCases: TestCase[], screenshoter: Screenshoter, ou
 
 			// run in parallel to increase speed
 			const goldenScreenshotPromise = screenshoter.generateScreenshot(goldenPageContent);
+
+			if (previousAttempts) {
+				// If a test has previously failed then attempt to run the tests in series (one at a time).
+				await goldenScreenshotPromise;
+			}
```
romfrancois, line 147: Pretty good idea!

romfrancois (approved):
This PR makes a lot of sense. Hopefully this retry in series mechanism will help reduce false positive and tests will finally make sense and be useful !

tests/e2e/graphics/graphics-test-cases.ts
kirchet: Yes, but should we do it inside a `try` `catch` block as we do it below?

```TS
const errors: string[] = [];
const failedPages: string[] = [];

let goldenScreenshot: PNG | null = null;

// run in parallel to increase speed
const goldenScreenshotPromise = screenshoter.generateScreenshot(goldenPageContent);

if (previousAttempts > 0) {
  // If a test has previously failed then attempt to run the tests in series (one at a time).
  try {
    goldenScreenshot = await goldenScreenshotPromise;
    writeTestDataItem('1.golden.png', PNG.sync.write(goldenScreenshot));
  } catch (e: unknown) {
    errors.push(`=== Golden page ===\n${(e as Error).message}`);
    failedPages.push('golden');
  }
}

const testScreenshotPromise = screenshoter.generateScreenshot(testPageContent);
...
if (previousAttempts === 0) {
  try {
    goldenScreenshot = await goldenScreenshotPromise;
    writeTestDataItem('1.golden.png', PNG.sync.write(goldenScreenshot));
  } catch (e: unknown) {
    errors.push(`=== Golden page ===\n${(e as Error).message}`);
    failedPages.push('golden');
  }
}
```

tests/e2e/graphics/helpers/screenshoter.ts
```diff
@@ -71,9 +71,28 @@ export class Screenshoter {
...
 			});
 
+			// move mouse to top-left corner
+			await page.mouse.move(0, 0);
+
+			const waitForMouseMove = page.evaluate(() => {
+				/* eslint-disable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-explicit-any */
+				if ((window as any).IGNORE_MOUSE_MOVE) { return Promise.resolve(); }
+				return new Promise<number[]>((resolve: (value: number[]) => void) => {
+					const chart = (window as any).getChartInstance();
```
kirchet, line 81: Why in one case we use the getter function `getChartInstance()` but in another case the global variable `IGNORE_MOUSE_MOVE`? Maybe it's better to add one more global variable like `window.chart`/`window.CHART` instead of adding a new function?

One more option is to change the signature of the `testCaseReady`
```ts
interface TestCaseResult {
  chart: IChartApi;
  ignoreMouseMove?: boolean;
}
testCaseReady(): TestCaseResult  | Promise<TestCaseResult>
```

But I'm not sure that we can get it from the browser context, store it, and then somehow provide it from the node context back to the browser context.

tests/e2e/graphics/helpers/screenshoter.ts
```diff
@@ -71,9 +71,28 @@ export class Screenshoter {
 				return (window as any).testCaseReady;
 			});
 
+			// move mouse to top-left corner
+			await page.mouse.move(0, 0);
+
+			const waitForMouseMove = page.evaluate(() => {
+				/* eslint-disable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-explicit-any */
+				if ((window as any).IGNORE_MOUSE_MOVE) { return Promise.resolve(); }
```
kirchet, line 79: We can define an interface for the tests and use it in all places where we use now `(window as any)`

```ts
export interface LibraryWindow extends Window {
  testCaseReady: Promise<void>;
  ....
}

if ((window as unknown as LibraryWindow).IGNORE_MOUSE_MOVE) {...
```

tests/e2e/graphics/helpers/screenshoter.ts
kirchet: However `testCaseReady` is a global variable

https://github.com/tradingview/lightweight-charts/blob/a7a712f04a82435e874a47728033d7fa6142e72b/tests/e2e/graphics/helpers/test-page-dummy.html#L20

So we can just change the `runTestCase` signature:

```ts
testCaseReady(): TestCaseResult  | Promise<TestCaseResult>;
```

tests/e2e/graphics/graphics-test-cases.ts
SlicedSilver: Thanks. I've now wrapped the await within a try/catch block. This ensures that the error message is outputted to the txt files with the test artefacts.

tests/e2e/graphics/helpers/screenshoter.ts
SlicedSilver: Changed the tests such that a variable on the global window is created and we can remove the `getChartInstance` function.

tests/e2e/graphics/helpers/screenshoter.ts
SlicedSilver: I've added the types (within `test case-window-type.ts` file). This required changing the `tsconfig.json` for the e2e tests folder.

tests/e2e/graphics/graphics-test-cases.ts
kirchet: We also need to check for `if (previousAttempts === 0) {` below, otherwise

```ts
errors.push(`=== Golden page ===\n${(e as Error).message}`);
failedPages.push('golden');
```

can be called twice

SlicedSilver: I can change the first try/catch (within `previousAttempts > 0` branch) to have an empty catch statement. The second try/catch will catch this error again (promise is in rejected state) and handle the error reporting (to failedPages, and errors arrays). This saves repeating some of the code.

```typescript
if (previousAttempts) {
  try {
    // If a test has previously failed then attempt to run the tests in series (one at a time).
    await goldenScreenshotPromise;
  } catch {
    // Will be caught again later...
  }
}
```

Is that alright?

kirchet: Yes, seems good

kirchet (approved):

PR #1187, opened 2022-10-26
fix scroll speed on Windows chromium browsers

### LINKED ISSUE #1168: Inconsistent Mouse Wheel Scroll Speed [Windows Chrome]
**Lightweight Charts Version:** 3.8.0

On windows chrome the pixel wheel delta is not adjusted to DPR (Device Pixel Ratio) of the screen, scroll might be too slow or too fast depending on DPR on windows.

Chromium bugs:
- https://bugs.chromium.org/p/chromium/issues/detail?id=1001735&q=wheel%20delta%20windows&can=2 
- https://bugs.chromium.org/p/chromium/issues/detail?id=1207308&q=wheel%20delta%20windows&can=2 

**Steps/code to reproduce:**

 On a windows computer (with a high DPR screen), compare the scroll speed between lightweight charts on Chrome and Firefox.

**Actual behavior:**

On Windows laptop with DPR 3.5, horizontal scrolling through the touchpad in chrome is much faster than in Firefox. When I set DPR to 1 the scroll speed becomes the same in all browsers.

**Expected behavior:**

Scroll speed should be the same in all browsers.

SlicedSilver:
**Type of PR:** bugfix (browser bug, not our bug)

**PR checklist:**

- [x] Addresses an existing issue: fixes #1168
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**

- Library will check if the browser is Chromium based and whether it is running on windows.
- If it is, then the scroll wheel speed will be scaled by the `devicePixelRatio` to ensure that the scroll speed is consistent between all the different browsers.

https://bugs.chromium.org/p/chromium/issues/detail?id=1001735 
https://bugs.chromium.org/p/chromium/issues/detail?id=1207308 

The browser check is using `navigator.userAgentData` which is a new API only supported by Chromium browsers (since mid 2021). This is a more reliable way of checking the user agent, and since we are only looking for Chromium browsers the browser support isn't important and actually improves the check.

https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData

### DIFF
--- a/src/gui/chart-widget.ts
+++ b/src/gui/chart-widget.ts
@@ -1,4 +1,5 @@
 import { ensureDefined, ensureNotNull } from '../helpers/assertions';
+import { userAgentWindowsChromium } from '../helpers/browsers';
 import { drawScaled } from '../helpers/canvas-helpers';
 import { Delegate } from '../helpers/delegate';
 import { IDestroyable } from '../helpers/idestroyable';
@@ -53,9 +54,12 @@ export class ChartWidget implements IDestroyable {
 	private _clicked: Delegate<MouseEventParamsImplSupplier> = new Delegate();
 	private _crosshairMoved: Delegate<MouseEventParamsImplSupplier> = new Delegate();
 	private _onWheelBound: (event: WheelEvent) => void;
+	private _windowsChrome: boolean;
 
 	public constructor(container: HTMLElement, options: ChartOptionsInternal) {
 		this._options = options;
+
+		this._windowsChrome = userAgentWindowsChromium();
 
 		this._element = document.createElement('div');
 		this._element.classList.add('tv-lightweight-charts');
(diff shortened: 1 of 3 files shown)


edew (commented):
Looks good to me

src/gui/chart-widget.ts
```diff
@@ -53,10 +54,13 @@ export class ChartWidget implements IDestroyable {
 	private _clicked: Delegate<MouseEventParamsImplSupplier> = new Delegate();
 	private _crosshairMoved: Delegate<MouseEventParamsImplSupplier> = new Delegate();
 	private _onWheelBound: (event: WheelEvent) => void;
+	private _windowsChrome: boolean;
 
 	public constructor(container: HTMLElement, options: ChartOptionsInternal) {
 		this._options = options;
 
+		this._windowsChrome = userAgentWindowsChromium();
```
kirchet, line 62: I think we can move this variable on the module level

src/helpers/browsers.ts
```diff
@@ -22,3 +22,17 @@ export function isChrome(): boolean {
...
+		navigator.userAgentData.platform !== 'Windows'
+	) {
+		return false;
+	}
+	return navigator.userAgentData.brands.some(
+		(brand: UADataBrand) => {
+			return brand.brand.includes('Chromium');
+		}
+	);
+}
```
kirchet, line 38: I suggest to split this function into two

* `isChromiumBased`
* `isWindows`

kirchet (changes requested):

SlicedSilver:
Merge conflict with `.size-limit.js`. Merging master into this branch and running the size check produces:
```
ESM
  Size limit: 44.2 kB
  Size:       43.74 kB gzipped

Standalone
  Package size limit has exceeded by 14 B
  Size limit: 44.3 kB
  Size:       44.31 kB gzipped
```

Therefore to fix the conflict would require values of at least `43.8` and `44.4`.

kirchet (dismissed):

kirchet (approved):

PR #1064, opened 2022-04-17
Implement basic box drawing

tpunt:
**Type of PR:** Enhancement

**PR checklist:**

- [ ] Addresses an existing issue: fixes #
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**

I have implemented basic box/rectangle drawing. This feature has been requested a few times in the past (#706, #408, in online chatrooms, on discord, etc). I realise it has been discussed to expose the raw canvas also as a potential solution, but it seems there is some complexity around this (e.g. there are multiple canvases). It also feels as though basic features such as this (that complement the line drawing abilities) would be much better suited to be integrated into the library itself (which would also benefit from having the objects rescaled upon the rescaling of axes).

Things done:
 - Box creation based upon two prices and two times
 - Box moving with the rescaling of the axes
 - Box border (with colour, styling, thickness, and the ability to disable it)
 - Box fill colour with configurable opacity

Things remaining to do:
 - Ability to add a title to the box
 - Add prices to the Y axis for the top/bottom of the box
 - Add tests
 - Update the documentation

If this is desirable to have in the main project, then I am happy to spend some more time on finishing things up. If this is not desirable though, then I'll simply use my fork, and this PR can be closed.

Code example:
```js
const box = {
  lowPrice: 2713.0,
  highPrice: 2797.0,
  earlyTime: 1641409200,
  lateTime: 1641823200,
  borderColor: '#00008B',
  borderWidth: 3,
  borderStyle: LightweightCharts.LineStyle.Dashed,
  fillColor: '#0ff',
  fillOpacity: 0.2,
  borderVisible: true,
  axisLabelVisible: false,
  title: 'My box',
};

const createdBox = mainSeries.createBox(box);

// mainSeries.removeBox(createdBox);
```

<details>
  <summary>Example pictures</summary>
  Screenshot 2022-04-17 at 16 52 55
  
  ezgif-2-4852b5dc20
</details>

**Is there anything you'd like reviewers to focus on?**

Whether this feature is desirable to have in the core library.

# PR Update:
See https://github.com/tradingview/lightweight-charts/pull/1064#issuecomment-1304812364 of this PR for more abilities.

Issue #1049, opened 2022-03-23
add Custom vertical Line not by LineSeries


zeeshan-shabbir:
is it possible we can show vertical line on time with no value required just line from up way to down in any graph (chandle bar area)graph? like vertical line on crosshair .
if not how to show vertical line?

zeeshan-shabbir:
waiting for response :-)

SlicedSilver:
Duplicate of #128, #127, and #965. Please post your suggestion there.

Issue #698, opened 2021-02-10
Vertical lines on candlestick chart vs line chart - it's not straight on candlestick chart


pavitrakumar78:
**Lightweight Charts Version:** 3.2.0

**Steps/code to reproduce:**

Candlestick chart: https://jsfiddle.net/pavitrakumar/qr2htzv9/
Line chart: https://jsfiddle.net/pavitrakumar/36kowhuq/
All I've done is, add
```
	var series2 = chart.addLineSeries({
		color: 'rgb(45, 120, 255)',
		lineWidth: 2,
	});
	var data2 = [
          {time: { year: 2018 ,month: 4 ,day: 18 }, value: 999},
          {time: { year: 2018 ,month: 4 ,day: 18 }, value: -999}
    ]
  series2.setData(data2);
```
to the example fiddles for candlestick and line chart.

please zoon on the y-axis.

**Actual behavior:**

The vertical line on candlestick is not straight - it's diagonal.

**Expected behavior:**

The vertical line on candlestick chart needs to be straight.

**Screenshots:**

![image](https://user-images.githubusercontent.com/5354407/107568357-e017de00-6c0c-11eb-84ed-2300f0181148.png)

![image](https://user-images.githubusercontent.com/5354407/107568385-e908af80-6c0c-11eb-8949-d0bcd8b7cf3e.png)

**CodeSandbox/JSFiddle/etc link:**

Candlestick chart: https://jsfiddle.net/pavitrakumar/qr2htzv9/
Line chart: https://jsfiddle.net/pavitrakumar/36kowhuq/

timocov:
> All I've done is, add

Is that intended that for candlestick charts you have 2019 instead of 2018 for that line?

But anyway, I don't think that we'll support such thing like having in the data 2 items with the same time. If you enable "development" build instead of production one you'll an error like `Assertion failed: data must be asc ordered by time, index=1, time=1524009600, prev time=1524009600`. The fact that at the time it may work isn't intended and might be broken in the future.

pavitrakumar78:
Is there any other way to add vertical lines? to candlestick/line charts?

timocov:
> Is there any other way to add vertical lines? to candlestick/line charts?

No, I don't think there is a way to do this at the time. We don't plan to add drawings though. Probably we'll provide an API to draw on the canvas from your code, but I believe it's far away from an implementation right now.

Issue #281, opened 2019-12-11
Custom metric / indicators


pleaseshutup:
I have my own API/data source hooked up to the tradingview charting library. I have a metric, lets call it "sentiment" that I want to plot along with the price. Is this possible? How do I even get started? At this point I don't think I can even do this.

Alternative: how can I include non-ohlcv data on the charts? I have data that can't be calculated just from ohlcv data. It has to be queried and plotted

timocov:
> I want to plot along with the price.

> how can I include non-ohlcv data on the charts?

What is it? Text, line, something else? Is it similar to #24?

pleaseshutup:
Lets say I have a data source like:
[
{time: 12312332432, num_positive_tweets: 453},
{time: 12312335534, num_positive_tweets: 234},
{time: 12312334343, num_positive_tweets: 343},
{time: 12312367766, num_positive_tweets: 634},
{time: 12312374545, num_positive_tweets: 112},
]
That data is already available along with my OHLCV data in my API. I want to put it on the chart.
I want to plot it as a line either over the OHLCV candles OR its own line below.

What are my options? Is this possible? How?

nikitamarcius:
@pleaseshutup 

To create indicators, need create separate array from OHLC (hight, close etc.), Then use [indicators lib](https://www.npmjs.com/package/tulind).

```javascript
const ohlc = [. . .]
const close = ohlc.map(i=> i.close)

tulind.indicators.sma.indicator([close], [3], function(err, results) {
  console.log("Result of sma is:");
  console.log(results[0]); // here is ur result [25, 33, 55, ...]
});

// From result by indicator (result[0]) need create new arr of objects [{price: 25, time: '10-10-12'}]
// NOTE: for SMA ind. first index 0-3 disabled 3 = [period]. [null, null, null, 25, 33, 55, ...]
```

timocov:
@pleaseshutup you can creat another line series for your tweets.

timocov:
I'm going to close the issue. If you'll have any question or updates - feel free to re-open it.

PR #1238, opened 2023-01-12
Plugins subsystem MVP

### LINKED ISSUE #238: Draw a line
Hi,

Is possible to draw a line using javascript?

Thank You.

eugene-korobko:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #238 #281 #698 #1049 #1064
- [x] Documentation update

**Overview of change:**

Created a MVP of the plugins subsystems.


SlicedSilver:
@eugene-korobko Could we add a graphics test case for this feature? Perhaps using the example you've previously shared.
We could also add a coverage test case for this (just a simple example which executes as much of the API code as possible).

src/api/iseries-primitive.ts
```diff
@@ -0,0 +1,98 @@
...
+	 * This method is called when viewport has been changed, so primitive have to recalculate / invalidate its data
+	 */
+	updateAllViews(): void;
+
+	/**
+	 * Returns array of labels to be drawn on the price axis used by the series
+	 *
+	 * @returns array of objects; each of then must implement ISeriesPrimitiveAxisView interface
+	 *
+	 * Try to implement this method such that the same array is returned if nothing changed, this would help the library to save memory and CPU.
```
SlicedSilver, line 77: What should be the ideal behaviour here?
- Rule 1: If the user doesn't need to change anything then they should return the same array. ✅

What should the user do if they want to make a change? Should they:
- Modify the existing array (`push`, ...) 
- or should they return a new array?

src/api/iseries-primitive.ts
```diff
@@ -0,0 +1,98 @@
+import { CanvasRenderingTarget2D } from 'fancy-canvas';
+
+/**
+ * This interface represents a label on the price or time axis
+ */
+export interface ISeriesPrimitiveAxisView {
```
SlicedSilver, line 6: Should the name include `Label` since the interface only deals with the label properties?

src/api/iseries-primitive.ts
```diff
@@ -0,0 +1,98 @@
...
+	/**
+	 * coordinate of the label, vertical for price axis and horizontal for time axis
+	 *
+	 * @returns coordinate. 0 means left and top
+	 */
+	coordinate(): number;
```
SlicedSilver, line 12: Should this type be `Coordinate` instead of number?
https://tradingview.github.io/lightweight-charts/docs/next/api#coordinate

src/tsconfig.model.json
```diff
@@ -7,6 +7,8 @@
 	"include": [
 		"./model/**/*.ts",
 		"./renderers/**/*.ts",
-		"./views/**/*.ts"
+		"./views/**/*.ts",
+		"./api/iseries-primitive.ts",
+		"./api/series-primitive-wrapper.ts",
```
kirchet, line 12: We shouldn't include anything from the API part

eugene-korobko: Look, this restriction is created to keep code clear, however with adding plugins we created a "backward" direction of dependencies. I do not think we have to write some additional code just to avoid restictions of our own.

src/api/iseries-primitive.ts
eugene-korobko: Nice comment! I've created caching based on reference to an array and did't think about possible changing array in place (we usually have anoter problem - creating new arrays with the same value). Will add this comment.

src/api/iseries-primitive.ts
eugene-korobko: This is the MVP and I tried to keep is as minimalistic as possible, so I desided that supporting labels is not required

src/tsconfig.model.json
SlicedSilver: Wouldn't your idea of moving the SDK into a separate repo (and npm package) resolve this?

src/api/iseries-primitive.ts
eugene-korobko: Yes, absolutely. I has a wrong idea that Coordinate type is not a part of API..

src/tsconfig.model.json
eugene-korobko: Well, this will be solved authomatically after moving interfaces to separate SDK package

kirchet: At least interfaces should still be a part of this package. I suggest just to move them from the API part to the model part. The model uses them so it looks like a model part. In the same way we expose for example `ChartOptions` that is declared in 'chart-model.ts' but at the same time is used the API interfaces

src/api/iseries-primitive.ts
kirchet: Not sure about it, in that case everyone who uses the library should cast `number`s to `Coordinate`s. Maybe it's more correct but requres more casts from the client code

src/tsconfig.model.json
eugene-korobko: We discussed and agreed that interfaces should be moved to separate SDK package to void plgins from requiring the library

eugene-korobko:
> @eugene-korobko Could we add a graphics test case for this feature? Perhaps using the example you've previously shared.
> 
> We could also add a coverage test case for this (just a simple example which executes as much of the API code as possible).

Added graphic test. It fails as expected

src/api/iseries-primitive.ts
eugene-korobko: Fixed

src/api/iseries-primitive.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * This interface represents a label on the price or time axis
+ */
+export interface ISeriesPrimitiveAxisView {
+	/**
+	 * coordinate of the label, vertical for price axis and horizontal for time axis
+	 *
+	 * @returns coordinate. 0 means left and top
+	 */
```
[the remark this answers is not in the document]
eugene-korobko: updated

src/api/iseries-primitive.ts
eugene-korobko: Well, my opinion: we should use coordinate everywhere or do not use it at all

eugene-korobko: On the other hand, if this interface will be moved to standalone SDK package, where Coordinate type should be defined?

Well, lets keep it a number for far?

romfrancois: Guess that's fine for now; it's just that without properly reading the doc we could interpret it differently but I think now it's OK.

SlicedSilver:
### To do

- [ ] Fix broken links on website
- [ ] Graphics e2e tests
- [ ] Coverage e2e tests
- [ ] Interactions e2e tests

SlicedSilver:
This change needs to be reviewed as it is part of the library code (not just tests and documentation changes): https://github.com/tradingview/lightweight-charts/pull/1238/commits/d4165772981d1920d057d9fa2cf3424bd19ecb16

SlicedSilver (dismissed):
Approving the recent e2e testing additions.
The rest of the changes has been previously reviewed via the #1343 PR which was merged into this branch.

SlicedSilver (approved):
Approving. The change was just to remove a link to an examples repo which hasn't been created yet. This repo will be published when the next release is ready (and the documentation will be updated).

PR #1245, opened 2023-01-19
add retries to unpkg download during website build

SlicedSilver:
**Type of PR:** bugfix

**PR checklist:**

- `N/A` Addresses an existing issue
- `N/A` Includes tests
- `N/A` Documentation update

**Overview of change:**

The CI job for building the website seems to randomly fail on the rare occasion that `unpkg.com` has an issue on their side.

```
[ERROR] Error: Cannot download file "https://unpkg.com/lightweight-charts@3.8/dist/typings.d.ts", error code=520
```

This PR adds 'retries' to the downloadFile function so that it will attempt more than once to download the file (with a short 200ms exponential backoff delay between attempts).

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -48,7 +48,13 @@ function httpGetJson(url) {
 	});
 }
 
-function downloadFile(urlString, filePath) {
+function delay(duration) {
+	return new Promise(resolve => {
+		setTimeout(resolve, duration);
+	});
+}
+
+function downloadFile(urlString, filePath, retriesRemaining = 0) {
 	return new Promise((resolve, reject) => {
 		let file;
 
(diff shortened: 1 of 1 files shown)


website/docusaurus.config.js
```diff
@@ -58,11 +64,18 @@ function downloadFile(urlString, filePath) {
...
 				url.pathname = response.headers.location;
-				downloadFile(url.toString(), filePath).then(resolve, reject);
+				downloadFile(url.toString(), filePath, retriesRemaining).then(resolve, reject);
 				return;
 			}
 
 			if (response.statusCode && (response.statusCode < 100 || response.statusCode > 299)) {
+				if (retriesRemaining > 0) {
+					logger.info(`Failed to download from ${urlString}, attempting again (${retriesRemaining - 1} retries remaining).`);
+					delay(200).then(() => {
```
edew, line 74: [minor] If we're only doing 2 retries this is fine, but if we were to do more we should probably use an exponential backoff.

edew (approved):

edew (approved):

PR #1281, opened 2023-02-22
fix unable to add markers when autoSize chart option is enabled

### LINKED ISSUE #1271: Inability to immediately add markers when `autoSize` chart option is enabled

**Steps/code to reproduce:**
1. Enabled `autoSize` in the chart options when creating a chart.
2. Add markers to the series before two requestAnimationFrames have occurred.

**Actual behavior:**

Markers aren't visible on the chart series.

**Expected behavior:**

Markers should be visible on the chart series.

**Glitch Code Sandbox link:**
https://glitch.com/edit/#!/jagged-thoughtful-tuba

The example has a few const variables at the top of the script which can be used to experiment with the test case:
```
const USE_AUTOSIZE = true;
const TRY_ADD_MARKERS_AFTER_RAF = false;
const TRY_ADD_MARKERS_AFTER_TWO_RAF = false;
const TRY_ADD_MARKERS_AFTER_5S = false;
```

victorbrambati:
**Type of PR:** bugfix

**PR checklist:**

* [x] Addresses an existing issue: fixes #1271 
* [x] Includes tests
* `N/A`  ~Documentation update~

**Overview of change:**
* The function `timeScale.isEmpty()` was returning true. This because there a condition with a `width === 0`. To resolve this [issue](https://github.com/tradingview/lightweight-charts/issues/1271), an initial size was added so that the markers could be recalculated.
* I changed the condition of `_recalculateMarkers()` to be easier to read.

**Glitch Code Sandbox link:**
https://glitch.com/edit/#!/hurricane-mahogany-output

### DIFF
--- a/src/gui/chart-widget.ts
+++ b/src/gui/chart-widget.ts
@@ -105,6 +105,7 @@ export class ChartWidget implements IDestroyable {
 			const containerRect = container.getBoundingClientRect();
 			width = width || containerRect.width;
 			height = height || containerRect.height;
+			this.resize(width, height);
 		}
 
 		// BEWARE: resize must be called BEFORE _syncGuiWithModel (in constructor only)
--- a/src/gui/chart-widget.ts
+++ b/src/gui/chart-widget.ts
@@ -105,14 +105,11 @@ export class ChartWidget implements IDestroyable {
 			const containerRect = container.getBoundingClientRect();
 			width = width || containerRect.width;
 			height = height || containerRect.height;
-			this.resize(width, height);
 		}
 
 		// BEWARE: resize must be called BEFORE _syncGuiWithModel (in constructor only)
 		// or after but with adjustSize to properly update time scale
-		if (!usedObserver) {
-			this.resize(width, height);
-		}
+		this.resize(width, height);
 
 		this._syncGuiWithModel();
 
--- a/src/model/series.ts
+++ b/src/model/series.ts
@@ -578,7 +578,7 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
 
 	private _recalculateMarkers(): void {
 		const timeScale = this.model().timeScale();
-		if (timeScale.isEmpty() || this._data.size() === 0) {
+		if (timeScale.isEmpty() || this._data.isEmpty()) {
 			this._indexedMarkers = [];
 			return;
 		}
--- a/src/model/series.ts
+++ b/src/model/series.ts
@@ -578,7 +578,7 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
 
 	private _recalculateMarkers(): void {
 		const timeScale = this.model().timeScale();
-		if (timeScale.isEmpty() || this._data.isEmpty()) {
+		if (!timeScale.hasPoints() || this._data.isEmpty()) {
 			this._indexedMarkers = [];
 			return;
 		}
--- a/src/model/time-scale.ts
+++ b/src/model/time-scale.ts
@@ -321,6 +321,10 @@ export class TimeScale {
 		return this._width === 0 || this._points.length === 0 || this._baseIndexOrNull === null;
 	}
 
+	public hasPoints(): boolean {
+		return this._points.length > 0;
+	}
+
 	// strict range: integer indices of the bars in the visible range rounded in more wide direction
 	public visibleStrictRange(): RangeImpl<TimePointIndex> | null {
 		this._updateVisibleRange();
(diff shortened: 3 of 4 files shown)


SlicedSilver:
@victorbrambati Thank you for your PR.
Having a quick look at the `graphics-tests` it appears that this additional fixes another bug I've noticed regarding the initial render not having the correct right offset (the last few bars aren't visible). 

We will conduct the code review and QA process before merging, but this will hopefully be done soon.

victorbrambati:
@SlicedSilver Surprising! Any changes that you need, you can reply me.

src/gui/chart-widget.ts
```diff
@@ -105,6 +105,7 @@ export class ChartWidget implements IDestroyable {
 			const containerRect = container.getBoundingClientRect();
 			width = width || containerRect.width;
 			height = height || containerRect.height;
+			this.resize(width, height);
```
SlicedSilver, line 108: Since the resize method is now being called if `usedObserver` is `true` or `false`.

Could we remove `this.resize` from within this `if` block and rather remove the `if (!usedObserver)` condition below?

```js
// replace this:
if (!usedObserver) {
	this.resize(width, height);
}
```
```js
// with this:
this.resize(width, height);
```

victorbrambati: Yes, have the same effect.

SlicedSilver: Thank you

SlicedSilver:
@victorbrambati I also think it would be good to have a graphics e2e test case which covers this bug. 
Is this something that you would like to work on? or would you prefer me to add the test to your PR?

I'm happy to do it but wanted to check first if you wanted to work on this.

victorbrambati:
@SlicedSilver Yes I want work on it.

SlicedSilver:
> @SlicedSilver Yes I want work on it.

Great :) Thank you.

Here is a some documentation on the graphics tests: https://github.com/tradingview/lightweight-charts/tree/master/tests#graphics

You essentially just need to make a `.js` file inside this folder: [`tests/e2e/graphics/test-cases`](https://github.com/tradingview/lightweight-charts/tree/master/tests/e2e/graphics/test-cases)

There are some examples there which you can use for inspiration, and you could do something similar to the Glitch example on the bug report.

SlicedSilver:
Thank you @victorbrambati 
The tests look good. You can clearly see that it failed on the current release (v4.0.0) and is fixed on your branch.

Golden:
1 golden

Test:
2 test

SlicedSilver (approved):

victorbrambati:
@SlicedSilver You're welcome.
This test too work to issue #1282 because the priceline is different.

src/model/series.ts
```diff
@@ -578,7 +578,7 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
 
 	private _recalculateMarkers(): void {
 		const timeScale = this.model().timeScale();
-		if (timeScale.isEmpty() || this._data.size() === 0) {
+		if (timeScale.isEmpty() || this._data.isEmpty()) {
```
kirchet, line 581: We need `timeScale` here to call `timeToIndex`.
```
const timePointIndex = ensureNotNull(timeScale.timeToIndex(marker.time, true));
```

So actually we should check that we can convert time to index.
```suggestion
		if (!timeScale.hasPoints() || this._data.isEmpty()) {
```

For now there is no `TimeScale.hasPoints` method, we need to implement it

```
public hasPoints(): boolean {
	return this._points.length > 0;
}
```

src/gui/chart-widget.ts
```diff
@@ -109,9 +109,7 @@ export class ChartWidget implements IDestroyable {
 
 		// BEWARE: resize must be called BEFORE _syncGuiWithModel (in constructor only)
 		// or after but with adjustSize to properly update time scale
-		if (!usedObserver) {
-			this.resize(width, height);
-		}
+		this.resize(width, height);
```
kirchet, line 112: This change seems unrelated to the issue and should be reverted

kirchet (changes requested):

src/gui/chart-widget.ts
SlicedSilver: Agree that the change isn't directly related to the markers issue but this change is still needed to address #1282.

This can be seen with the `initial-options/use-observer` test-case. https://app.circleci.com/pipelines/github/tradingview/lightweight-charts/1942/workflows/17dd9cbf-9dc1-4e7c-95ed-e3098cc57ae0/jobs/25041/artifacts

If I revert the change above then that test will pass again (showing the old incorrect behaviour) but in this case we want it to fail.

PR #1294, opened 2023-03-08
fix line rendering bug when out of visible range

### LINKED ISSUE #1293: Random baseline showing up in chart

**Steps/code to reproduce:**
In my case i had a chart and three baselines, one to simulate a trade profit, one to simulate a trade loss and the last one to follow the candlestick close. (Screenshot 1)
I think the bug will show with even fewer baselines, but i didn't test it.

**Actual behavior:**
As you can see in the screenshot (Screenshot 2) whenever I scroll to the left and I pass a baseline a random baseline appears in the chart, this doesn't happend if i pass the baseline scrolling to the right. I noticed also that if you programmatically set the visible range  and move to a baseline without stopping on any baseline before the bug won't show up.

**Screenshots:**
Screenshot 2023-03-05 164353

Screenshot 2023-03-05 160123

SlicedSilver:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #1293
- [x] Includes tests

**Overview of change:**
When a line based series (line, baseline, area) is on a chart where it is possible to scroll it completely off the screen to the right.
Then if the series is scrolled off screen then it will still draw on the chart.

This change corrects the `walkLine` function so that it doesn't draw anything if the `to` value is equal to zero.

### DIFF
--- a/src/renderers/walk-line.ts
+++ b/src/renderers/walk-line.ts
@@ -17,7 +17,7 @@ export function walkLine<TItem extends LinePoint, TStyle>(
 	styleGetter: (renderingScope: MediaCoordinatesRenderingScope, item: TItem) => TStyle,
 	finishStyledArea: (ctx: CanvasRenderingContext2D, style: TStyle, areaFirstItem: LinePoint, newAreaFirstItem: LinePoint) => void
 ): void {
-	if (items.length === 0 || visibleRange.from >= items.length) {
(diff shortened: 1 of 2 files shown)


romfrancois (approved):
The change looks good & makes sense to me.

kirchet (approved):

Nipheris (approved):

PR #1301, opened 2023-03-14
Ignore resize method if autoSize is active, and added API to check if active

SlicedSilver:
**Type of PR:** enhancement

**PR checklist:**

- [ ] Addresses an existing issue
- [x] Includes tests
- [x] Documentation update

**Overview of change:**
When `autoSize` chart option is enabled then we should ignore any requests to change the `height` and `width` via the `resize` method. 

Added a `autoSizeActive` method to the Chart API for checking if the setting is enabled and the ResizeObserver was successfully created.


SlicedSilver:
- Failing graphics tests are expected because a new test was added.
- `dts-changes` expected as a new method has been added to the API

eugene-korobko (approved):

PR #1330, opened 2023-04-12
fix: Add an option to disable crosshair (#749)

### LINKED ISSUE #749: Add an option to disable crosshair
We need a flag or a method to hide the crosshair. Moreover this feature should disable long tap on canvas.

The case: 
- Make the crosshair invisible
- Start the chart
- Make a long tap on canvas
- The chart looks like freezed

luk707:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #749 
- [x] Includes tests
- [x] Documentation update

**Overview of change:**

Adds option to disable the crosshair

Tasklist:

- [x] Add hidden option for crosshair
- [x] Disable rendering when hidden option enabled
- [ ] Disable certain interactions when crosshair disabled


luk707:
I've raised this incomplete PR to discuss implementation specifics regarding #749. The issuer needs to be able to disable the crosshair and certain touch interactions for their use-case with the mobile wrapper. I think it makes sense to separate this into two distinct options: `CrosshairMode.Hidden` (purely for visual changes), & another option to disable user interactions and gestures. Any comments would be appreciated :)

SlicedSilver:
@luk707 I agree that it is better to split the original request into two separate changes. 👍

I believe that the second part of the request is to disable the 'long press to activate crosshair mode' on mobile so that the user can scroll the chart even if they accidentally long press before moving their finger. I.e. more options for `TrackingMode`

0220tiezo1323 (changes requested):
Yup

edew (dismissed):

SlicedSilver (dismissed):

SlicedSilver (approved):

edew (approved):

SlicedSilver:
It doesn't make sense why the CI check for memleaks is failing.
I made an [identical branch](https://github.com/tradingview/lightweight-charts/tree/memleaks-testing), and it passes just fine (https://app.circleci.com/pipelines/github/tradingview/lightweight-charts/2189/workflows/450d1a40-d22d-4580-b38a-edfd2bd3b116).

I think we can disable the requirement for this check, merge the changes, and then re-enable it.

@edew what do you think?

PR #1335, opened 2023-04-17
use GitHub actions for website deployment

SlicedSilver:
Automatic deployment of the documentation website when pushes are made to master.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-22.04
     if: ${{ github.ref == 'refs/heads/master' }}
     permissions:
-      contents: read
+      contents: write
       pages: write
     concurrency:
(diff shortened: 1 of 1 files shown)


edew (changes requested):

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - '.github/workflows/deploy.yml'
+      - 'src/**'
+      - 'website/**'
+  pull_request:
```
edew, line 11: What does the empty `pull_request` block do?

SlicedSilver: It is included on the sample configuration: https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-docusaurus

I assume that it is probably safe to remove, but also safe to leave in.

edew (approved):

PR #1337, opened 2023-04-17
added localisation option for percentage formatter

### LINKED ISSUE #1291: PriceScale Percentage formatting
version: 4.0.0

Hello guys! Can someone help me with formatting percentage values? It would be nice if I can provide precision here to see more accurate result. Instead of 58.80 -> 58.80123 more decimals for example.

I am using global percentage mode for my chart, and have no idea how to lead formatter to percent values. By the way my localization formatter works only with Normal Price Scale Mode.

I would be grateful for any advice, thanks!

image
image

### LINKED ISSUE #1328: Format left price scale
I'd like to format the percentage labels of the left price scale. Currently it is not possible to format the rendered label values. I can only format the right price scale. But maybe I'm missing something.

I want to remove the last two digits of the percentage value.

```ts
      leftPriceScale: {
        visible: true,
        mode: PriceScaleMode.Percentage,
        borderVisible: false,
      },
```

```ts
    this._percentage = this._chart.addLineSeries({
      priceScaleId: 'left',
      color: 'rgba(92, 163, 239, 0)',
      baseLineVisible: false,
      priceLineVisible: false,
      lastValueVisible: false,
      priceFormat: {
        type: 'custom',
        formatter: (price: number) => {
          return `${price.toFixed(0)}%`;
        },
      },
    });
```

SlicedSilver:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #1328, fixes #1291
- [x] Includes tests
- [x] Documentation update

**Overview of change:**
Add support for overriding the percentage formatter within the localisation options.

This adds the ability to support locales which don't place the percentage sign at the end of the number ([examples](https://phrase.com/blog/posts/number-localization/#toc_5)), and additionally allows the formatter to show more or fewer decimal points.

### DIFF
--- a/src/model/price-scale.ts
+++ b/src/model/price-scale.ts
@@ -12,6 +12,7 @@ import { Coordinate } from './coordinate';
 import { FirstValue, IPriceDataSource } from './iprice-data-source';
 import { LayoutOptions } from './layout-options';
 import { LocalizationOptions } from './localization-options';
+import { PriceFormatterFn } from './price-formatter-fn';
 import { PriceRangeImpl } from './price-range-impl';
 import {
 	canConvertPriceRangeFromLog,
@@ -766,7 +767,7 @@ export class PriceScale {
 
 	public formatPricePercentage(price: number, baseValue: number): string {
 		price = toPercent(price, baseValue);
-		return this._formatPercentage(price);
+		return this._formatPercentage(price, percentageFormatter);
 	}
 
 	public sourcesForAutoScale(): readonly IPriceDataSource[] {
@@ -1002,23 +1003,22 @@ export class PriceScale {
 		return null;
 	}
 
-	private _formatPrice(price: BarPrice, fallbackFormatter?: IPriceFormatter): string {
-		if (this._localizationOptions.priceFormatter === undefined) {
+	private _formatValue(value: BarPrice | number, formatter: PriceFormatterFn | undefined, fallbackFormatter?: IPriceFormatter): string {
+		if (formatter === undefined) {
 			if (fallbackFormatter === undefined) {
 				fallbackFormatter = this.formatter();
 			}
-
-			return fallbackFormatter.format(price);
+			return fallbackFormatter.format(value);
 		}
 
-		return this._localizationOptions.priceFormatter(price);
+		return formatter(value as BarPrice);
 	}
 
-	private _formatPercentage(percentage: number): string {
-		if (this._localizationOptions.percentageFormatter === undefined) {
-			return this.formatter().format(percentage);
-		}
+	private _formatPrice(price: BarPrice, fallbackFormatter?: IPriceFormatter): string {
+		return this._formatValue(price, this._localizationOptions.priceFormatter, fallbackFormatter);
+	}
 
-		return this._localizationOptions.percentageFormatter(percentage);
+	private _formatPercentage(percentage: number, fallbackFormatter?: IPriceFormatter): string {
+		return this._formatValue(percentage, this._localizationOptions.percentageFormatter, fallbackFormatter);
 	}
 }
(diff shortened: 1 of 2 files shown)


src/model/price-scale.ts
```diff
@@ -1012,4 +1013,12 @@ export class PriceScale {
 
 		return this._localizationOptions.priceFormatter(price);
 	}
+
+	private _formatPercentage(percentage: number): string {
+		if (this._localizationOptions.percentageFormatter === undefined) {
+			return this.formatter().format(percentage);
```
eugene-korobko, line 1019: Originally we call `percentageFormatter.format(price)` here, now we use regular formatter, why did we change it?
Do we have tests (maybe graphics?) for percentage scale?

SlicedSilver: So the part which is different is the formatter used within `formatPricePercentage()`. 
- This method is only used by `lastValueData` within the `Series` class (`src/model/series.ts`) to populate a value called `formattedPricePercentage`.
- This value is only used within `src/views/price-axis/series-price-axis-view.ts`. When this is used then it is always checking that the price scale is a percentage scale (`this._source.priceScale().isPercentage()`) 
- thus the result should be the same because the formatter should already be correctly set to the percentage formatter.

So essentially the `percentageFormatter.format` isn't actually ever used for any values displayed.

I'll fix the code such that if the `formatPricePercentage` is called on a non-percentage scale that it can default back to the original. This isn't currently used, but could be in the future.

There is a graphics test now for the percentage scale formatter.

I think we need to add an e2e test for the series-api and the formatter function. Looks like there is a potential uncovered bug, but also to catch future bugs

src/model/price-scale.ts
```diff
@@ -1002,23 +1003,22 @@ export class PriceScale {
 		return null;
 	}
 
-	private _formatPrice(price: BarPrice, fallbackFormatter?: IPriceFormatter): string {
-		if (this._localizationOptions.priceFormatter === undefined) {
+	private _formatValue(value: BarPrice | number, formatter: PriceFormatterFn | undefined, fallbackFormatter?: IPriceFormatter): string {
+		if (formatter === undefined) {
```
eugene-korobko, line 1007: This is not a bug, however the whole function looks like 

return (formatter ?? fallbackFormatter ?? this.formatter()).format(value);

SlicedSilver: So the current code actually compiles to the smallest bundle size (and is maybe easier to read). Using one of the options below increases the bundle size by about `0.15 kB`. Your thoughts?

### Option 1:

```ts
private _formatValue(value: BarPrice | number, formatter: PriceFormatterFn | undefined, fallbackFormatter?: IPriceFormatter): string {
	return formatter?.(value as BarPrice) ?? fallbackFormatter?.format(value) ?? this.formatter().format(value);
}

private _formatPrice(price: BarPrice, fallbackFormatter?: IPriceFormatter): string {
	return this._formatValue(price, this._localizationOptions.priceFormatter, fallbackFormatter);
}

private _formatPercentage(percentage: number, fallbackFormatter?: IPriceFormatter): string {
	return this._formatValue(percentage, this._localizationOptions.percentageFormatter, fallbackFormatter);
}
```

### Option 2:

```ts
// remove _formatValue

private _formatPrice(price: BarPrice, fallbackFormatter?: IPriceFormatter): string {
	return this._localizationOptions?.priceFormatter?.(price) ?? fallbackFormatter?.format(price) ?? this.formatter().format(price);
}

private _formatPercentage(percentage: number, fallbackFormatter?: IPriceFormatter): string {
	return this._localizationOptions?.percentageFormatter?.(percentage) ?? fallbackFormatter?.format(percentage) ?? this.formatter().format(percentage);
}
```

eugene-korobko (approved):

PR #1343, opened 2023-04-21
Plugins and abstract series

SlicedSilver:
## Main changes:
- Added lifecycle methods `attached` and `detached` to the primitive interface.
	- These allow the plugin to know when it has been attached or detached from the series which is useful for creating and destroying things like event listeners
- The `attached` method provides 3 items to the primitive. 
	- The first two are the chart and series to which the primitive has been attached. This results in less code for the primitive end-user because the initial plugin examples generally required being created with chart and series.
	- The third parameter is a callback which can be used to tell the chart to 'update'. This is useful for primitives which would like to trigger redraws at specific times. For example, the anchored text example could be used to display a countdown clock. Without this update method, the primitive wouldn't be able to trigger an update and could only be re-rendered when the user interacts with the chart.
- Added a `data()` getter on the series api. This allows the user (or a plugin) to get the complete set of data for the series. This is essentially an extension of the existing `dataByIndex` getter.
	- This is useful for primitives so that they can draw items related to the data (such as an indicator) without relying on the developer having to keep updating the data within the primitive as well as the series.
- Added  `subscribeDataChanged`  which allow subscribing to an event which occurs whenever the data for the series is changed (either via `setData`, or `update`).
	- This event includes whether the change was 'full' or 'update', so the listener can determine if they need to get a copy of all of the data, or just the final data point in order to keep a separate copy of the data in sync.
- Added support for primitives drawing on the time and price scales (similar to what was possible on the main pane).
	- The primitive can only draw on the price scale associated with the series.
- Added `zOrder` options for bottom, normal and top for the paneViews (main chart, time, and price scales) so that it is possible to draw on any of the canvases. Primary use cases include drawing watermarks below the grid lines, and drawing on top of the crosshair lines (like a plus button on a crosshair label).
	- Technically there is actually one more canvas level available (top, but below crosshair) but not sure where it would be useful.
- Added `fixedCoordinate` to the AxisViews so that the labels can be defined at a fixed location (instead of automatically moving to find a space to fit).
- Made as much of the primitive interface optional so that primitive (plugin) developers don't need to write as much boilerplate code containing methods which just return empty arrays.
- Added `autoscaleInfo` property to the primitive interface which allows primitives to extend the base autoscaleInfo so that they can ensure that the chart is automatically scaled to fit the drawings if desired. Includes a warning that this should be a highly optimised method since it could be called very often.
- Added the ability to include any extra data on the built-in series' data with the `customValues` field. This allows users to provide additional info on each data point which could be used in a primitive or within event handlers. The customValues object will be retained when processed by the library and later returned via one of the data getters. Normally additional fields are stripped and removed.
- Added abstract series. An abstract series is essentially a base series which requires a custom renderer to be provided. This allows developers to create their own chart types (and potentially allow us to refactor our library in the future to allow tree-shaking unused chart types). See the `IAbstractSeriesPaneView` interface and `addAbstractSeries` method.
	- Abstract series can have their own Data Structure which will be inferred correctly by Typescript and ensure type safety with things like `setData` and `update`
	- Similarly the types are inferred for the options.
	- The main restrictions for an abstract series is that it uses the same time and barSpacing constraints of the current built-in chart types, so it isn't possible to create bars where the width is determined by the volume as an example.
	- The view provides a way for an abstract series to provide the 'equivalent' OHLC values which are used for autoscale, and crosshair positions.
- Added `visible` and `tickVisible` to the AxisViews to cover showing and hiding labels
- Added documentation for these additions to the library. See the Plugins section on the documentation preview.
- Added `chartElement` getter method to the chart API. This is useful for plugins which need to know the exact size of the entire chart, or need to add their own event listeners (such as mouse events for the entire chart including the scales).
- Added `setCursorStyle` method to the chart API. This is useful for plugins to be able to set the cursor style, such as when hovering over something that should be interactive. This is exposed on the chart API to save some KB and because it may be useful to normal developers not implementing a plugin.

#### Documentation preview:

- Documentation preview (choose `next` in the version dropdown): https://lighthearted-marshmallow-fcb80c.netlify.app/lightweight-charts/ 
- Plugins page on documentation preview: https://lighthearted-marshmallow-fcb80c.netlify.app/lightweight-charts/docs/next/plugins/intro

**👨‍💻 A repo contains a few example plugins is available on the internal GitLab: `msilverwood/lwc-plugin-examples`.**

### Note to reviewers

- ~`src/tsconfig.model.json` has been modified. Please check if this change makes sense.~
- I think it would be better to rename `priceAxisViews` to `priceAxisLabelView`, and `timeAxisViews` to `timeAxisLabelView`. This is because it is potentially confusing since I've added `priceAxisPaneViews` and `timeAxisPaneViews`. 
	- Unless we plan to extend those views to support more than labels in the future.
- I'm not set on any of the names currently used. We should discuss if the following names are the best, or not:
	- abstract → for the custom series ✅
	- ~primitive → for the drawings~
	- plugins ✅ / ~extensions~ for the overall grouping of these features
- The documentation added needs work. I would appreciate feedback on what needs to be explained better.
	- ~I forgot to add the TM where I've mentioned Lightweight Charts™️~

### Still to Do
- [ ] Add more tests (graphics & coverage & unit tests).
- [x] Unsubscribe method for data changed events.

### Additional Thoughts

~I think that there might be a use-case for being able to **attach primitives to the chart** (in addition to a series). There are a few plugins which would make more sense to be attach to the chart, and which don't need a price scale for converting prices into y coordinates. Example: watermark, legend, attribution tradingview logo.~ **Update:** adding this requires quite a lot of boilerplate code which adds bloat to the library size. Probably best to leave this until it is really needed. There is a workaround which is to add an empty series for 'chart' primitives.

~Additionally, it might be nice to have a getter for either the **total chart height**, or just the height of the main plot area. Currently this can only be measured using the DOM api.~ This can be done using the `chartElement()` method and checking the `clientHeight` and `clientWidth` properties.

It would also be useful to have a getter on the Chart for all the series (`getAllSeries()`) which could be used by a legend-style plugin which would like to know which series are available. This is probably not needed because [seriesData](https://tradingview.github.io/lightweight-charts/docs/api/interfaces/MouseEventParams#seriesdata) can be used within a crosshairMoved event handler. So it's main use case would be getting this information before a user has interacted with the chart with a crosshair. If we do add `getAllSeries()` then it would make sense to also have a subscriber for knowing when the list has been changed.


eugene-korobko:
```Added subscribeDataChanged which allow subscribing to an event which occurs whenever the data for the series is changed (either via setData, or update).```

Well, this looks inconsistent with other functions. Primitive optionally implements attached method to be notified, why not to go the same way here and to add 'dataChanges' optional method?

**Comment by eugene-korobko** [comment]
From technical point of veiw, custom series sounds better than abstract, becuase of "abstract" means class whose objects cannot be instantiated.

**Comment by SlicedSilver** [comment]
> From technical point of veiw, custom series sounds better than abstract, becuase of "abstract" means class whose objects cannot be instantiated.

Agreed, I will change it to Custom Series. I used 'abstract' because there is a need to 'supply' something (renderer in the case of the implementation, and the class in the case of the addAbstractSeries). As we know, naming things is difficult 😅

File: src/gui/time-axis-widget.ts [hunk 1]
```diff
@@ -315,6 +332,33 @@ export class TimeAxisWidget implements MouseEventHandlers, IDestroyable {
...
 			this._drawLabels([...this._chart.model().serieses(), this._chart.model().crosshairSource()], topTarget);
+			this._drawAdditionalSources(topTarget, sourceTopPaneViews);
+		}
+	}
+
+	public refreshCursorType(): void {
+		this._setCursor(this._defaultCursorType);
+	}
+
+	private _drawAdditionalSources(target: CanvasRenderingTarget2D, paneViewsGetter: IPaneViewsGetter): void {
```
eugene-korobko, line 343: Why do we use pane views here? We have time axis views/renderers

src/model/iabstract-series.ts
```diff
@@ -0,0 +1,143 @@
...
+	 */
+	time: number;
+	/**
+	 * Original data for the item.
+	 */
+	originalData: TData;
+	/**
+	 * Color assigned for the item, typically used for price line and price scale label.
+	 */
+	barColor: string;
```
eugene-korobko, line 30: We already have `color?: string;` in the AbstractData, what is the difference between it and barColor?

src/model/iabstract-series.ts
```diff
@@ -0,0 +1,143 @@
...
+	 * @param isHovered - Whether the series is hovered.
+	 * @param hitTestData - Optional hit test data for the series.
+	 */
+	draw(
+		target: CanvasRenderingTarget2D,
+		priceConverter: PriceToCoordinateConverter,
+		isHovered: boolean,
+		hitTestData?: unknown
+	): void;
+	// hitTest?(x: Coordinate, y: Coordinate): HoveredObject | null;
```
eugene-korobko, line 81: Why don't we need hittest?

src/model/iabstract-series.ts
```diff
@@ -0,0 +1,143 @@
...
+ * OHLC equivalent values for the custom series.
+ *
+ * The order of items is defined as follows:
+ *
+ * - open
+ * - high
+ * - low
+ * - close
+ */
+export type AbstractSeriesPricePlotValues = [number, number, number, number];
```
eugene-korobko, line 94: As I said above, I do not see why we suppose absract series to have ohlc data. If we do not support custom data, we do not need custom series (since series is a part of the model), just custom views/renderers

src/model/series.ts
```diff
@@ -130,6 +163,8 @@ export class Series<T extends SeriesType = SeriesType> extends PriceDataSource i
 			this._lastPriceAnimationPaneView = new SeriesLastPriceAnimationPaneView(this as Series<'Area'> | Series<'Line'> | Series<'Baseline'>);
 		}
 
+		this._abstractPaneView = abstractPaneView ?? null;
```
eugene-korobko, line 166: having both _abstractPaneView and _paneView looks strange, what about providing pane view factory, not instance? This would allow to implement all the types with unified way

src/tsconfig.model.json
```diff
@@ -7,6 +7,8 @@
 	"include": [
 		"./model/**/*.ts",
 		"./renderers/**/*.ts",
-		"./views/**/*.ts"
+		"./views/**/*.ts",
```
eugene-korobko, line 10: Well, we've already discussed this.
Originally there was an idea to have internal structure where model knows nothing about GUI and both model and gui know nothing about api.

Like MVC: model does not know about views and controller, views know mode and do now know controller, controllers knows both

So, lets review why model now imports giu and api

src/api/get-series-data-creator.ts
```diff
@@ -22,14 +24,20 @@ import {
...
 };
 
 function singleValueData(plotRow: PlotRow): SingleValueData {
-	return {
+	const data: SingleValueData = {
 		value: plotRow.value[PlotRowValueIndex.Close],
 		time: plotRow.originalTime as unknown as Time,
 	};
+
+	if (plotRow.customValues !== undefined) {
```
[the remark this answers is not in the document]
SlicedSilver: The reason for this check is to ensure that the data only includes a `customValues` property if it was supplied. If this check is removed then whenever data is returned back from the library (like with `series.dataByIndex()`) it will contain `customValues: undefined`. 

If we want more concise source code then we could do this 👇 but it doesn't decrease the bundle size so it's just a coding style change.
```js
return {
  value: plotRow.value[PlotRowValueIndex.Close],
  time: plotRow.originalTime as unknown as Time,
  ...(plotRow.customValues ? { customValues: plotRow.customValues } : {})
};
```

src/api/get-series-plot-row-creator.ts
```diff
@@ -96,23 +98,44 @@ function getCandlestickSeriesPlotRow(time: TimePoint, index: TimePointIndex, ite
 	return res;
 }
 
+// The returned data is used for scaling the series, and providing the current value for the price scale
+export type AbstractDataToPlotRowValueConverter = (item: AbstractData | WhitespaceData) => [
```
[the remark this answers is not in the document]
SlicedSilver: > all the plots are used for scaling, the last one is used as current value

👍This makes more sense

src/api/get-series-plot-row-creator.ts
```diff
@@ -96,23 +98,44 @@ function getCandlestickSeriesPlotRow(time: TimePoint, index: TimePointIndex, ite
...
 	return (row as Partial<SeriesPlotRow>).value !== undefined;
 }
 
 type SeriesItemValueFnMap = {
-	[T in keyof SeriesDataItemTypeMap]: (time: TimePoint, index: TimePointIndex, item: SeriesDataItemTypeMap[T], originalTime: OriginalTime) => Mutable<SeriesPlotRow<T> | WhitespacePlotRow>;
+	[T in keyof SeriesDataItemTypeMap]: (time: TimePoint, index: TimePointIndex, item: SeriesDataItemTypeMap[T], originalTime: OriginalTime, dataToPlotRow?: AbstractDataToPlotRowValueConverter) => Mutable<SeriesPlotRow<T> | WhitespacePlotRow>;
 };
 
+function wrapCustomValues<T extends SeriesPlotRow | WhitespacePlotRow>(plotRow: Mutable<T>, bar: SeriesDataItemTypeMap[SeriesType]): Mutable<T> {
+	if (bar.customValues !== undefined) {
```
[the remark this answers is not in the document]
SlicedSilver: Agreed, if the `update` doesn't have customValues then it shouldn't keep the old value. I'll add a test case for this.

src/api/get-series-plot-row-creator.ts
```diff
@@ -123,6 +146,7 @@ const seriesPlotRowFnMap: SeriesItemValueFnMap = {
 	Baseline: wrapWhitespaceData(getBaselineSeriesPlotRow),
 	Histogram: wrapWhitespaceData(getColoredLineBasedSeriesPlotRow),
 	Line: wrapWhitespaceData(getColoredLineBasedSeriesPlotRow),
+	Abstract: getAbstractSeriesPlotRow,
```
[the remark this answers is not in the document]
SlicedSilver: I will double check this. From my memory, since a custom series has its own data structure it no longer makes sense to check for `value` or `open` values to determine if a point is whitespace. Additionally we don't need to wrap for customValues since an custom series can have any data structure anyway.

src/model/iabstract-series.ts
SlicedSilver: I'll add `hitTest` 👍

src/model/iabstract-series.ts
SlicedSilver: The price plot values are used for the autoscale and current value for things like the crosshair. Since the custom series data can be of any structure we need a way to know these price values. The custom series doesn't require that the data contains OHLC. This is essentially just a shorthand for getting `{max: number, min: number, current: number}` for each data point so that we can autoScale and show the crosshair correctly.

I agree that we don't need it to be 4 items. I'm simplify this for the user.

src/model/iabstract-series.ts
SlicedSilver: The difference is that `color` is the property which can be set by the end-user when creating their data such that they can control the color used for the price line and label for that specific point.

The library when processing the data, and then finally providing the data to the renderer will use this color if it is available (or fallback to the series options if not) and define it as `barColor`. We do this for all the series types, even lineSeries.

`barColor` is the value only passed to the renderer in this specific case, and it is the expected name to be used by the current price line and label.

src/api/get-series-data-creator.ts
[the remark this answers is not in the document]
SlicedSilver: Implemented the more concise code style mentioned above.

src/api/get-series-plot-row-creator.ts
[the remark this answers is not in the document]
SlicedSilver: There is no longer a restriction to return an array of four numbers. It can be only any length greater than zero.

src/api/get-series-plot-row-creator.ts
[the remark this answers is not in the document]
SlicedSilver: Added a test case for this within the data-layer unit tests.

src/api/get-series-plot-row-creator.ts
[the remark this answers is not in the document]
SlicedSilver: Refactored the code so that custom series now support WhitespaceData correctly. Tested with the example plugins.

src/gui/time-axis-widget.ts
SlicedSilver: Removed the usage of 'PaneView' and created types for time and price views.

src/model/iabstract-series.ts
SlicedSilver: Refactored (as mentioned above) to allow an array of any length instead of requiring 4 items.

src/model/series.ts
SlicedSilver: Refactored, `_abstractPaneView` property has been removed.

src/tsconfig.model.json
SlicedSilver: I need to spend a bit more time on this one. It's a bit complicated to debug but essentially this restriction was bypassed so that the attached lifecycle hook could have the API types returned (IChartApi, and ISeriesApi). Since then I think there are more parts of the code breaking this rule. Mainly because the CustomSeries is kinda both a model and an API.

Should we duplicate types so we can fix this problem, or perhaps pursue a different approach? Maybe the types for CustomSeries can be in their own folder (which is both API and model/view)?

src/api/chart-api.ts
```diff
@@ -258,17 +284,31 @@ export class ChartApi implements IChartApi, DataUpdatesConsumer<SeriesType> {
 		return this._chartWidget.autoSizeActive();
 	}
 
-	private _addSeriesImpl<TSeries extends SeriesType>(
+	public chartElement(): HTMLDivElement {
+		return this._chartWidget.element();
+	}
+
+	public setCursorStyle(style: string | null): void {
```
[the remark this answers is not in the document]
SlicedSilver: Adding HitTest to the primitives is simple enough when we only need to consider the normal zOrder placement. If we want to support this correctly then it requires correctly sorting all the pane views via source order and zOrder before running the hit tests.

What do you think about the Primitive having the hit test instead of its internal views / renderers? That way it is simpler and faster for the library to check and run. If the primitive has multiple views and needs the hit test logic in the renderer then that's an implementation detail for the plugin developer.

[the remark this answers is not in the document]
SlicedSilver: Added hitTest to primitives. The primitive is responsible for returning a single hit test result even if it has a few views. This simplifies the code for the library so it doesn't need to test every view that the primitive supplies and then also determine it's z stack order.

The hit test result from the primitive can include a cursor style to be used on the main chart pane. This can be used instead of the setCursorStyle method to change the cursor for things like buttons drawn via the primitive.

src/model/iabstract-series.ts
SlicedSilver: https://github.com/tradingview/lightweight-charts/pull/1343#discussion_r1195716125

src/api/get-series-data-creator.ts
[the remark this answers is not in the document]
kirchet: The spread operator compiles with the current target to very long expression that uses `Object.assign`. Also if the targed object type is internal it will not work correctly.

```ts
export interface ExternalAPIInterface {
  a: number;
}

interface InternalInterface {
  a: number;
}

const a1: ExternalAPIInterface = { a: 1 };
const a2: InternalInterface = { a: 2 };
```

The code above compiles to

```js
const a1 = { a: 1 };
const a2 = { _internal_a: 2 };
```

So we cannot use `const a2: InternalInterface = { ...a1 } `. It compiles without any errors but works incorrectly.

https://github.com/timocov/ts-transformer-properties-rename

So I suggest to avoid the spread operator or use it very carefully

src/api/get-series-plot-row-creator.ts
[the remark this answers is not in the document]
kirchet: > Agreed, if the update doesn't have customValues then it shouldn't keep the old value. I'll add a test case for this

What if we want to replace a custom series value with a whitespace?

src/api/get-series-data-creator.ts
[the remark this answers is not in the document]
SlicedSilver: I've changed the code back to previous version which didn't use the spread operator.
In this case, it did still work correctly when using production minified build but reverting back to the previous approach did reduce the bundle size slightly.

src/api/get-series-plot-row-creator.ts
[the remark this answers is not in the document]
SlicedSilver: I'll add some unit tests for this situation.

eugene-korobko:
> xpose it on the SeriesApi

Yes, optional interface method does not make creating custom code more complex

src/tsconfig.model.json
eugene-korobko: I have no straight answer here. I understand the pros and cons for this restriction. Maybe we need a discusson of all the team about it?

src/model/iabstract-series.ts
eugene-korobko: >I think that we should add hitTest to the primitives

Yes, sorry my comment was wrong, we do not need hittest in model, so lets juts remove the commented line

src/api/chart-api.ts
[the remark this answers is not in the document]
eugene-korobko: Well, I do not like hittest method in the model, however I agree that  we do not want to copy the same code of calculating top-most hittest to several places. Maybe free function like "sourceHittest" would be better?

src/api/get-series-plot-row-creator.ts
```diff
@@ -99,15 +99,14 @@ function getCandlestickSeriesPlotRow(time: TimePoint, index: TimePointIndex, ite
...
-];
+export type CustomDataToPlotRowValueConverter = (item: CustomData | WhitespaceData) => number[];
 
 function getCustomSeriesPlotRow(time: TimePoint, index: TimePointIndex, item: CustomData | WhitespaceData, originalTime: OriginalTime, dataToPlotRow?: CustomDataToPlotRowValueConverter): Mutable<SeriesPlotRow<'Custom'>> {
-	const value = ensureDefined(dataToPlotRow)(item);
+	const values = ensureDefined(dataToPlotRow)(item);
+	const max = Math.max(...values);
+	const min = Math.min(...values);
+	const last = values[values.length - 1];
+	const value: PlotRowValue = [last, max, min, last];
```
eugene-korobko, line 109: Not sure why so we still need 4 numbers here, what is the semantic?

SlicedSilver: The value needs to be of type `PlotRowValue` as defined within `PlotRow`. This is the same for all the current series types. For example, for the Line series we set this value to `[val, val, val, val]` (see `function getColoredLineBasedSeriesPlotRow`) even though there is only a single value for a line series. 

The internals of the library expect that this is an array of 4 numbers (which are representing OHLC). I suspect that this is done to reduce the bundle size, and simplify the internal code. 

If we want to change this then it is probably a different PR since it will make sense to change it for all the single value series types (`line`, `area`, `baseline`, `histogram`) as well.

src/api/chart-api.ts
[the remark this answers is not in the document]
SlicedSilver: Sorry, I'm don't fully understand.

The latest change is that:
- The existing hit test logic will now also call the `hitTest` method (part of the ISeriesPrimitive interface) of all the primitives attached to all the sources. If the primitive hitTest is from the top zOrder then it will pick that result as the hit. For the other zOrders then we will check the normal hitTests, and pick the 'top' hitTest if found (could be from a primitive or a built-in).

Where do you suggest that the sourceHittest function be placed? The `findBestPrimitiveHitTest` function is a free function which is being evoked within the existing `hitTest` method of the `PaneWidget`. Is the suggestion that the logic inside `PaneWidget.hitTest` be extracted into a free function that accepts `sources`? For example, the `_hitTestPaneView` method doesn't need to be method of the class (we could extract it into a free function).

src/model/iabstract-series.ts
SlicedSilver: Removed comment from `icustom-series.ts` file.

src/tsconfig.model.json
SlicedSilver: I've refactored the code so that the `tsconfig.model.json` can be returned to how it was.

src/api/iseries-primitive-api.ts
```diff
@@ -0,0 +1,33 @@
...
+	 */
+	requestUpdate: () => void;
+}
+
+/**
+ * Interface for series primitives. It must be implemented to add some external graphics to series.
+ */
+export type ISeriesPrimitive = ISeriesPrimitiveBase<
+	SeriesAttachedParameter<SeriesType>
+>;
```
SlicedSilver, line 33: On the generated documentation site, all the types from Base are correctly showing except `hitTest` which can only be seen on the `ISeriesPrimitiveBase` page. Not sure why the types for everything else is coming across correctly except for `hitTest` 🤔

src/api/chart-api.ts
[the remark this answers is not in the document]
SlicedSilver: > I do not understand how primitive could use "serCursorStyle" method becuase of the primitive can't know if it is top-most.

`setCursorStyle` has been removed.

The primitive can have a hitTest function which should return the 'top' hit test of the views it contains. As part of that hitTest result it includes the z order layer of the it (`top`, `normal`, `bottom`). The library can then determine which hitTest to use because it knows the sources order, the normal library hit tests and the primitive hit tests. It can determine which hitTest is on top if there are multiple 'hits'.

> So I suppose the primitive to optionally add "cursor" property to the hittest result and if this primitive is the topmost one, pane widget will set corresponding cursor if required.

This has been implemented now. The primitive hitTest can have a cursorStyle property. If that hitTest is the best hit then the chart will show that cursor style.

eugene-korobko (approved):

PR #1348, opened 2023-04-24
Implemented line-based series' point markers support

### LINKED ISSUE #365: Scatter chart
Is it possible to create a simple scatter chart without a line, and only showing data points like you can do in excel. I looked at the markers but you cannot specify both x and y coordinates for it.

 
image

kirchet:
**Type of PR:** 

**PR checklist:**

- [x] Addresses an existing issue: fixes #365
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

```
const mainSeries = chart.addLineSeries({ pointMarkersVisible: true });
```

image

```
const mainSeries = chart.addLineSeries({ pointMarkersVisible: true, lineVisible: false });
```

image


.size-limit.js
```diff
@@ -4,21 +4,21 @@ module.exports = [
 	{
 		name: 'CJS',
 		path: 'dist/lightweight-charts.production.cjs',
-		limit: '44.46 KB',
+		limit: '44.7 KB',
```
SlicedSilver, line 7:
```suggestion
		limit: '44.75 KB',
```

SlicedSilver:
If we stick with `MediaCoordinatesRenderingScope` instead of `BitmapCoordinatesRenderingScope` then we can reduce the build size slightly because of the fewer `* horizontalPixelRatio` and `* verticalPixelRatio` parts of code.

The issue is the correction code in `walk-line.ts` which needs the pixelRatio:
```ts
const tickWidth = Math.max(1, Math.floor(horizontalPixelRatio));
const correction = (tickWidth % 2) / 2;
```

So perhaps we could add `horizontalPixelRatio` and `verticalPixelRatio` to the information that fancy-canvas supplies for the media scope, and thus only use the pixelRatio in the correction code.

Here is a comparison of the size difference:
| Build | Bitmap | Media | Difference |
|--------|--------|--------|--------|
| CJS | 44.74 kB | 44.69 kB | -0.05 kb |
| ESM | 44.66 kB | 44.61 kB | -0.05 kb |
| Standalone ESM | 46.36 kB | 46.29 kB | -0.07 kb |
| Standalone | 46.41 kB | 46.34 kB | -0.07 kb |

It's a tiny improvement but perhaps it could be made in other places as well.

SlicedSilver:
```
ESM
  Package size limit has exceeded by 1 B
  Size limit: 44650 B
  Size:       44651 B gzipped
```
🙄🤬

I guess that the date in the version string isn't as compressible (gzip) as last time. 🤷‍♂️

abhi-agridence:
Hi, any timeline for this PR to go to release soon ? Scatter chart is quite important when we want to show comparative self data distribution vs stock price.

SlicedSilver:
> Hi, any timeline for this PR to go to release soon ? Scatter chart is quite important when we want to show comparative self data distribution vs stock price.

HI Abhi, this pull request still needs a code review. Once we have completed that then we will merge it into the master branch. Once it is in the master branch then it will be included within the next version release. At this time, we don't have a date for the next version since there are a few other features and issues which are still being worked on.

Since this is an open source project, you could check out this branch and build your own version of the library containing this feature.

FYI: this feature isn't a 'scatter plot' but rather a line plot where you can find the line and only show the points. The main difference being that a single x axis point can only have a single price point, and that the x axis isn't guaranteed to be linear.

kirchet:
> @kirchet What are your thoughts on splitting / refactoring the internal logic of the `walkLine` function (within `src/renderers/walk-line.ts`)? The function is currently responsible for drawing the line and drawing the point markers using two separate loops. Could we extract those two separate drawing actions into new functions? It would also help reduce the complexity of that function.
> 
> Ideally we would only want to loop over the items once but I suspect that the two separate loops are required to ensure that the points are plotted on top of the line and the logic required to draw the point markers on the previous item during a single loop might be just as expensive for the CPU Usage / bundle size.

Done https://github.com/tradingview/lightweight-charts/pull/1348/commits/228844d4b5281e40f06d9ac27430036ec30c04e5

SlicedSilver:
Increased the bar spacing in the e2e graphics tests so that the line portion is more visible.
barSpacing-change

SlicedSilver (approved):

SlicedSilver (approved):
Approving again after updating size-limit so it can be merged with master.

Obiwanbde:
Hi,
I hope this is the right place to inform about my issue for scatter lines. 

I have simply draw a candlestick chart and add a new line series with:

`chart.addLineSeries({ pointMarkersVisible: true, lineVisible: false }).setData([` 

but I do not see any difference than a regular line and no points. Even changing as pointMarkersVisible: false, results a visible line. 
psarattempt


You can see my code in [https://jsfiddle.net/Obiwanbde/dqy1kgpv/2/](url)

Can you help find what am I doing wrong ? or the library is not available yet for standalone usage?

Thanks in advance.

SlicedSilver:
Hi @Obiwanbde 
This addition to the library hasn't been released yet. It has been merged to master, but will only be released when create the next version of the library (most likely `4.1`) and publish it on NPM. If you would like to use the feature right now then you could download the repo, and build your own version using the master branch.

PR #1360, opened 2023-05-17
Spring cleaning

SlicedSilver:
**Type of PR:** enhancement

**Overview of change:**

A variety of changes to mainly reduce the bundle size slightly.

- ~The main change is to remove the list of css colors, and rather use the browser to convert the names for us.~
  - ~The unit test for named colors has been moved to a graphics e2e test because it now requires the Browser API.~
  - This change has been reverted. Perhaps a future version could switch to a different approach?
- removed log10 function since the native version is supported on everything except ie11, but we were already using the native version in the code so ie11 support was already broken...
- Since we only use the roundRect function with the 4 item array, we can remove the extra code for handling the other lengths of array. Additionally, we should be able to switch to the native version in the future since Firefox has added support.
- The `lowerbound` and `upperbound` functions were almost identical. Refactored to be a single function.

Each commit has a change to the `size-limit` file so it is clear what difference each commit makes to the bundle size.

**Bundle Size Changes**
```
44.46 → 44.31
44.38 → 44.25
46.08 → 45.95
46.13 → 45.99
```


src/helpers/mathex.ts
```diff
@@ -24,27 +24,11 @@ export function equal(x1: number, x2: number, epsilon: number): boolean {
...
-	let minVal = arr[0];
-	for (let i = 1; i < arr.length; ++i) {
-		if (arr[i] < minVal) {
-			minVal = arr[i];
-		}
-	}
-
-	return minVal;
+	return Math.min(...arr);
 }
```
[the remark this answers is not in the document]
kirchet: However if there are too many items in the provided array we cannot use `Math.min(...arr)` because it fails in runtime

src/helpers/algorithms.ts
```diff
@@ -1,51 +1,23 @@
...
-
-export function upperbound<TArrayElementType, TValueType>(
+export function boundCompare<TArrayElementType, TValueType>(
 	arr: readonly TArrayElementType[],
 	value: TValueType,
-	compare: UpperBoundComparatorType<TValueType, TArrayElementType>,
+	compare: BoundComparatorType<TArrayElementType, TValueType>,
+	lower: boolean,
 	start: number = 0,
 	to: number = arr.length): number {
```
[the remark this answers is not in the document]
SlicedSilver: Even better 👍

src/helpers/mathex.ts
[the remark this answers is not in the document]
SlicedSilver: 🤔
So this change would limit our `min` function to arrays of length of `65535` or less. The code to detect this and split would probably end up being larger and slower than the 'old' function. So better to remove this change.

src/helpers/algorithms.ts
```diff
@@ -1,22 +1,23 @@
...
-	compare: LowerBoundComparatorType<TArrayElementType, TValueType>,
+	compare: BoundComparatorType<TArrayElementType, TValueType>,
 	start: number = 0,
 	to: number = arr.length): number {
 	let count: number = to - start;
 	while (0 < count) {
 		const count2: number = (count >> 1);
 		const mid: number = start + count2;
-		if (compare(arr[mid], value)) {
+		if (Boolean(compare(arr[mid], value)) === lower) {
```
eugene-korobko, line 20: Why do we need casting to Boolean here?

src/helpers/color.ts
```diff
@@ -238,23 +81,51 @@ const rgbRe = /^rgb\(\s*(-?\d{1,10})\s*,\s*(-?\d{1,10})\s*,\s*(-?\d{1,10})\s*\)$
...
+			cssColorResults[name] = rgb;
+			return rgb;
+		}
+		return undefined;
+	} finally {
+		document.body.removeChild(div);
+	}
+}
+
+const cssColorResults: Record<string, Rgba> = {};
```
eugene-korobko, line 104: Lets move it above the first usage?

src/helpers/color.ts
```diff
@@ -238,23 +81,51 @@ const rgbRe = /^rgb\(\s*(-?\d{1,10})\s*,\s*(-?\d{1,10})\s*,\s*(-?\d{1,10})\s*\)$
  */
 const rgbaRe = /^rgba\(\s*(-?\d{1,10})\s*,\s*(-?\d{1,10})\s*,\s*(-?\d{1,10})\s*,\s*(-?[\d]{0,10}(?:\.\d+)?)\s*\)$/;
 
-function colorStringToRgba(colorString: string): Rgba {
-	colorString = colorString.toLowerCase();
+function getColorRgbFromBrowser(name: string): Rgba | undefined {
+	if (name === 'black') {
+		return [0 as RedComponent, 0 as GreenComponent, 0 as BlueComponent, 1 as AlphaComponent];
+	}
+	const div = document.createElement('div');
```
eugene-korobko, line 88: I'm confused with performance of this approach. Adding and removing elements from DOM could be expensive.
Is there any negative side effects in case of having the div alwys mounted? Maybe we could mount it inside the chart?

SlicedSilver: The performance is slower than having an object containing all the css color names, however having all the css color names is increasing the bundle size quite a bit.
If the user is concerned about performance then they could rather specify colors using rgb strings.
Since creating the DOM element is expensive, we memo the results so we only need to do it once per color name. It is highly unlikely that a chart would be using more than a handful of color names.

I think the benefit of a smaller bundle outweighs the cost of creating a dom element in rare cases where users are specifying colors using names.

I can move the mounting of this element inside the chart's div element 👍

timocov: Does inserting into body and removing it and computing styles work well with Web Components? Also, wouldn't getting a computed styles force to do a force reflow/layout? https://gist.github.com/paulirish/5d52fb081b3570c81e3a#calling-getcomputedstyle

src/helpers/color.ts
```diff
@@ -238,23 +81,55 @@ const rgbRe = /^rgb\(\s*(-?\d{1,10})\s*,\s*(-?\d{1,10})\s*,\s*(-?\d{1,10})\s*\)$
...
+	try {
+		div.style.color = name;
+		// use absolute to ensure that it doesn't reflow the rest of the page layout
+		div.style.position = 'absolute';
+		div.style.left = '-9999px';
+		document.body.appendChild(div);
+		const rgbString = window.getComputedStyle(div).color;
```
timocov, line 97: IIRC there are cases when `window.getComputedStyle` can return `null` (e.g. when an element is being loaded inside an iframe - see https://github.com/souporserious/get-node-dimensions/issues/8 and https://github.com/souporserious/react-measure/issues/87)

SlicedSilver: It appears to work correctly within iFrames. We aren't trying to get dimensions which could be the difference.

SlicedSilver:
It seems that we would prefer to keep the css color names list in the source code instead of switching to using `getComputedStyle`.

What are the thoughts of maybe doing one of the following instead (BREAKING CHANGES):
- 🚀 Move the color name parsing into an exported function which the user can use to convert names to rgb strings. So that only the users which need it will actually import it. It could be tree-shaked for the majority of users.
- 😕 Remove it from the library completely. Much like we leave it to the user to handle timezones, we can leave it to the user to manage css color names if they want to use them.
- 👀 Use `getComputedStyle` but add a message to the documentation that there is a slight performance cost for this.
- ❤️ Leave it as it was. It's working, and the cost is 1.1KB (2.5% of bundle size) is okay.

Emojis are for voting via the GH reactions.

timocov:
> Move the color name parsing into an exported function which the user can use to convert names to rgb strings. So that only the users which need it will actually import it. It could be tree-shaked for the majority of users.

I'd probably suggest slightly different (and extreme) variation of this idea, but that would also give another nice feature out of the box. Instead of providing a function that parses color (a general problem with functions in options API is that the library doesn't work when a function should behave differently thus everything needs to be invalidated - which can be avoided by forcing a user to provide another reference of course) you can provide a new option at the root level _of the chart_ called something like `palette`, `colors` or `colorsMap` or so that is basically a color-mapping object from a name to a rgb(a) representation of a color. This mapping should be used everywhere _before_ actually treating a value as a color i.e. somewhere close to renderers or pane views (to achieve smooth color changing without a hustle). And the library can provide the default "browser-based" object that can be used by users but if not then will be tree-shaked.

With this feature users don't need to worry about changing colors of all previously created objects or even some particular objects if they want/need to maintain their color schemes - they can create objects with some predefined color names and then change them in one go by providing another mapping. For example, you can do something like this:

```js
const palette = {
    '--tv-color-black': '#000000',
};

const chart = createChart(document.body, { palette });
const series = chart.addLineSeries({ color: '--tv-color-line-chart-color' });
```

and then later, once a user changed the theme of the page, you can simply do:

```js
const newPalette = {
    '--tv-color-black': '#ffffff',
};

chart.applyOptions({ palette: newPalette });
```

and the chart will adapt all colors automatically (in this case - a color of the previously created line series), so you don't need to traverse all previously created objects to change their colors (and it will solve a problem of the documentation website when you change a theme it refreshed chart's state 😉).

Issues that I see with this approach:
- probably requires some color over where colors can be parsed and exposed from API - it _might_ require some work to achieve "smooth" update with a single "light invalidation"
- a demand of themed-charts feature is questionable (but considering how it works on tradingview.com probably there are use cases for it)

SlicedSilver:
Reverted the changes which removed the css color names from the source code.

0220tiezo1323 (changes requested):
Yes

edew (dismissed):

edew (approved):

PR #1379, opened 2023-06-17
fix candlestick width at smaller bar spacing

SlicedSilver:

**PR checklist:**

- `N/A` Addresses an existing issue
- [x] Includes tests
- `N/A` Documentation update

**Overview of change:**
~Changes the calculation for the width of candlesticks at smaller bar spacing values.~
Adds a test case for the candlestick widths at smaller bar spacing values.

Decided that the existing behaviour is better for chart readability (instead of preventing slight overlapping). So the changes to the width calculation have been reverted.

---
 
~**Problem:**
Previously, there were cases where the candlestick width would be larger than the bar spacing therefore candlesticks could overlap. Preview of issue: https://codesandbox.io/p/sandbox/lightweight-charts-js-starter-forked-fnycf7?file=%2Fsrc%2Fmain.js%3A19%2C1
dpr-1~

~**Solution:**
Adjusted the `optimalCandlestickWidth` function such that the calculated width (in bitmap pixels) should never exceed the barSpacing gap (in bitmap pixels) until the width reaches a single pixel.~

~Plot of the new function results: https://jsfiddle.net/msilverwood/jx0zLo6f/~
```
Blue - candle width (pixels)
Red - bar spacing (pixels)
Green - bar spacing (media size)
```

### DIFF
--- a/src/renderers/optimal-bar-width.ts
+++ b/src/renderers/optimal-bar-width.ts
@@ -3,9 +3,11 @@ export function optimalBarWidth(barSpacing: number, pixelRatio: number): number 
 }
 
 export function optimalCandlestickWidth(barSpacing: number, pixelRatio: number): number {
+	const barSpacingSpecialCaseFrom = 2.5;
 	const barSpacingSpecialCaseTo = 4;
-	if (barSpacing < barSpacingSpecialCaseTo) {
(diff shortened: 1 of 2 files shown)


SlicedSilver:
Change not required.

PR #1388, opened 2023-06-29
Fix percentage mode and zero first value

### LINKED ISSUE #1386: Percentage mode & zero first value
**Lightweight Charts™ Version:** 4.0.1

**Steps/code to reproduce:**

* Create several line series on one price scale
* The first value of one of the series should be `0`
* Enable percentage mode

```js
import { createChart, PriceScaleMode } from "lightweight-charts";

const chartOptions = {
  height: 200,
  rightPriceScale: {
    borderVisible: false,
    mode: PriceScaleMode.Percentage
  },
  layout: { textColor: "black", background: { type: "solid", color: "white" } }
};
const chart = createChart(document.getElementById("container"), chartOptions);
const series1 = chart.addLineSeries({
  color: "#2962FF"
});
series1.setData([
  { time: 1522033200, value: 0 },
  { time: 1529895600, value: -3 },
  { time: 1537758000, value: 3 }
]);
const series2 = chart.addLineSeries({
  color: "#FF2962"
});
series2.setData([
  { time: 1522033200, value: 1 },
  { time: 1529895600, value: -1 },
  { time: 1537758000, value: 2 }
]);
```

**Actual behavior:**

No series painted

**Expected behavior:**

The only one series should be hidden (that starts from the zero value)

**Screenshots:**

nothing painted:
image

https://codesandbox.io/s/percentage-mode-forked-l6d855?file=/src/index.js

SlicedSilver:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #1386
- [x] Includes tests

**Overview of change:**

When merging price ranges for the price scale, infinite values would always be used if one of the values was infinite (which is the case with percentage mode and zero first value). The code change prevents Infinite values from always winning, but fallback to Infinite if required.

### DIFF
--- a/src/model/price-range-impl.ts
+++ b/src/model/price-range-impl.ts
@@ -2,6 +2,15 @@ import { isNumber } from '../helpers/strict-type-checks';
 
 import { PriceRange } from './series-options';
 
+function computeFiniteResult(
+	values: number[],
+	method: (...values: number[]) => number,
+	fallback: number
+): number {
+	const result = method(...values.filter(Number.isFinite));
+	return Number.isFinite(result) ? result : fallback;
+}
+
 export class PriceRangeImpl {
 	private _minValue: number;
 	private _maxValue!: number;
@@ -43,8 +52,8 @@ export class PriceRangeImpl {
 			return this;
 		}
 		return new PriceRangeImpl(
-			Math.min(this.minValue(), anotherRange.minValue()),
-			Math.max(this.maxValue(), anotherRange.maxValue())
+			computeFiniteResult([this.minValue(), anotherRange.minValue()], Math.min, -Infinity),
+			computeFiniteResult([this.maxValue(), anotherRange.maxValue()], Math.max, Infinity)
 		);
 	}
 
--- a/src/model/price-range-impl.ts
+++ b/src/model/price-range-impl.ts
@@ -2,13 +2,22 @@ import { isNumber } from '../helpers/strict-type-checks';
 
 import { PriceRange } from './series-options';
 
+function ensureFiniteWithFallback(value: number, fallback: number): number {
+	return Number.isFinite(value) ? value : fallback;
+}
+
 function computeFiniteResult(
-	values: number[],
 	method: (...values: number[]) => number,
+	valueOne: number,
+	valueTwo: number,
 	fallback: number
 ): number {
-	const result = method(...values.filter(Number.isFinite));
-	return Number.isFinite(result) ? result : fallback;
+	const firstFinite = Number.isFinite(valueOne);
+	return firstFinite && Number.isFinite(valueTwo)
+		? ensureFiniteWithFallback(method(valueOne, valueTwo), fallback)
+		: firstFinite
+		? valueOne
+		: valueTwo;
 }
 
 export class PriceRangeImpl {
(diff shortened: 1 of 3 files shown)


src/model/price-range-impl.ts
```diff
@@ -2,6 +2,24 @@ import { isNumber } from '../helpers/strict-type-checks';
...
+	valueOne: number,
+	valueTwo: number,
+	fallback: number
+): number {
+	const firstFinite = Number.isFinite(valueOne);
+	return firstFinite && Number.isFinite(valueTwo)
+		? method(valueOne, valueTwo)
+		: firstFinite
+		? valueOne
+		: ensureFiniteWithFallback(valueTwo, fallback);
```
kirchet, line 20:
```suggestion
	return method(
		Number.isFinite(valueOne) ? valueOne : fallback,
		Number.isFinite(valueTwo) ? valueTwo : fallback
	);
```

src/model/price-range-impl.ts
```diff
@@ -2,6 +2,24 @@ import { isNumber } from '../helpers/strict-type-checks';
...
+}
```
kirchet, line 21:
```suggestion
function computeFiniteResult(
	method: (...values: number[]) => number,
	valueOne: number,
	valueTwo: number,
	fallback: number
): number {
	return method(
		ensureFiniteWithFallback(valueOne, fallback),
		ensureFiniteWithFallback(valueTwo, fallback)
	);
}
```

SlicedSilver: We would like the fallback to be possibly an Infinite value so this wouldn't work.

We expect that this: `computeFiniteResult(Math.min, -Infinity, 55, -Infinity)` should return `55`. With this change it would return `-Infinity`.

Essentially the fallback should be used if BOTH values are Infinite.

Where we are using the code, we would like it to be Infinite if both values are infinite, but finite if either value (or both) are finite.

I suspect that the function (and the variables) could be named better.

Maybe: `fallback` → `fallbackIfBothNonFinite`

kirchet: Then probably we need to filter out infinite values first

```ts
function finiteValuesOrNull(...values: number[]): numer[] | null {
  const result = values.filter(Numer.isFinite);
  return result.length ? result : null;
}
...
return new PriceRangeImpl(
  Math.min(...(finiteValuesOrNull(this.minValue(), anotherRange.minValue()) || [-Infinity])),
  Math.max(...(finiteValuesOrNull(this.maxValue(), anotherRange.maxValue()) || [Infinity]))
);
```

SlicedSilver: My original commit used `filter` but I changed it to the current version because I wanted to avoid using arrays so that the function is a bit faster (avoid creating and spreading array if possible).

The original commit had:
```typescript
function computeFiniteResult(
 	values: number[],
	method: (...values: number[]) => number,
 	fallback: number
 ): number {
 	const result = method(...values.filter(Number.isFinite));
 	return Number.isFinite(result) ? result : fallback;
 }
```

This function will be called on every scroll / scale / etc... so I tried to make it a bit faster (didn't actually benchmark the difference).

kirchet: OK, anyway I think we should get rid of `ensureFiniteWithFallback`

```suggestion
function computeFiniteResult(
	method: (...values: number[]) => number,
	valueOne: number,
	valueTwo: number,
	fallback: number
): number {
	const firstFinite = Number.isFinite(valueOne);
	const secondFinite = Number.isFinite(valueTwo);
	
	if (firstFinite && secondFinite) {
		return method(valueOne, valueTwo);
	}
	
	return !firstFinite && !secondFinite ? fallback : (firstFinite ? valueOne : valueTwo);
}
```

kirchet (approved):

bitcoinfinancier (approved):

PR #1399, opened 2023-08-01
Allow tick mark label length to be overridden

### LINKED ISSUE #1396: Timescale tickmarks overlapping

**Steps/code to reproduce:**

Define custom tickMarkFormatter that returns long strings

**Actual behavior:**

Labels on time scale overlap each other

**Expected behavior:**

Labels sould be sparsed to avoid overlapping

**Screenshots:**

overlapping

https://jsfiddle.net/ekorobko/12kn75tb/2/

edew:
**Type of PR:** 

**PR checklist:**

- [x] Addresses an existing issue: fixes #1396 
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Allow tick mark label length to be overridden.

By default the library assumes that each tick mark labels has a length of approximately 8 characters.

This PR adds the `tickMarkMaxCharacterLength` option for the time scale that can override the assumed length of the tick mark labels.


tests/e2e/graphics/test-cases/time-scale/do-not-render-overlapping-marks.js
```diff
@@ -0,0 +1,313 @@
+function runTestCase(container) {
+	const chart = window.chart = LightweightCharts.createChart(container, {
+		handleScroll: false,
+		handleScale: false,
+		tickMarkFormatter: time => `long prefix ${time}`,
```
SlicedSilver, line 5:
```suggestion
		timeScale: { tickMarkFormatter: time => `long prefix ${time}` },
```

src/gui/time-axis-widget.ts
```diff
@@ -420,19 +421,37 @@ export class TimeAxisWidget implements MouseEventHandlers, IDestroyable {
...
-					ctx.fillText(tickMark.label, coordinate, yText);
+					drawTickMark(tickMark, 2);
+				}
+			}
+
+			// draw base marks
+			ctx.font = this._baseFont();
+			for (const tickMark of tickMarks) {
+				if (tickMark.weight < maxWeight) {
+					drawTickMark(tickMark, 0.5);
```
SlicedSilver, line 454: Could we increase the tolerance a bit? 
Screenshot 2023-08-01 at 17 04 31

src/gui/time-axis-widget.ts
```diff
@@ -420,19 +421,37 @@ export class TimeAxisWidget implements MouseEventHandlers, IDestroyable {
...
+				const coordinate = tickMark.needAlignCoordinate ? this._alignTickMarkLabelCoordinate(ctx, tickMark.coord, tickMark.label) : tickMark.coord;
+				const labelWidth = this._widthCache.measureText(ctx, tickMark.label);
+				const labelWidthHalf = labelWidth / 2;
+				const leftTextCoordinate = Math.floor(coordinate - labelWidthHalf) + tolerance;
+				const rightTextCoordinate = Math.ceil(coordinate + labelWidthHalf) + tolerance;
+				const range = new RangeImpl(leftTextCoordinate, rightTextCoordinate);
+				const rangeOverlaps = ranges.some((x: RangeImpl<number>) => x.contains(range.left()) || x.contains(range.right()));
+
+				if (rangeOverlaps) {
+					return;
```
kirchet, line 436: Not sure that we should check if the tick marks are overlapped at the painting stage. I think we should somehow change tick marks calculation algorithm

https://github.com/tradingview/lightweight-charts/blob/981ca1b5f4dc5eb2edf80883ccd578f7a05521d3/src/model/time-scale.ts#L488-L496

Probably we should provide the current font size instead of `maxLabelWidth`

Also we should ensure that visible marks are consistent when we scroll the chart. The same marks should be visible if we scroll the chart a little bit and the longest/shortest mark becomes outside of the visible range

edew:
This file can be used for debugging or testing [debug.html.zip](https://github.com/tradingview/lightweight-charts/files/12261688/debug.html.zip)

src/model/layout-options.ts
```diff
@@ -67,7 +67,7 @@ export interface LayoutOptions {
 	/**
 	 * Font size of text on scales in pixels.
 	 *
-	 * @defaultValue `11`
```
edew, line 70: Updated as the actual default value is 12.

src/model/time-scale.ts
```diff
@@ -993,6 +997,24 @@ export class TimeScale {
 
 		this._correctBarSpacing();
 	}
+
+	private _updateLabelLengthEstimate(): void {
```
edew, line 1001: Added to estimate the label length by calling the custom formatter (if defined). It assumes that the custom formatter will return the same length for all labels.

SlicedSilver: For the 'options' chart example, we can't make this assumption (that the length will be consistent) because the horizontal scale can be prices instead of dates.
So the returned string for `0` might be `'$0.0'` but if the chart also includes larger prices then the returned string would get larger (`123456789` → `'$123 456 789.00'`).

We would need to test it on an example like `tests/e2e/graphics/test-cases/horizontal-price-scale.js` on the https://github.com/tradingview/lightweight-charts/pull/1323 PR (and set the numbers so that large price values are generated)

tests/e2e/graphics/test-cases/time-scale/do-not-render-overlapping-marks-font-size-24.js
```diff
@@ -0,0 +1,319 @@
...
+	const chart = window.chart = LightweightCharts.createChart(container, {
+		handleScroll: false,
+		handleScale: false,
+		layout: {
+			fontSize: 24,
+		},
+		timeScale: {
+			tickMarkMaxCharacterLength: 'long prefix 0000-00-00'.length,
+			tickMarkFormatter: time => `long prefix ${time}`,
+		},
```
SlicedSilver, line 11: Could we add the following option so that the test case doesn't have a crosshair label (and makes it easier to see the timescale tick marks)?

```js
crosshair: { vertLine: { labelVisible: false } }
```

and the same for the other test case.

SlicedSilver (approved):
LGTM. 👍

This resolves the issue while maintaining the current behavior, unless the client opts for a change.

It is a quick and straightforward solution. ✅
We can revisit the idea of implementing a more advanced approach in the future, but only if there is a genuine need for variable tick mark widths. At present, it is not worthwhile considering the associated effort and performance impact.

tests/e2e/graphics/test-cases/time-scale/do-not-render-overlapping-marks-font-size-24.js
edew: Good idea, done

SlicedSilver (approved):

kirchet (dismissed):

SlicedSilver (approved):

PR #1409, opened 2023-08-22
Use label length when building marks

### LINKED ISSUE #1396: Timescale tickmarks overlapping

edew:
**Type of PR:** 

**PR checklist:**

- [x] Addresses an existing issue: fixes #1396 
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Use the label length when building tick marks.

These changes stop overlapping labels, but it is a breaking change as it does not preserve the existing time scale behaviour.

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -1,24 +1,28 @@
 // eslint-env node
 
+// We should set these values as 0.01 higher than the number we get when running the size check.
+// The reason is that the size can change very slightly on each build because we include the build date as a comment in the output.
+// This seems to affect the compressed size (some dates are more compressible than others).
+
 module.exports = [
 	{
 		name: 'CJS',
 		path: 'dist/lightweight-charts.production.cjs',
-		limit: '47.68 KB',
+		limit: '47.69 KB',
 	},
 	{
 		name: 'ESM',
 		path: 'dist/lightweight-charts.production.mjs',
-		limit: '47.63 KB',
+		limit: '47.64 KB',
 	},
 	{
 		name: 'Standalone-ESM',
 		path: 'dist/lightweight-charts.standalone.production.mjs',
-		limit: '49.36 KB',
+		limit: '49.37 KB',
 	},
 	{
 		name: 'Standalone',
 		path: 'dist/lightweight-charts.standalone.production.js',
-		limit: '49.40 KB',
+		limit: '49.41 KB',
 	},
 ];
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -4,21 +4,21 @@ module.exports = [
 	{
 		name: 'CJS',
 		path: 'dist/lightweight-charts.production.cjs',
-		limit: '47.62 KB',
+		limit: '47.68 KB',
 	},
 	{
 		name: 'ESM',
 		path: 'dist/lightweight-charts.production.mjs',
-		limit: '47.56 KB',
+		limit: '47.63 KB',
 	},
 	{
 		name: 'Standalone-ESM',
 		path: 'dist/lightweight-charts.standalone.production.mjs',
-		limit: '49.29 KB',
+		limit: '49.36 KB',
 	},
 	{
 		name: 'Standalone',
 		path: 'dist/lightweight-charts.standalone.production.js',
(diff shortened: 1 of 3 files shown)


SlicedSilver:
Whether this is a 'breaking change' is debatable. It doesn't change the API so everyone's charts should continue to work. The difference is when the time scale labels are either shorter than 8 character (then it is possible for more tick marks & grid lines to appear), or when they are longer than 8 characters.

IMO I don't think it is a true breaking change, and could be included in a minor release.

edew:
> Whether this is a 'breaking change' is debatable. It doesn't change the API so everyone's charts should continue to work. The difference is when the time scale labels are either shorter than 8 character (then it is possible for more tick marks & grid lines to appear), or when they are longer than 8 characters.
> 
> IMO I don't think it is a true breaking change, and could be included in a minor release.

Agreed, maybe breaking change is the wrong way of putting it.

If we were to merge this I don't think it needs a major release.

src/model/tick-marks.ts
```diff
@@ -16,18 +16,35 @@ export interface TickMark {
...
 	marks: readonly TickMark[];
+	defaultMaxIndexesPerMark: number;
+}
+
+export interface TickMarkBuildResult {
+	marks: readonly TickMark[];
+	maxLabelWidth: number;
+}
+
+function roundToOneDecimalPlace(n: number): number {
```
edew, line 34: Why round to one decimal place?

If we do not then even very small changes in the spacing will cause the marks to be rebuilt. Limiting this to increments of 0.1 seems to be an acceptable compromise but probably there is a cleverer way to handle this.

src/model/tick-marks.ts
```diff
@@ -51,20 +68,23 @@ export class TickMarks<HorzScaleItem> {
...
-		if (this._cache === null || this._cache.maxIndexesPerMark !== maxIndexesPerMark) {
-			this._cache = {
-				marks: this._buildMarksImpl(maxIndexesPerMark),
-				maxIndexesPerMark,
-			};
+	public build(spacing: number, fontSize: number): TickMarkBuildResult {
+		const maxLabelWidth = (fontSize + 4) * 5;
+		const widthPerCharacterEstimate = maxLabelWidth / 8;
+		// "default" because we will adjust the value based on actual label widths
+		const defaultMaxIndexesPerMark = roundToOneDecimalPlace(maxLabelWidth / spacing);
```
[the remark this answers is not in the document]
SlicedSilver:
```suggestion
		// Why round to one decimal place?
		// If we do not then even very small changes in the spacing will cause the marks to be rebuilt. Limiting this to increments of 0.1 seems to be an acceptable compromise.
		const defaultMaxIndexesPerMark = roundToOneDecimalPlace(maxLabelWidth / spacing);
```

.size-limit.js
```diff
@@ -4,21 +4,21 @@ module.exports = [
...
 		name: 'Standalone-ESM',
 		path: 'dist/lightweight-charts.standalone.production.mjs',
-		limit: '49.29 KB',
+		limit: '49.36 KB',
 	},
 	{
 		name: 'Standalone',
 		path: 'dist/lightweight-charts.standalone.production.js',
-		limit: '49.33 KB',
+		limit: '49.40 KB',
```
SlicedSilver, line 22: We should set these values as 0.01 higher than the number we get when running the size check.
The reason is that the size can change very slightly on each build because we include the build date as a comment in the output. This seems to affect the compressed size (some dates are more compressible than others).

I suggest these numbers:
47.69
47.64
49.36
49.41

and it would probably be a good idea to add a comment to the top of this files which explains this. Because it is not obvious.

SlicedSilver (approved):

0220tiezo1323 (changes requested):
Asp

Issue #1163, opened 2022-09-09
synchronizing across multiple windows with lightweight charts


peterswanson:
hello, I was directed by those at TV to send this question here. I am using the lightweight charts library to display price and volume synchronized so that you could scale the price and the volume would stay connected on the time axis. Also it would be great to get the cursor to line up on both charts at the same time like it happens in TV

![image](https://user-images.githubusercontent.com/3860609/189452228-92f62d84-fe17-4344-a73f-f26f422843e7.png)
The issue is that when I create separate charts using the library, neither of these functionalities occurs (understandably) and so I was hoping that you might know of a way to get this working?
here is how it looks now when you scale the price chart:
![image (1)](https://user-images.githubusercontent.com/3860609/189452235-3141ae2f-73b1-44e2-b45e-f6680b81ccf2.png)

(note the charts on the Y-axis aren't synchronized and the crosshair isn't showing on the volume chart. )

SlicedSilver:
We have an example showing how to configure the chart to display both the price and volume on a single chart here: https://tradingview.github.io/lightweight-charts/tutorials/how_to/price-and-volume 

You could also display the volume on a separate price scale like this: https://tradingview.github.io/lightweight-charts/tutorials/how_to/two-price-scales

**I can understand that the examples may not entirely meet your requirements. For your specific question, there isn't a way to 'link' two separate charts together at the moment.** 

You can however achieve something very similar with quite a bit of extra code. You can use the [crosshairMove](https://tradingview.github.io/lightweight-charts/docs/api/interfaces/IChartApi#subscribecrosshairmove) event to know where the cursor is within one of the charts and then determine the matching data point in the second chart and display a Priceline (https://tradingview.github.io/lightweight-charts/docs/api/interfaces/ISeriesApi#createpriceline). That would handle getting a horizontal 'crosshair' on both charts. To get the vertical crosshair, I would suggest hiding the crosshair on the charts (https://tradingview.github.io/lightweight-charts/docs/api/interfaces/CrosshairOptions) and then using a transparent canvas element placed on top of the charts (or below if the charts have a transparent background), and then draw your own vertical crosshair. 

To assist with getting the charts to have similar price scale widths you could look at these tips: https://stackoverflow.com/questions/71901106/a-way-to-set-pricescale-width-in-lightweight-charts/73209722#73209722

[...]

PR #1412, opened 2023-08-23
Add option to set minimum dimension for scales

### LINKED ISSUE #1062: A way to set PriceScale width - currently there's no way to keep two charts' PriceScale width in sync one above the other
Screenshot from 2022-04-16 17-58-43

I have two charts (one for price, one for volume). I keep them in sync by usingthe various API observer functions. Works wonders.

Except the price scale width - I can get it using the priceScale('right').width() call, but I don't see any way to set it.
I'd like to be able to keep the width of both chart's price scales in sync (and keep them set to the max width of the two).

Thanks!

SlicedSilver:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #1062 (related: #1163, #50)
- [x] Includes tests
- [x] Documentation update

**Overview of change:**
Adds options to set minimum dimensions for the price and time scales.

Setting a minimum height / width is useful for ensuring that multiple charts positioned in a vertical / horizontal stack each have an identical price scale width (or time scale height), or for plugins which require a bit more space within the scale pane.

The minimum dimension is only considered when that scale should be visible, and also it can be exceeded if the scale needs more space.

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -4,21 +4,21 @@ module.exports = [
 	{
 		name: 'CJS',
 		path: 'dist/lightweight-charts.production.cjs',
-		limit: '47.62 KB',
+		limit: '47.66 KB',
 	},
 	{
 		name: 'ESM',
 		path: 'dist/lightweight-charts.production.mjs',
-		limit: '47.56 KB',
+		limit: '47.60 KB',
 	},
 	{
 		name: 'Standalone-ESM',
 		path: 'dist/lightweight-charts.standalone.production.mjs',
-		limit: '49.29 KB',
+		limit: '49.32 KB',
 	},
 	{
 		name: 'Standalone',
 		path: 'dist/lightweight-charts.standalone.production.js',
-		limit: '49.33 KB',
+		limit: '49.37 KB',
 	},
 ];
--- a/src/api/options/price-scale-options-defaults.ts
+++ b/src/api/options/price-scale-options-defaults.ts
@@ -14,4 +14,5 @@ export const priceScaleOptionsDefaults: PriceScaleOptions = {
 		bottom: 0.1,
 		top: 0.2,
 	},
+	minimumWidth: 0,
 };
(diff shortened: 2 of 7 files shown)


0220tiezo1323 (changes requested):
Pay changes need

edew (approved):

0220tiezo1323 (approved):

edew:
The d.ts changes are expected

```
diff --git a/./dts-changes/base.d.ts b/./dts-changes/compare.d.ts
index 5670b54c..ac83c537 100644
--- a/./dts-changes/base.d.ts
+++ b/./dts-changes/compare.d.ts
@@ -1147,6 +1147,19 @@ export interface HorzScaleOptions {
         * With this flag equal to `true`, marks of the same weight are either all drawn or none are drawn at all.
         */
        uniformDistribution: boolean;
+       /**
+        * Define a minimum height for the time scale.
+        * Note: This value will be exceeded if the
+        * time scale needs more space to display it's contents.
+        *
+        * Setting a minimum height could be useful for ensuring that
+        * multiple charts positioned in a horizontal stack each have
+        * an identical time scale height, or for plugins which
+        * require a bit more space within the time scale pane.
+        *
+        * @defaultValue 0
+        */
+       minimumHeight: number;
 }
 /**
  * The main interface of a single chart using time for horizontal scale.
@@ -2720,6 +2733,19 @@ export interface PriceScaleOptions {
         * @defaultValue `false`
         */
        ticksVisible: boolean;
+       /**
+        * Define a minimum width for the price scale.
+        * Note: This value will be exceeded if the
+        * price scale needs more space to display it's contents.
+        *
+        * Setting a minimum width could be useful for ensuring that
+        * multiple charts positioned in a vertical stack each have
+        * an identical price scale width, or for plugins which
+        * require a bit more space within the price scale pane.
+        *
+        * @defaultValue 0
+        */
+       minimumWidth: number;
```

PR #1431, opened 2023-09-28
don't shift chart when new data replaces existing whitespace

### LINKED ISSUE #1201: Chart with whitespaces shifts on realtime

**Steps/code to reproduce:**

Open attached sample, wait for 6 seconds

**Actual behavior:**

red series shifts to the left

**Expected behavior:**

res series should keep its place

Additionally:

if replace whitespaces with some values

SlicedSilver:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #1201
- [x] Includes tests
- [x] Documentation update

**Overview of change:**
When a realtime update is for a time point which already exists on the time scale then we won't shift the visible range of the chart anymore.

**Pipeline Checks:**
- Changes to DTS are expected. ✅
- Changes to Graphics Tests are expected (new test `do-not-shift-range-when-replacing-whitespace` was added). ✅
- It is expected that the `histogram-add-new-color-on-update` and `whitespace-updates` tests will fail. They have been manually checked and the old behaviour was incorrect. ✅

### DIFF
--- a/src/model/time-scale.ts
+++ b/src/model/time-scale.ts
@@ -147,12 +147,21 @@ export interface HorzScaleOptions {
 	/**
 	 * Shift the visible range to the right (into the future) by the number of new bars when new data is added.
 	 *
-	 * Note that this only applies when the last bar is visible, and the new data point is not being placed
-	 * at the same time value as an existing whitespace data point (on a different series).
+	 * Note that this only applies when the last bar is visible.
 	 *
 	 * @defaultValue `true`
 	 */
 	shiftVisibleRangeOnNewBar: boolean;
+
+	/**
+	 * Don't shift the visible range to the right (into the future) by the number of new bars when new data is added IF
+	 * that data is placed on existing whitespace data points (times).
+	 *
+	 * {@link shiftVisibleRangeOnNewBar} needs to be enabled for this to have any effect.
+	 *
+	 * @defaultValue `false`
+	 */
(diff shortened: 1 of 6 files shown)


src/model/time-scale.ts
```diff
@@ -153,6 +153,16 @@ export interface HorzScaleOptions {
...
 
+	/**
+	 * Should shift the visible range to the right (into the future) by the number of new bars when new data is added IF
+	 * that data is placed on existing whitespace data points (times).
+	 *
+	 * {@link shiftVisibleRangeOnNewBar} needs to be enabled for this to have any effect.
+	 *
+	 * @defaultValue `true`
+	 */
+	shiftVisibleRangeWhenNewBarReplacesWhitespace: boolean;
```
eugene-korobko, line 164: Do we really need this option? It is difficult to imagine why somebody needs this shifting

eugene-korobko: Well, from my point of view, old behavior is an obvious bug and we needn't keep back compatibility.
Formally, all bug fixes are "breaking changes" since they change the behavior, however we do not provide options to keep old behavior.
So, I'd like to select case #2.

What do you think?

SlicedSilver: Sounds good 👍

eugene-korobko (approved):

PR #1445, opened 2023-10-18
disabling touch scrolling should apply to scales as well

### LINKED ISSUE #1441: On iOS page is not scrollable if handleScroll is false and doing scroll action on Y axis
**Lightweight Charts™ Version:** 4.1.0

**Steps/code to reproduce:**

1. Open JSFiddle link on iOS touch device
2. Scroll down to the chart
3. Try to scroll down more if you are doing the scroll action on Y axis

**Actual behavior:**

Nothing happens, page is not scrollable if you try to perform scroll action on Y axis.

**Expected behavior:**

Page should be scrollable, since we disabled chart scroll handling.

**JSFiddle:** https://jsfiddle.net/ceyzgab2/2/

SlicedSilver:

**PR checklist:**

- [x] Addresses an existing issue: fixes #1441
- [x] Includes tests
- `N/A?` ~Documentation update~

**Overview of change:**
If the `vertTouchDrag` option for `handleScroll` is set to `false` then touch scrolling over the price scale will now allow the page to scroll as expected.

If the `horzTouchDrag` option for `handleScroll` is set to `false` then touch scrolling over the time scale will now allow the page to scroll as expected.

- Docs: https://tradingview.github.io/lightweight-charts/docs/api/interfaces/HandleScrollOptions

### DIFF
--- a/src/gui/price-axis-widget.ts
+++ b/src/gui/price-axis-widget.ts
@@ -145,7 +145,7 @@ export class PriceAxisWidget implements IDestroyable {
 			this._topCanvasBinding.canvasElement,
 			handler,
 			{
-				treatVertTouchDragAsPageScroll: () => false,
+				treatVertTouchDragAsPageScroll: () => !this._options.handleScroll.vertTouchDrag,
 				treatHorzTouchDragAsPageScroll: () => true,
 			}
 		);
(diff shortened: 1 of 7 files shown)


SlicedSilver:
FYI: even though new tests were added, it is expected that the pipeline shows no errors. This is because the tests were added to the interactions tests which aren't comparing old and new behaviour versus the golden version, but rather only on the new version.
I've tested these tests on the golden version and the 2 which are expected to fail do fail 👍

PR #1492, opened 2024-01-07
plugin examples: fix plugin base _fireDataUpdated #1491

### LINKED ISSUE #1491: Plugin not responding to data update
**Lightweight Charts™ Version:** 4.1.1

**Steps/code to reproduce:**

Have not tried this myself on your example, but you should be able to reproduce the issue with the change described here.

In the [session highlight example](https://github.com/tradingview/lightweight-charts/blob/86da4820bc2d81126122496ae6c7bca1b98b216a/plugin-examples/src/plugins/session-highlighting/example/example.ts), move `candleSeries.setData(data);` to the end of the file, after `candleSeries.attachPrimitive(sessionHighlighting);`.

```ts
// ...
const candleSeries = chart.addCandlestickSeries();
const data = generateAlternativeCandleData();
// make sure this is commented out or removed to be able to reproduce the problem
// candleSeries.setData(data);

function getDate(time: Time): Date {
// ...
// ...
const sessionHighlighting = new SessionHighlighting(sessionHighlighter);
candleSeries.attachPrimitive(sessionHighlighting);
candleSeries.setData(data);
```

**Actual behavior:**

Session highlighting does not work in the above case - when `attachPrimitive` is called before `setData`.

**Expected behavior:**

It should work :). Plugin should respond to data update.

**Potential solution:**

The issue seems to be caused by this line:

https://github.com/tradingview/lightweight-charts/blob/61d45db057c36c1b0d95e9a0d909fb5a60e5a19f/plugin-examples/src/plugins/plugin-base.ts#L25C5-L25C5

`_fireDataUpdated` is using `this` inside the function body:

```ts
	private _fireDataUpdated(scope: DataChangedScope) {
		if (this.dataUpdated) {
			this.dataUpdated(scope);
		}
	}
```

This `this` is not bound to a `SessionHighlighting` object (which inherits from `PluginBase`), but to something else. That other thing does not have `this.dataUpdated`, and it will always be `undefined`, whether `SessionHiglighting` object defines `dataUpdated` method or not.

Line 25 in `plugin-base.ts` should probably be changed:

```ts
// old
this._series.subscribeDataChanged(this._fireDataUpdated);

// fixed (option 1)
this._series.subscribeDataChanged((scope) => this._fireDataUpdated(scope));

// fixed(option 2)
this._series.subscribeDataChanged(this._fireDataUpdated.bind(this));
```

After the change, `this` inside `_fireDataUpdated` should be properly bound to the `SessionHighlighting` instance.

SlicedSilver:
**Type of PR:** Plugin examples bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #1491

**Overview of change:**

`_fireDataUpdated` within `plugin-examples/src/plugins/plugin-base.ts` wasn't correctly bound to the instance, and thus the data update notifications weren't being correctly sent to the plugin instance.

The session highlighting example has been updated to include simulated streaming data as a test case.

### DIFF
--- a/plugin-examples/src/plugins/plugin-base.ts
+++ b/plugin-examples/src/plugins/plugin-base.ts
@@ -28,6 +28,7 @@ export abstract class PluginBase implements ISeriesPrimitive<Time> {
 	}
 
 	public detached() {
+		if (this._series) this._series.unsubscribeDataChanged(this._fireDataUpdated);
 		this._chart = undefined;
 		this._series = undefined;
 		this._requestUpdate = undefined;
@@ -41,7 +42,10 @@ export abstract class PluginBase implements ISeriesPrimitive<Time> {
 		return ensureDefined(this._series);
 	}
 
-	private _fireDataUpdated(scope: DataChangedScope) {
+	// This method is a class property to maintain the
+	// lexical 'this' scope (due to the use of the arrow function)
+	// and to ensure its reference stays the same, so we can unsubscribe later.
+	private _fireDataUpdated = (scope: DataChangedScope) => {
 		if (this.dataUpdated) {
 			this.dataUpdated(scope);
 		}
--- a/plugin-examples/src/plugins/plugin-base.ts
+++ b/plugin-examples/src/plugins/plugin-base.ts
@@ -28,7 +28,7 @@ export abstract class PluginBase implements ISeriesPrimitive<Time> {
 	}
 
 	public detached() {
-		if (this._series) this._series.unsubscribeDataChanged(this._fireDataUpdated);
+		this._series?.unsubscribeDataChanged(this._fireDataUpdated);
 		this._chart = undefined;
 		this._series = undefined;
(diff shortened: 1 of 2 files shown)


SlicedSilver:
On second thought, it is a bit strange to add simulated real-time updates to the session highlighting example since it is showing daily bars. So probably better to remove the changes made to `plugin-examples/src/plugins/session-highlighting/example/example.ts`.

plugin-examples/src/plugins/plugin-base.ts
```diff
@@ -28,6 +28,7 @@ export abstract class PluginBase implements ISeriesPrimitive<Time> {
 	}
 
 	public detached() {
+		if (this._series) this._series.unsubscribeDataChanged(this._fireDataUpdated);
```
romfrancois, line 31: Nit pick: could you please wrap this around brackets?

romfrancois (approved):

PR #1494, opened 2024-01-12
Enable device pixel content box support

SlicedSilver:
**Type of PR:** enhancement

**Overview of change:**
We have been blocked from adding the Device Pixel Content Box (DPCB) support (provided by the Resize Observer) to the library because viewport emulation (provided by the browsers (chromium / firefox) to tools like puppeteer) doesn't report the correct measurements adjusted by the deviceScaleFactor. In simpler terms, when we are running graphics tests at different device pixel ratios (DPR), the browser was reporting the measurements at the machines native DPR instead of the emulated DPR. Since we were unable to accurately run our graphics e2e tests we decided to disable this feature and wait for the browsers to fix the issue.

Instead of continuing to wait for the browsers to fix the viewport emulation, we can instead polyfill the Resize Observer when running our tests such that emulated DPR is used for the DPCB measurements. This PR implements the `@juggle/ResizeObserver` polyfill so that our tests can run with DPCB support enabled within the library. This allows LWC to enable DPCB support.

DPCB support will enhance the library by improving the rendering of the charts in cases where the browser has `privacy.resistFingerprinting` enabled, or when the viewport is scaled by browser zooming.

You can learn more about the advantages here: [Pixel-perfect rendering with devicePixelContentBox  |  Articles  |  web.dev](https://web.dev/articles/device-pixel-content-box)

The way the polyfill is loaded and inserted into the test case is not the most elegant implementation. Suggestions are welcomed.


SlicedSilver:
There is a single graphics test which is failing:
- make-series-hidden

The gridlines are being drawn in the wrong places for all the DPRs except `1`.
Golden:
make-series-hidden-golden

Test:
make-series-hidden-test

Issue #1438, opened 2023-10-09
Bold timescale values are not always correct and there is no option to turn this functionality off


humpedli:
**Lightweight Charts™ Version:** 4.1.0

Automatic bold texts are sometimes not working very well on timescale (x axis). If this feature could be disabled, it would be nice.

Good example (month-day):

<img width="1252" alt="month" src="https://github.com/tradingview/lightweight-charts/assets/8111798/37594a6b-0ea0-4188-9f97-d9e2411b90e7">

Bugs (year-month):

<img width="1256" alt="year" src="https://github.com/tradingview/lightweight-charts/assets/8111798/9333145b-8f05-4d73-a937-f4fe7f17ed06">
<img width="1264" alt="three_years" src="https://github.com/tradingview/lightweight-charts/assets/8111798/12a0fdf6-7358-48ea-81aa-6f1758a4d108">

We are in GMT+2 timezone, so I am converting datetimes to UTC to pass to the chart, later I use timeFormatter to display them in correct time zone. If I'm not doing this I also have issues with bold texts (HH:mm - 00:00 - not bolded vs 02:00 - bolded). In bugs section on second screenshot you can see that first 2 bold texts are correct, the last one is not. Data is the same for all 3 texts: HH:mm = 00:00.

The best would be an option to turn this functionality off, because it's not always needed.

I saw that you closed this request, but it would be a good feature to disable bold texts to ignore unwanted bugs: https://github.com/tradingview/lightweight-charts/issues/545

SlicedSilver:
Lets break this into 2 separate topics:
1. Ability to disable bold text on the timescale.
2. Bug: Incorrect tick mark has been made bold.

---

### 1. Option to disable bold text for tick marks in the timescale 

This seems like a reasonable request.

---

### 2. Incorrect tick mark has been made bold

It would seem that the correct tick mark is being made bold, but the formatted text is incorrect.

Could you please share the data you've used for this example? Or even better, a small reproducible example on a site like jsFiddle / code sandbox?

My initial thought is that the code you are using to handle the GMT+2 timezone difference (the formatter) might have a slight bug in it. Perhaps you are subtracting by a fixed amount of milliseconds instead of using the browser APIs for this. Something as strange as the few leap milliseconds that are added/remove at the end of each year might be causing this problem.

humpedli:
@SlicedSilver thank you for your answer.

[...]

PR #1510, opened 2024-02-05
add option to disable bold labels in time scale

### LINKED ISSUE #1509: Add an option to not bold time tickmarks

Recently we added an ability to use custom type of time scale points. And for many types like "future expiration" weight value has no sence - all expiration are of the same significance. There is no problem to return 1 as weight for all the points, however this makes all the marks on horz scale bold and there is no way to disable it.

I'd like suggesting adding one more option to time scale options like `disableBoldMarks` that makes time axis drawing all the marks with regular fold

SlicedSilver:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: closes #1509, #1438
- [x] Includes tests
- [x] Documentation update

**Overview of change:**
Adds `allowBoldLabels` option to time scale options. It is enabled by default, but can be set to `false` if you don't want bold labels.

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -4,21 +4,21 @@ module.exports = [
 	{
 		name: 'CJS',
 		path: 'dist/lightweight-charts.production.cjs',
-		limit: '47.96 KB',
+		limit: '47.98 KB',
 	},
 	{
 		name: 'ESM',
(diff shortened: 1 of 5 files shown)


eugene-korobko (approved):

SlicedSilver:
closes #1438

PR #1514, opened 2024-02-08
Fix Firefox Rendering Issue

carterjfulcher:
Type: bugfix 

The relative position of the chart pane widget causes issues on React/Firefox. This persists OS's, but was tested primarily on MacOS and Windows. 

The relative position causes the canvas to fail to render, and shifts it's position. Here's before this change:

image

And here's after, like normal:

image


SlicedSilver:
@carterjfulcher Thank you.

Do you have a small reproduction example for this bug?

It seems strange that this is an issue and I would like to investigate the cause further.

Changing from 'relative' to 'static' appears to be fine here because the element is the only child of a container element which is itself `position: relative`. However, I would rather avoid making unnecessary changes as it could potentially disrupt the chart for other users who utilize external CSS styling. So it would be good to know why this is occurring in your react element on Firefox.

carterjfulcher:
> @carterjfulcher Thank you.
> 
> Do you have a small reproduction example for this bug?
> 
> It seems strange that this is an issue and I would like to investigate the cause further.
> 
> Changing from 'relative' to 'static' appears to be fine here because the element is the only child of a container element which is itself `position: relative`. However, I would rather avoid making unnecessary changes as it could potentially disrupt the chart for other users who utilize external CSS styling. So it would be good to know why this is occurring in your react element on Firefox.

Hi @SlicedSilver. Total agree, I'll work on a easily reproducible example and follow up on this shortly.

SlicedSilver:
@carterjfulcher 
You edited the `.gitignore` and now your PR has the `dist` and `lib` folders included. These folders should not be committed to the repo. 

Could you revert your new commits, and rather provide an example via a site like JSFiddle / codesandbox?

carterjfulcher:
@SlicedSilver whoops, this was for an internal build. I'll rebase

PR #1536, opened 2024-03-07
fix possible ReDoS

illetid:
**Type of PR:**  bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes [CLL-263]

**Overview of change:**
color regex  `const rgbaRe = /^rgba\(\s*(-?\d{1,10})\s*,\s*(-?\d{1,10})\s*,\s*(-?\d{1,10})\s*,\s*(-?[\d]{0,10}(?:\.\d+)?)\s*\)$/;` contains possible weakness, for cases like `'rgba(0,0,0,' + '\t'.repeat(54773) + '\x00'` . Changed regex to `/^rgba\(\s*(-?\d{1,10})\s*,\s*(-?\d{1,10})\s*,\s*(-?\d{1,10})\s*,\s*(-?\d*\.?\d+)\s*\)$/`

### DIFF
--- a/src/helpers/color.ts
+++ b/src/helpers/color.ts
@@ -195,16 +195,14 @@ function normalizeRgbComponent<T extends RedComponent | GreenComponent | BlueCom
 }
 
 function normalizeAlphaComponent(component: AlphaComponent): AlphaComponent {
-	return Number.isNaN(component)
-		? (1 as AlphaComponent)
-		: ((!(component <= 0) && !(component > 0)
-			? (0 as AlphaComponent)
-			: component < 0
-				? (0 as AlphaComponent)
-				: component > 1
-					? (1 as AlphaComponent)
-					: // limit the precision of all numbers to at most 4 digits in fractional part
-					Math.round(component * 10000) / 10000) as AlphaComponent);
+	if (Number.isNaN(component)) {
+		return 1 as AlphaComponent;
+	}
+	if (component <= 0 || component > 1) {
+		return (component < 0 ? 0 : 1) as AlphaComponent;
+	}
+	// limit the precision of all numbers to at most 4 digits in fractional part
+	return Math.round(component * 10000) / 10000 as AlphaComponent;
 }
 
 /**
--- a/src/helpers/color.ts
+++ b/src/helpers/color.ts
@@ -196,7 +196,7 @@ function normalizeRgbComponent<T extends RedComponent | GreenComponent | BlueCom
 
 function normalizeAlphaComponent(component: AlphaComponent): AlphaComponent {
 	if (component <= 0 || component > 1) {
-		Math.min(Math.max(component, 0), 1) as AlphaComponent;
+		return Math.min(Math.max(component, 0), 1) as AlphaComponent;
 	}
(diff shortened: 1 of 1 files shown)


src/helpers/color.ts
```diff
@@ -195,11 +195,16 @@ function normalizeRgbComponent<T extends RedComponent | GreenComponent | BlueCom
...
+	return Number.isNaN(component)
+		? (1 as AlphaComponent)
+		: ((!(component <= 0) && !(component > 0)
+			? (0 as AlphaComponent)
+			: component < 0
+				? (0 as AlphaComponent)
+				: component > 1
+					? (1 as AlphaComponent)
+					: // limit the precision of all numbers to at most 4 digits in fractional part
+					Math.round(component * 10000) / 10000) as AlphaComponent);
```
SlicedSilver, line 207: This code (and the old code) is difficult to read. Would it be possible to refactor this function to be neater? 
For example there are a lot of `as AlphaComponent`, it should be possible to only use `as AlphaComponent` once (on the final result).
Additionally it should be possible change this condition `!(component <= 0) && !(component > 0)` to something a bit more clear. 
It should be fine to write this method in multiple statements which the minifier will convert back into a single return statement.

Essentially, can we make the code easier to read while not increasing the bundle size?

src/helpers/color.ts
```diff
@@ -195,11 +195,14 @@ function normalizeRgbComponent<T extends RedComponent | GreenComponent | BlueCom
...
-				// limit the precision of all numbers to at most 4 digits in fractional part
-				Math.round(component * 10000) / 10000) as AlphaComponent;
+	if (Number.isNaN(component)) {
+		return 1 as AlphaComponent;
+	}
+	if (component <= 0 || component > 1) {
+		return (component < 0 ? 0 : 1) as AlphaComponent;
+	}
+	// limit the precision of all numbers to at most 4 digits in fractional part
+	return Math.round(component * 10000) / 10000 as AlphaComponent;
```
romfrancois, line 205: Isn't that similar to something like 
```
Math.max(0, Math.min(1, component));
```

Maybe it's less readable 🤷‍♂️

src/helpers/color.ts
```diff
@@ -195,11 +195,14 @@ function normalizeRgbComponent<T extends RedComponent | GreenComponent | BlueCom
...
 
 function normalizeAlphaComponent(component: AlphaComponent): AlphaComponent {
-	return (!(component <= 0) && !(component > 0) ? 0 as AlphaComponent :
-		component < 0 ? 0 as AlphaComponent :
-			component > 1 ? 1 as AlphaComponent :
-				// limit the precision of all numbers to at most 4 digits in fractional part
-				Math.round(component * 10000) / 10000) as AlphaComponent;
+	if (Number.isNaN(component)) {
+		return 1 as AlphaComponent;
+	}
```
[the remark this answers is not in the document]
illetid: I had the idea to simplify the regex and add a runtime check, but I found a regex fix, so now this part is obsolete and removed this part

src/helpers/color.ts
illetid: changed, thanks

romfrancois (approved):

PR #1538, opened 2024-03-08
fix: marker positioning

### LINKED ISSUE #1382: Histogram has some space below bars when markers are added
**Lightweight Charts™ Version:** 4.0.1

**Steps/code to reproduce:**

Add histogram to the chart, set priceScaleId: '', add markers to the histogram series

**Actual behavior:**

Some space appears below the histogram

**Expected behavior:**

No space between histogram and bottom of the chart

**Screenshots:**

image

https://jsfiddle.net/exuqzLh5/

illetid:
**Type of PR:**  bugfix, enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #1382
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**
Fixed an issue where `marginBelow` wasn't applied. Added an enhancement that fixes #1382: When a series has only one type of markers, we apply only one margin, i.e. for markers with `position: "aboveBar"` we remove bottom margin and vice versa.

### DIFF
--- a/src/views/pane/series-markers-pane-view.ts
+++ b/src/views/pane/series-markers-pane-view.ts
@@ -136,9 +136,18 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
 				const barSpacing = this._model.timeScale().barSpacing();
 				const shapeMargin = calculateShapeMargin(barSpacing);
 				const marginsAboveAndBelow = calculateShapeHeight(barSpacing) * 1.5 + shapeMargin * 2;
+				let above = marginsAboveAndBelow as Coordinate;
+				let below = marginsAboveAndBelow as Coordinate;
+				const position = this._hasAllMarkerSamePosition();
+				if (position === 'aboveBar') {
+					below = 0 as Coordinate;
+				}
+				if (position === 'belowBar') {
+					above = 0 as Coordinate;
+				}
 				this._autoScaleMargins = {
-					above: marginsAboveAndBelow as Coordinate,
-					below: marginsAboveAndBelow as Coordinate,
+					above,
+					below,
 				};
 			} else {
 				this._autoScaleMargins = null;
@@ -148,6 +157,14 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
 		}
 
 		return this._autoScaleMargins;
+	}
+	protected _hasAllMarkerSamePosition(): string|null {
+		const markers = this._series.indexedMarkers();
+		const markersPositions = markers.every((i: InternalSeriesMarker<TimePointIndex>) => i.position === markers[0].position);
+		if (markersPositions) {
+			return markers[0].position;
+		}
+		return null;
 	}
 
 	protected _makeValid(): void {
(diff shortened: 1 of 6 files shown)


src/views/pane/series-markers-pane-view.ts
```diff
@@ -149,6 +151,14 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
 
 		return this._autoScaleMargins;
 	}
+	protected _hasAllMarkerSamePosition(): string|null {
```
SlicedSilver, line 154: This method will be evaluated on every chart render (such as during scrolling, zooming, and all other actions). Since it will be evaluated so often and it's result would only ever change if `setMarkers` is used to change the markers → I think we could try to memo the function for performance reasons. It should be fine to only invalidate the result when `setMarkers` is used on the series.

SlicedSilver:
We should also have a look at the e2e graphics test cases. I think there are more tests failing than expected such as the tests which have markers positioned above, and below.

SlicedSilver (approved):
LGTM 👍

src/views/pane/series-markers-pane-view.ts
```diff
@@ -131,14 +131,16 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
...
 				const barSpacing = this._model.timeScale().barSpacing();
 				const shapeMargin = calculateShapeMargin(barSpacing);
 				const marginsAboveAndBelow = calculateShapeHeight(barSpacing) * 1.5 + shapeMargin * 2;
+				const position = this._hasAllMarkerSamePosition();
+
 				this._autoScaleMargins = {
-					above: marginsAboveAndBelow as Coordinate,
-					below: marginsAboveAndBelow as Coordinate,
+					above: position === 'belowBar' ? 0 : marginsAboveAndBelow,
+					below: position === 'aboveBar' ? 0 : marginsAboveAndBelow,
```
kirchet, line 143: The condition looks not correct, we should not add margins in case when position is `'inBar'`

src/views/pane/series-markers-pane-view.ts
```diff
@@ -131,14 +131,16 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
 	}
 
 	public autoScaleMargins(): AutoScaleMargins | null {
-		if (this._autoScaleMarginsInvalidated) {
+		if (this._autoScaleMarginsInvalidated && this._dataInvalidated) {
```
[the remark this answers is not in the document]
SlicedSilver: Apologies, my mistake for mentioning to try run the calculation less often at that point.
We can instead limit the `_dataInvalidated` check to only the `_hasAllMarkerSamePosition()` method because that doesn't rely on barSpacing.

src/views/pane/series-markers-pane-view.ts
kirchet: yes, you're right

src/views/pane/series-markers-pane-view.ts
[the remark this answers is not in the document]
kirchet: Yes we can, but we need to add one additional flag like `_hasAllMarkerSamePositionInvalidated` and use it in `_hasAllMarkerSamePosition` or update cached value in `_makeValid`

PR #1539, opened 2024-03-12
fix: markers margin adjustment

illetid:
added @kirchet  suggestions from  #1538

### DIFF
--- a/src/views/pane/series-markers-pane-view.ts
+++ b/src/views/pane/series-markers-pane-view.ts
@@ -18,6 +18,7 @@ import {
 	SeriesMarkersRenderer,
 } from '../../renderers/series-markers-renderer';
 import {
+	calculateAdjustedMargin,
 	calculateShapeHeight,
 	shapeMargin as calculateShapeMargin,
 } from '../../renderers/series-markers-utils';
@@ -96,7 +97,7 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
 	private _autoScaleMarginsInvalidated: boolean = true;
 
 	private _autoScaleMargins: AutoScaleMargins | null = null;
-	private _makersPositions: MarkerPositions | null = null;
+	private _markersPositions: MarkerPositions | null = null;
 	private _renderer: SeriesMarkersRenderer = new SeriesMarkersRenderer();
 
 	public constructor(series: ISeries<SeriesType>, model: IChartModelBase) {
@@ -113,7 +114,7 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
 		this._autoScaleMarginsInvalidated = true;
 		if (updateType === 'data') {
 			this._dataInvalidated = true;
-			this._makersPositions = this._getMarkerPositions();
+			this._markersPositions = null;
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


src/views/pane/series-markers-pane-view.ts
```diff
@@ -111,6 +113,7 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
 		this._autoScaleMarginsInvalidated = true;
 		if (updateType === 'data') {
 			this._dataInvalidated = true;
+			this._makersPositions = this._getMarkerPositions();
```
kirchet, line 116: To not mix immediate and lazy updates we need just invalidate markers positions state here. We can set `_markersPositions` to `null` here and update them in `_getMarkerPositions`

```TS
private getMarkersPositions(): MarkerPositions {
  if (this._markersPositions === null) {
    this._markersPositions = ....; // calculate positions here
  }
  return this._markersPositions;
}
```

src/views/pane/series-markers-pane-view.ts
```diff
@@ -151,13 +154,29 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
...
+
+	protected _calculateAdjustedMargin(margin: number, hasSide: boolean, hasInBar: boolean): number {
+		if (hasSide) {
+			return margin;
+		} else if (hasInBar) {
+			return Math.ceil(margin / 2);
 		}
-		return null;
+
+		return 0;
```
kirchet, line 165: This method doesn't use `this` so I suggest to move this logic to a function.

kirchet (changes requested):

src/views/pane/series-markers-pane-view.ts
```diff
@@ -151,13 +155,22 @@ export class SeriesMarkersPaneView implements IUpdatablePaneView {
...
+				if (!acc[marker.position]) {
+					acc[marker.position] = true;
+				}
+				return acc;
+				// eslint-disable-next-line @typescript-eslint/tslint/config
+			}, {
+				inBar: false,
+				aboveBar: false,
+				belowBar: false,
+			});
```
kirchet, line 171: Everything looks good except this `// eslint-disable-next-line @typescript-eslint/tslint/config`. I guess it's a formatting issue. Try to change the code to

```suggestion
			this._markersPositions = this._series.indexedMarkers().reduce(
				(acc: MarkerPositions, marker: InternalSeriesMarker<TimePointIndex>) => {
					if (!acc[marker.position]) {
						acc[marker.position] = true;
					}
					return acc;
				},
				{
					inBar: false,
					aboveBar: false,
					belowBar: false,
				}
			);
```

kirchet: Thanks

kirchet:
For some reason the pipeline step `install-deps` failed. @SlicedSilver can you take a look at this?

SlicedSilver:
> For some reason the pipeline step `install-deps` failed. @SlicedSilver can you take a look at this?

It appears to be an issue with downloading and installing chromium for puppeteer. https://app.circleci.com/pipelines/github/tradingview/lightweight-charts/2324/workflows/1f3fb2b3-814d-4e8d-b8d4-3346cf8a0bde/jobs/31071

We could try bump the puppeteer version to the latest but it is possible that memlab would also need to be bumped to a newer version. I'll experiment on a different branch.

SlicedSilver:
> @SlicedSilver also to update puppeteer we need to drop node 16, they dropped support for it in 22 major version

We are updating the node requirement in this different PR: https://github.com/tradingview/lightweight-charts/pull/1527
and there is also a task to update all the dependencies before v5 release (CLL-257)

SlicedSilver (approved):

kirchet (approved):

PR #1556, opened 2024-04-05
fix throws "Value is undefined" when hovering mouse over the chart

### LINKED ISSUE #1499: Calling `chart.subscribeCrosshairMove()` throws "Value is undefined" when hovering mouse over the chart
**Lightweight Charts™ Version:** 4.1.2

**Steps/code to reproduce:** 

1) Provide valid data
2) Add a call to `chart.subscribeCrosshairMove()`
3) Hover the mouse over the timeline
4) Alternates between throwing errors and triggering the callback just fine.

See the stackblitz example and screen recording below.
**Actual behavior:**

When calling `chart.subscribeCrosshairMove()`, an error "Value is undefined" is thrown when hovering mouse over the chart. 

**Expected behavior:**

An error should not be thrown.. ? The callback is still firing fine most of the time. However the error also keeps randomly throwing.

**Screenshots:**

https://github.com/tradingview/lightweight-charts/assets/12685308/77dfb684-a049-4b27-8ee5-28ff7e6e7830

Here's the minimal viable reproducable example: https://stackblitz.com/edit/nuxt-starter-ruhnxp?file=app.vue.

nikepan:
**Type of PR:** bugfix

**PR checklist:**

- [+] Addresses an existing issue: fixes #1499 

**Overview of change:**
This fix ignores the bug because the events are asynchronous and when a series is deleted, we can't know about it. We probably need to fix the Click/DblClick events as well, but simultaneous deletion of a series and a click is very rare.

### DIFF
--- a/src/gui/chart-widget.ts
+++ b/src/gui/chart-widget.ts
@@ -826,9 +826,9 @@ export class ChartWidget<HorzScaleItem> implements IDestroyable, IChartWidgetBas
 	): void {
 		try {
 			this._crosshairMoved.fire(() => this._getMouseEventParamsImpl(time, point, event));
-		} catch (e: any) {
-			if (e !instanceof ValueIsUndefinedError) {
-				throw e;
+		} catch (error) {
+			if (!(error instanceof ValueIsUndefinedError)) {
+				throw error;
 			}
 		}
 	}
--- a/src/gui/chart-widget.ts
+++ b/src/gui/chart-widget.ts
@@ -1,6 +1,6 @@
 import { Size, size } from 'fancy-canvas';
 
-import { ensureDefined, ensureNotNull } from '../helpers/assertions';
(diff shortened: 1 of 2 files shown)


SlicedSilver:
@nikepan Thanks for submitting your PR.

I haven't been able to reproduce the initial bug but based on the error in the console seen within the video I think that this PR: https://github.com/tradingview/lightweight-charts/pull/1529 would potentially fix the issue. Could you please check if the other PR does indeed fix your issue?

nikepan:
> @nikepan Thanks for submitting your PR.
> 
> I haven't been able to reproduce the initial bug but based on the error in the console seen within the video I think that this PR: #1529 would potentially fix the issue. Could you please check if the other PR does indeed fix your issue?

Yes. Your fix works fine. Thank you!

PR #824, opened 2021-08-24
WIP: fix multi-pane feature

ntf:
**Type of PR:** enhancement

**PR checklist:**

- [ ] Addresses an existing issue: fixes #50 
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**
This is base on the pane widget and pane separator feature previously implemented in 22c6b03897c926cb9a0295d9b0bbc4fe2b5d83a5.

**Is there anything you'd like reviewers to focus on?**
API design. New API added so far (To be finalized) 

Things I have implemented so far:

- series option to configure which pane a series should be rendered ( pane: number to SeriesOptionsCommon )
- `ChartApi#removePane(index: number)` so that you can remove a pane (and series on it) programmatically.
- `ChartApi#swapPane(first: number, second: number)` for swapping the positions of the two pane.
- code to handle series.applyOptions( { pane: newIndex } );
- `ChartApi#getPaneElements()` for getting container elements to each pane. (I need the container to draw legends to each pane)

Things I am also implementing but not going to be included in this PR:

- React binding with pane and legend supports

Demo here (It maybe outdated): https://jsfiddle.net/adrianntf/6qea5ytv/2/

You can use below command to install my version to your project locally:

`yarn add git+https://github.com/ntf/lightweight-charts.git`
or `npm install git+https://github.com/ntf/lightweight-charts.git`

You should not use this directly in your production environment. However, you may review my code and merge into your own folk of the project.

PR #1557, opened 2024-04-05
Enable pane support

### LINKED ISSUE #50: Multiple panes support
Looking through the documentation i dont see anything along these lines. I want to have my main chart but then add some additional data in a separate chart window. I believe some apps call it an Indicator Window. I was curious if its possible with this library.

illetid:
**Type of PR:** enhancement

**PR checklist:**

- [ ] Addresses an existing issue: fixes #50 
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**
Based on community implementation https://github.com/tradingview/lightweight-charts/pull/824


kirchet (changes requested):

src/model/series.ts
```diff
@@ -261,12 +262,13 @@ export class Series<T extends SeriesType> extends PriceDataSource implements IDe
 		return this._options as SeriesOptionsMap[T];
 	}
 
-	public applyOptions(options: SeriesPartialOptionsInternal<T>): void {
+	public applyOptions(options: SeriesPartialOptionsInternal<T> | DeepPartial<SeriesOptionsCommon>): void {
 		const targetPriceScaleId = options.priceScaleId;
```
kirchet, line 266: If we need a new field then it should be included in `SeriesPartialOptionsInternal<T>`

This options can be useful only to move a series from one pane to another.
I would add a special method to do it instead

```TS
ISeriesAPi {
  ...
  moveToPane(paneIndex: number): void;
}
```

src/gui/chart-widget.ts
```diff
@@ -805,10 +825,10 @@ export class ChartWidget<HorzScaleItem> implements IDestroyable, IChartWidgetBas
 
 	private _onPaneWidgetClicked(
 		time: TimePointIndex | null,
-		point: Point | null,
+		details: Point & PaneInfo | null,
```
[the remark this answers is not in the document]
illetid: it will work for click cases because we register them on the pane creation phase, but not for `_onPaneWidgetCrosshairMoved` because we register listener in `constructor` and we don't know about the pane  there, so pass paneIndex with coordinates looks like nice common solutions. But maybe I didn't get your idea correctly

src/gui/pane-separator.ts
```diff
@@ -128,11 +167,12 @@ export class PaneSeparator implements IDestroyable {
 		this._paneA.setStretchFactor(newUpperPaneStretch);
 		this._paneB.setStretchFactor(newLowerPaneStretch);
 
-		this._chartWidget.model().fullUpdate();
+		this._chartWidget.adjustSize();
```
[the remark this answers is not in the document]
illetid: with `this._chartWidget.model().fullUpdate()` we have glitches when resizing panes via dragging pane separator. It works great with `adjustSize()` though. I'm not sure why, will investigate

src/model/chart-model.ts
```diff
@@ -628,10 +656,90 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
...
+			}
+		});
+
+		this._suppressSeriesMoving = true;
+		if (index !== this._panes.length - 1) {
+			for (let i = index + 1; i < this._panes.length; i++) {
+				const pane = this._panes[i];
+				pane.orderedSources().forEach((source: IPriceDataSource) => {
+					if (source instanceof Series) {
+						(source as Series<SeriesType>).applyOptions({ pane: i - 1 });
```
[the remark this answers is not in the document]
illetid: fixed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/7b747ff7dea8192a423a32b06b53b4a2d1acb1ea could you pls check if you had something like that in your mind?

src/model/chart-model.ts
```diff
@@ -439,6 +453,7 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
 
 	private readonly _timeScale: TimeScale<HorzScaleItem>;
 	private readonly _panes: Pane[] = [];
+	private readonly _panesToIndex: Map<Pane, number> = new Map<Pane, number>();
```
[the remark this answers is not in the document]
illetid: removed, thanks

src/model/series.ts
illetid: fixed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/7b747ff7dea8192a423a32b06b53b4a2d1acb1ea could you pls check if you had something like that in your mind?

src/api/ichart-api.ts
```diff
@@ -330,6 +335,35 @@ export interface IChartApiBase<HorzScaleItem = Time> {
...
+	 */
+	removePane(index: number): void;
+
+	/**
+	 * swap the position of two panes.
+	 *
+	 * @param first - the first index
+	 * @param second - the second index
+	 */
+	swapPane(first: number, second: number): void;
```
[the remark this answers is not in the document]
illetid: renamed, thanks

src/model/chart-model.ts
```diff
@@ -335,6 +338,11 @@ export interface ChartOptionsBase {
 	 * Basic localization options
 	 */
 	localization: LocalizationOptionsBase;
+
+	/**
+	 * Panes options
+	 */
+	panes: PanesOptions;
```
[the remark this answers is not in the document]
illetid: moved to layout options, thanks for the suggestion

src/model/series.ts
kirchet: Yes, thanks

src/gui/chart-widget.ts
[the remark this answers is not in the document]
kirchet: We can easely bind the proper value:

```TS
paneWidget.clicked().subscribe(this._onPaneWidgetClicked.bind(this, paneWidget), this);
```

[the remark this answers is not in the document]
illetid: I tried to do that, but in the end, we call `_getMouseEventParamsImpl` for all cases: click, double click, and hover. And for click and double click, we can pass pane via the `bind,` but we can't do that for hover listener `this.model().crosshairMoved().subscribe(this._onPaneWidgetCrosshairMoved.bind(this), this);` on line 110.  I don't know what is the best solution for this case, maybe you have some ideas

src/model/chart-model.ts
```diff
@@ -607,7 +617,19 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
 	}
 
 	public createPane(index?: number): Pane {
-		const pane = new Pane(this._timeScale, this);
+		if (index !== undefined) {
+			if (index > this._panes.length) {
+				for (let i = this._panes.length; i < index; i++) {
+					this.createPane(i);
+				}
```
kirchet, line 624: Probably we should first validate/modify index like `index = Math.max(0, Math.min(this._panes.length, index))`

src/model/series.ts
```diff
@@ -257,6 +258,14 @@ export class Series<T extends SeriesType> extends PriceDataSource implements IDe
 		return this._barColorerCache;
 	}
 
+	public paneIndex(): number {
+		return this._paneIndex;
+	}
+
+	public setPaneIndex(paneIndex: number): void {
+		this._paneIndex = paneIndex;
```
kirchet, line 266: We don't need to store pane index in series and keep it in sync when we move series from pane to pane or remove some pane. We can calculate id dynamically

```TS
class ChartModel {
  ...
  public paneIndex(series: ISeries): number | null {
    for (let i = 0; i < this._panes.length; i += 1) {
      if (this._panes[i].contains(series)) {
        return i;
      }
    }
    return null;
  }
}
```

kirchet (changes requested):

src/gui/chart-widget.ts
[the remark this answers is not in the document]
illetid: fixed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/ed08ef7fb1f5f948d4ce0cfd919fb544b24e8e35

src/model/chart-model.ts
illetid: fixed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/6bdd22c8c7c6d35c9c4803704d4326287b74a218

src/model/series.ts
illetid: fixed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/6bdd22c8c7c6d35c9c4803704d4326287b74a218

kirchet:
Also take a look at horizontal crosshair line visibility, it should be painted on the active pane only. The same logic should be applied for crosshair price label as well.

image

And for some reason vertical line is not painted on the first pane

src/model/chart-model.ts
```diff
@@ -628,10 +649,89 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
...
+		if (index !== this._panes.length - 1) {
+			for (let i = index + 1; i < this._panes.length; i++) {
+				const pane = this._panes[i];
+				pane.orderedSources().forEach((source: IPriceDataSource) => {
+					if (source instanceof Series) {
+						this.moveSeriesToPane(source as Series<SeriesType>, i - 1);
+					}
+				});
+			}
+		}
```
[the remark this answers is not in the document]
illetid: thanks for the suggestion, fixed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/cc0421f9aa4016a98c343826e33c2aeecbddeb9f

src/model/chart-model.ts
```diff
@@ -628,10 +649,89 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
...
+		firstPane.orderedSources().forEach((source: IPriceDataSource) => {
+			if (source instanceof Series) {
+				this.moveSeriesToPane(source as Series<SeriesType>, second);
+			}
+		});
+		secondPane.orderedSources().forEach((source: IPriceDataSource) => {
+			if (source instanceof Series) {
+				this.moveSeriesToPane(source as Series<SeriesType>, first);
+			}
+		});
```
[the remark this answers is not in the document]
illetid: fixed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/cc0421f9aa4016a98c343826e33c2aeecbddeb9f

illetid:
> Also take a look at horizontal crosshair line visibility, it should be painted on the active pane only. The same logic should be applied for crosshair price label as well.
> 
> image
> 
> And for some reason vertical line is not painted on the first pane


thanks for spotting this one, fixed crosshair in https://github.com/tradingview/lightweight-charts/pull/1557/commits/5fc37ee4450b9db08174c048ed61b045e0d63677

src/model/chart-model.ts
```diff
@@ -450,6 +456,8 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
 	private readonly _priceScalesOptionsChanged: Delegate = new Delegate();
 	private _crosshairMoved: Delegate<TimePointIndex | null, Point | null, TouchMouseEventData | null> = new Delegate();
 
+	private _suppressSeriesMoving: boolean = false;
```
kirchet, line 459: Seems that it's not used anymore

src/model/crosshair.ts
```diff
@@ -260,7 +259,15 @@ export class Crosshair extends DataSource {
 	}
 
 	public paneViews(pane: Pane): readonly IPaneView[] {
-		return this._pane !== null ? [this._paneView, this._markersPaneView] : [];
+		if (this._pane === null) {
+			return [];
+		}
```
kirchet, line 264: Looks that these lines can be removed

src/model/crosshair.ts
```diff
@@ -293,7 +300,10 @@ export class Crosshair extends DataSource {
 	}
 
 	public updateAllViews(): void {
-		this._paneView.update();
+		this._model.panes().forEach((pane: Pane) => {
+			this._crosshairPaneViewCache.get(pane)?.update();
+		});
+		this._crosshairPaneViewCache.get(this._pane as Pane)?.update();
```
kirchet, line 306: This line can be removed, we updated all views for each pane above

kirchet (changes requested):

src/model/crosshair.ts
illetid: thanks, removed unused code in [7391663](https://github.com/tradingview/lightweight-charts/pull/1557/commits/7391663dc766c84f6b1b129f4663e990db4fb206)

src/gui/pane-separator.ts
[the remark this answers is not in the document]
kirchet: Fixed in e612037c3effdd583901d4ef6e6f027be24f8dc0

src/gui/chart-widget.ts
```diff
@@ -760,8 +779,9 @@ export class ChartWidget<HorzScaleItem> implements IDestroyable, IChartWidgetBas
 
 	private _getMouseEventParamsImpl(
 		index: TimePointIndex | null,
-		point: Point | null,
-		event: TouchMouseEventData | null
+		details: Point | null,
```
kirchet, line 782: Let's rename it back to `point` here and in another places in this file

src/model/chart-model.ts
```diff
@@ -1047,4 +1151,12 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
 
 		return layoutOptions.background.color;
 	}
+
+	private _cleanupIfPaneIsEmpty(pane: Pane): void {
+		if (pane.dataSources().length === 0 && this._panes.length > 1) {
+			this._panes.splice(this.getPaneIndex(pane), 1);
+			const mask = new InvalidateMask(InvalidationLevel.Full);
+			this._invalidate(mask);
```
kirchet, line 1159:
```suggestion
			this.fullUpdate();
```

src/model/pane.ts
```diff
@@ -59,9 +59,11 @@ export class Pane implements IDestroyable {
 		if (options.leftPriceScale) {
 			this._leftPriceScale.applyOptions(options.leftPriceScale);
 		}
-		if (options.rightPriceScale) {
+
+		if (this._rightPriceScale.id() === DefaultPriceScaleId.Right && options.rightPriceScale) {
```
kirchet, line 63: Why do we check additionaly for `this._rightPriceScale.id() === DefaultPriceScaleId.Right` but don't do the same for the left price scale?

src/views/pane/crosshair-marks-pane-view.ts
```diff
@@ -57,44 +59,54 @@ export class CrosshairMarksPaneView implements IUpdatablePaneView {
...
 	}
 
-	private _updateImpl(): void {
+	private _updateImpl(pane: Pane): PaneRendererMarks[] {
 		const forceHidden = this._crosshair.options().mode === CrosshairMode.Hidden;
 
-		const serieses = this._chartModel.serieses();
+		const serieses = this._chartModel.serieses()
+			.map((datasource: IPriceDataSource, index: number): [ISeries<SeriesType>, number] => [datasource as ISeries<SeriesType>, index])
+			.filter((entry: [IPriceDataSource, number]) => pane.dataSources().includes(entry[0]));
```
kirchet, line 86: There is no need to get series from the whole chart, we should use series from the provided pane only

src/views/pane/crosshair-marks-pane-view.ts
```diff
@@ -37,7 +39,7 @@ export class CrosshairMarksPaneView implements IUpdatablePaneView {
 	private readonly _compositeRenderer: CompositeRenderer = new CompositeRenderer();
 	private _markersRenderers: PaneRendererMarks[] = [];
 	private _markersData: MarksRendererData[] = [];
-	private _invalidated: boolean = true;
+	private _validated: Map<Pane, PaneRendererMarks[]> = new Map();
```
kirchet, line 42: Instead of this we should create  `CrosshairMarksPaneView` for each pane.

kirchet (changes requested):

src/model/chart-model.ts
```diff
@@ -862,9 +927,13 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
 		return this._priceScalesOptionsChanged;
 	}
 
-	public createSeries<T extends SeriesType>(seriesType: T, options: SeriesOptionsMap[T], customPaneView?: ICustomSeriesPaneView<HorzScaleItem>): Series<T> {
-		const pane = this._panes[0];
-		const series = this._createSeries(options, seriesType, pane, customPaneView);
+	public createSeries<T extends SeriesType>(seriesType: T, options: SeriesOptionsMap[T], paneIndex: number = 0, customPaneView?: ICustomSeriesPaneView<HorzScaleItem>): Series<T> {
+		if (this._panes.length - 1 <= paneIndex) {
+			paneIndex = Math.max(0, Math.min(this._panes.length, paneIndex));
```
eugene-korobko, line 932: It looks like we can just use `this._panes.length` here

Since this._panes.length - 1 <= paneIndex, Math.min(this._panes.length, paneIndex) is always equal to this._panes.length
And also paneIndex could not be negative

src/api/ichart-api.ts
```diff
@@ -330,6 +335,35 @@ export interface IChartApiBase<HorzScaleItem = Time> {
...
+	 * @param second - the second index
+	 */
+	swapPanes(first: number, second: number): void;
+
+	/**
+	 * Returns the DOM elements of the chart panes.
+	 *
+	 * @returns Array of the chart panes.
+	 */
+	getPaneElements(): HTMLElement[];
```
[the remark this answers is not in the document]
karaevmaksim: How we can place legends for each pane without that?

src/gui/chart-widget.ts
illetid: changed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/f90efe5b0bdd61fbd3bc3f9eaeb15d02fd34c3b0

src/views/pane/crosshair-marks-pane-view.ts
illetid: fixed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/cced5695984042f22399ae78514a26ef6437699d

src/views/pane/crosshair-marks-pane-view.ts
illetid: thanks, fixed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/cced5695984042f22399ae78514a26ef6437699d

src/model/pane.ts
illetid: it was part of the original community implementation, forgot to cleanup, reverted this change, thanks

src/model/chart-model.ts
illetid: we use it to normalize user input in cases like if the user calls `chart.addLineSeries({}, 3);` but we have only one pane, instead of adding empty panes, we add only one new pane with this series, not sure if it is the best way to handle this, maybe we should add some assertions

src/api/iseries-api.ts
```diff
@@ -335,4 +335,13 @@ export interface ISeriesApi<
...
 	detachPrimitive(primitive: ISeriesPrimitive<HorzScaleItem>): void;
+
+	/**
+	 * Move the series to another pane.
+	 *
+	 * If the pane with the specified index does not exist, the pane will be created.
+	 *
+	 * @param paneIndex - The index of the pane. Should be a number between 0 and the total number of panes - 1.
+	 */
+	moveToPane(paneIndex: number): void;
```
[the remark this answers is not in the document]
illetid: could you pls check https://github.com/tradingview/lightweight-charts/pull/1557/commits/d9c4ab498ed84e56497d40ccb4c23a7a41c164a9 I've added Api to get pane/series on chart lvl and on PaneApi and SeriesApi to make it easier for user, but not sure is it okay?

src/model/crosshair.ts
kirchet: Added back again in https://github.com/tradingview/lightweight-charts/pull/1557/commits/cced5695984042f22399ae78514a26ef6437699d

src/views/pane/crosshair-marks-pane-view.ts
```diff
@@ -59,54 +60,44 @@ export class CrosshairMarksPaneView implements IUpdatablePaneView {
...
 
-	private _updateImpl(pane: Pane): PaneRendererMarks[] {
+	private _updateImpl(): void {
 		const forceHidden = this._crosshair.options().mode === CrosshairMode.Hidden;
-
-		const serieses = this._chartModel.serieses()
-			.map((datasource: IPriceDataSource, index: number): [ISeries<SeriesType>, number] => [datasource as ISeries<SeriesType>, index])
-			.filter((entry: [IPriceDataSource, number]) => pane.dataSources().includes(entry[0]));
-
+		const serieses = this._pane.dataSources() as ISeries<SeriesType>[];
```
kirchet: Now a pane can contain only series, so let's rename `IPane.dataSources()` to `series():  ISeries<SeriesType>[]`

kirchet (changes requested):

src/views/pane/crosshair-marks-pane-view.ts
kirchet: UPD: The same is applicable for `IPane.orderedSources()`

kirchet: Or we need to add a type guard like `isSeries(source: IDataSource): source is ISeries<SeriesType>`

src/api/ichart-api.ts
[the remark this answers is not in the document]
illetid: removed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/ed8195b4695122c7ef0611cad025502b60ead31d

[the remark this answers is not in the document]
illetid: @karaevmaksim you will be able to access HTML elements from the javascript or CSS, that API exposed just a `<tr>` element. you can query them with something like `document.querySelector(".tv-lightweight-charts table tr:nth-child(odd)`, note that we need to use `odd` to filter out pane separators

src/views/pane/crosshair-marks-pane-view.ts
illetid: added in https://github.com/tradingview/lightweight-charts/pull/1557/commits/5a4260f789f21506a2dda61252f4cec22dd9b226

src/api/ichart-api.ts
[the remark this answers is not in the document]
SlicedSilver: We already expose an API to get a reference to the element created by the library for the chart: https://tradingview.github.io/lightweight-charts/docs/api/interfaces/IChartApi#chartelement

Yes, we can expect the developer to use a query selector on that element to find the pane elements but if it is not increasing the bundle size then I think we can expose this API for getting the pane element.

If we decide to rearrange panes via css instead of how they are defined in the DOM, or start adding more elements which aren't panes then a querySelector might not be so simple to write to find the nth Pane.

In general, it is a useful API method to have if you are writing plugins which require some UI rendered outside of the canvas (a legend might be a good example).

We can address some concerns about exposing these elements by adding more details to the documentation comment, such as 'storing references to these elements might result in a memory leak if the pane element is removed by the library... etc'

src/api/iseries-api.ts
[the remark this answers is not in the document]
eugene-korobko: Yes, they look nice

src/model/chart-model.ts
eugene-korobko: Yes, I understand this, I mean we can use 

```
paneIndex = this._panes.length;
```
instead of 

```
paneIndex = Math.max(0, Math.min(this._panes.length, paneIndex));
```

Or we could remove `if` and keep `Math.min(this._panes.length, paneIndex)`, there is no sence to have both `if` and `min`

src/api/ichart-api.ts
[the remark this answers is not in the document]
eugene-korobko: I generally do not mind for exposiong DOM elements, however I believe API should be systemic to avoid breaking changes in future. 

Why don't to have `paneElement` in `paneAPI` like chart does? Why don't we have getters for scales and other elements?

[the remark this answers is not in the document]
SlicedSilver: Getting into more philosophical discussions here, but as a user of the library I would prefer a known breaking change over an unknown breaking change. Using `querySelector` requires knowing how the library generates the internal dom (i.e knowing the internal logic which is subject to change). If we change how we do this then it will be breaking change for users who needed to use `querySelector` and it wouldn't be in our release notes.
For me, this is more likely then changing our internal code to not have an element for each pane but if we do then we mention it as a breaking change.

My personal opinion is that this would be useful (or at least 'nice') to include. It will be highly likely that we would want to add buttons to the pane for actions such as: remove pane, move pane, make pane fullscreen, ... which would appear on the top right of the pane (like on CL). I was expecting that we would release this as a Plugin because it would be such a common request but not required to be part of the 'core' library.

If we still don't want to add this method then a compromise could be to at least add an attribute to the element in the DOM so it is easy to query select such as `<div data-pane-index="2">...</div>`.

src/model/chart-model.ts
illetid: fixed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/70f36d359cbd59c852267b17cdedc252df7a848b . I changed the signature of `_createSeries` to receive `pane` instance and removed that `Math.max` check from `createSeries` because we do it anyway in `createPane`

src/api/chart-api.ts
```diff
@@ -312,8 +326,23 @@ export class ChartApi<HorzScaleItem> implements IChartApiBase<HorzScaleItem>, Da
...
+	public getPaneBySeries(seriesApi: SeriesApi<SeriesType, HorzScaleItem>): IPaneApi<HorzScaleItem> {
+		const series = this._seriesMap.get(seriesApi);
+		const pane = ensureNotNull(this._chartWidget.model().paneForSource(series as IPriceDataSource));
+		return this._getPaneApi(pane);
+	}
+
+	public getSeriesByPane(paneApi: PaneApi<HorzScaleItem>): ISeriesApi<SeriesType, HorzScaleItem>[] {
+		const pane = this._chartWidget.model().panes()[paneApi.paneIndex()];
+		return pane.series().map((source: Series<SeriesType>) => this._mapSeriesToApi(source)) ?? [];
+	}
```
kirchet, line 342: It's either should be a method in `PaneApi` or this method should have an argument like `paneIndex: number`

I prefer the first option.

The same is for `getPaneBySeries`, probably we should move it to `SeriesApi.getPane(): IPaneApi`

cc @eugene-korobko

illetid: We have the `series()` method in `PaneApi`, which is a call of `getSeriesByPane` same we have `getPane` method  in SeriesApi, which call of `getPaneBySeries`

src/api/ichart-api.ts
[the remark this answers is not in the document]
eugene-korobko: ok, what about moving it to IPAneApi? 

IPaneAPI.getHTMLElement()

src/api/chart-api.ts
kirchet: I still think that moving these methods to `IPaneApi.getSeries()` and `ISeriesApi.getPane()` is a better option

illetid: we need to use `chartApi`'s `_seriesMap` and it is internal property, I don't think we should expose it, and I didn't figure out how we could get it without method duplication, maybe you have some ideas

src/api/ichart-api.ts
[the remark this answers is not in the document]
illetid: added `IPaneAPI.getHTMLElement()`  in https://github.com/tradingview/lightweight-charts/pull/1557/commits/6af4122540fd859b46a05da50e191a739604b6c7 , thanks for suggestion

src/api/chart-api.ts
illetid: we have  `IPaneApi.getSeries()` and `ISeriesApi.getPane()` and they just call `ChartApi` methods because we need access to internal `_seriesMap` property, what do you think about this approach? @kirchet @SlicedSilver

kirchet: I don't see anything wrong/bad if we add a method like `ChartApi.getSeriesApi(series: ISeries): ISeriesApi`. But anyway we can work around it with a getter provided to `PaneApi` constructor.

```TS
export class PaneApi<HorzScaleItem> implements IPaneApi<HorzScaleItem> {
  ...
  private readonly _seriesApiGetter: (series: ISeries) => ISeriesApi;

  public constructor(
    chartApi: IChartApiBase<HorzScaleItem>,
    chartWidget: IChartWidgetBase,
    pane: Pane,
    seriesApiGetter: (series: ISeries) => ISeriesApi
) {
  this._seriesApiGetter = seriesApiGetter;
}
```

illetid: thanks for the suggestion, changed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/9c5b9f73a1c7bd3d39e54520a4078bc1cb217615

src/model/chart-model.ts
```diff
@@ -628,10 +646,58 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
...
 	}
 
+	public removePane(index: number): void {
+		if (this._panes.length === 1) {
+			return;
+		}
+		this._panes.splice(index, 1);
+		this.fullUpdate();
+	}
+
```
SlicedSilver, line 659: This implementation allows the use of a negative number.
`splice` supports the use of a negative number so if this is intended then it should be mentioned in the JSDoc comment on the API.

So `-2` would mean the second last pane in the array will be removed. 

We should either document this, or add an assert to restrict the numbers supplied.

src/model/chart-model.ts
```diff
@@ -628,10 +646,58 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
...
+		if (this._panes.length < 2) {
+			return;
+		}
+
+		assert(paneIndex >= 0 && paneIndex < this._panes.length, 'Invalid pane index');
+		const targetPane = this._panes[paneIndex];
+
+		const totalStretch = this._panes.reduce((prevValue: number, pane: Pane) => prevValue + pane.stretchFactor(), 0);
+		const totalHeight = this._panes.reduce((prevValue: number, pane: Pane) => prevValue + pane.height(), 0);
+		const maxPaneHeight = totalHeight - 30 * (this._panes.length - 1);
```
SlicedSilver, line 670: Don't use 30 but rather use the constant you defined elsewhere.

src/model/chart-model.ts
```diff
@@ -954,6 +1022,31 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
...
 	}
 
+	public seriesPaneIndex(series: Series<SeriesType>): number | null {
+		for (let i = 0; i < this._panes.length; i += 1) {
+			if (this._panes[i].series().includes(series)) {
+				return i;
+			}
+		}
+		return null;
+	}
```
SlicedSilver, line 1032:
```suggestion
	private _seriesPaneIndex(series: Series<SeriesType>): number {
		return this._panes.findIndex((pane: Pane) => pane.series().includes(series));
	}
```
Does it need to be public, and do we need to be able to return `null` if not found?
To me it only looks like it is used in `moveSeriesToPane()` and we don't check for null (only that it is different to the new index)

src/gui/chart-widget.ts
```diff
@@ -317,9 +324,18 @@ export class ChartWidget<HorzScaleItem> implements IDestroyable, IChartWidgetBas
...
 	}
 
-	public paneSize(): Size {
-		// we currently only support a single pane.
-		return ensureDefined(this._paneWidgets[0]).getSize();
+	public paneSize(paneIndex: number): Size {
+		return ensureDefined(this._paneWidgets[paneIndex]).getSize();
+	}
+
+	public adjustSize(): void {
```
SlicedSilver, line 331: Does this need to be public?

src/model/default-price-scale.ts
```diff
@@ -1,6 +1,6 @@
 export const enum DefaultPriceScaleId {
 	Left = 'left',
-	Right = 'right',
+	Right = 'right'
```
SlicedSilver, line 3: Unneeded change?

src/model/pane.ts
```diff
@@ -151,6 +159,10 @@ export class Pane implements IDestroyable {
 		this.updateAllSources();
 	}
 
+	public series(): readonly Series<SeriesType>[] {
```
SlicedSilver, line 162: Could this be `ISeries` instead?

src/model/pane.ts
```diff
@@ -39,7 +45,7 @@ export class Pane implements IDestroyable {
 	private _leftPriceScale: PriceScale;
 	private _rightPriceScale: PriceScale;
 
-	public constructor(timeScale: ITimeScale, model: IChartModelBase) {
+	public constructor(timeScale: ITimeScale, model: IChartModelBase, initialPaneIndex: number = 0) {
```
SlicedSilver, line 48: I think we can remove `initialPaneIndex` here

src/model/series.ts
```diff
@@ -171,7 +171,6 @@ export class Series<T extends SeriesType> extends PriceDataSource implements IDe
 		this._priceAxisViews = [priceAxisView];
 
 		this._panePriceAxisView = new PanePriceAxisView(priceAxisView, this, model);
-
```
SlicedSilver, line 174: [Minor] unneeded change?

src/views/pane/crosshair-marks-pane-view.ts
```diff
@@ -69,22 +63,33 @@ export class CrosshairMarksPaneView implements IUpdatablePaneView {
...
+			});
+			this._compositeRenderer.setRenderers(this._markersRenderers);
+		}
+	}
+
 	private _updateImpl(): void {
 		const forceHidden = this._crosshair.options().mode === CrosshairMode.Hidden;
-
-		const serieses = this._chartModel.serieses();
+		const serieses = this._pane.orderedSources() as ISeries<SeriesType>[];
```
SlicedSilver, line 81: Do we need the type assertion here? If possible it would be better to not use `as`

src/views/pane/crosshair-pane-view.ts
```diff
@@ -45,8 +46,8 @@ export class CrosshairPaneView implements IPaneView {
 	}
 
 	private _updateImpl(): void {
+		const pane = this._pane;
```
SlicedSilver, line 49: in this function we do the following:
```js
const pane = this._pane;
// ...
const crosshairOptions = pane.model().options().crosshair;
// ...
data.horzLine.visible = visible && this._source.horzLineVisible(this._pane);
```
Let's either use `this._pane` everywhere, or `pane` everywhere

tests/e2e/coverage/test-cases/panes/options.js
```diff
@@ -0,0 +1,29 @@
+function simpleData() {
+	return [
+		{ time: 1663740000, value: 10 },
+		{ time: 1663750000, value: 20 },
+		{ time: 1663760000, value: 30 },
+	];
+}
+
+function interactionsToPerform() {
+	return [];
```
SlicedSilver, line 10: [Minor] Would be good if we tested the drag adjustment of the pane separator.

Only if it is easy. `tests/e2e/interactions/test-cases/markers/text-hit-test.js` is a similar-ish type of test case

tests/e2e/helpers/perform-interactions.ts
```diff
@@ -39,7 +39,8 @@ export type InteractionTarget =
 	| 'timescale'
 	| 'leftpricescale'
 	| 'rightpricescale'
-	| 'pane';
+	| 'pane'
+	| 'paneSeparator';
```
SlicedSilver, line 43: 👍

SlicedSilver (changes requested):

src/api/options/layout-options-defaults.ts
```diff
@@ -10,5 +10,9 @@ export const layoutOptionsDefaults: LayoutOptions = {
 	textColor: '#191919',
 	fontSize: 12,
 	fontFamily: defaultFontFamily,
+	panes: {
+		enableResize: true,
+		separatorColor: '#2B2B43',
```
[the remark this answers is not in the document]
illetid: I picked it to be the same as timeline borderColor, and `#E0E3EB` has a lack of contrast and looks like chart grid lines
image

src/model/chart-model.ts
```diff
@@ -429,6 +429,12 @@ export interface IChartModelBase {
 	setTimeScaleAnimation(animation: ITimeScaleAnimation): void;
 
 	stopTimeScaleAnimation(): void;
+	moveSeriesToPane(series: Series<SeriesType>, newPaneIndex: number): void;
```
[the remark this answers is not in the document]
illetid: this method uses `series.applyOptions` and we don't have this method  in `ISeries` declaration

src/api/ichart-api.ts
```diff
@@ -330,6 +335,28 @@ export interface IChartApiBase<HorzScaleItem = Time> {
 	 */
 	takeScreenshot(): HTMLCanvasElement;
 
+	/**
+	 * Removes a pane with index
+	 *
+	 * @returns array of pane's Api
+	 */
+	panes(): IPaneApi<HorzScaleItem>[];
```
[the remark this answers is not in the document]
illetid: changed, thanks

src/api/ichart-api.ts
```diff
@@ -367,7 +394,8 @@ export interface IChartApiBase<HorzScaleItem = Time> {
 	 * Returns the dimensions of the chart pane (the plot surface which excludes time and price scales).
 	 * This would typically only be useful for plugin development.
 	 *
+	 * @param paneIndex - The index of the pane
 	 * @returns Dimensions of the chart pane
 	 */
-	paneSize(): PaneSize;
+	paneSize(paneIndex: number): PaneSize;
```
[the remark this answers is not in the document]
illetid: Changed tsdoc to specify the default parameter, but we can't have both optional parameter and default value

src/gui/chart-widget.ts
```diff
@@ -485,6 +501,9 @@ export class ChartWidget<HorzScaleItem> implements IDestroyable, IChartWidgetBas
 		const stretchPixels = totalPaneHeight / totalStretch;
 
 		let accumulatedHeight = 0;
+
+		const pixelRatio = document.body.ownerDocument.defaultView?.devicePixelRatio || 1;
```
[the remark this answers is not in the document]
illetid: thanks, changed to `window`

src/gui/pane-separator.ts
```diff
@@ -80,68 +74,161 @@ export class PaneSeparator implements IDestroyable {
...
 	}
-
 	private _updateBorderColor(): void {
-		this._cell.style.background = this._chartWidget.options().timeScale.borderColor;
+		this._cell.style.background = this._chartWidget.options().layout.panes.separatorColor;
 	}
 
+	private _mouseOverEvent(event: TouchMouseEvent): void {
+		if (this._handle !== null) {
+			this._handle.element.style.backgroundColor = 'hsla(225,8%,72%,.2)';
```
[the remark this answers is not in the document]
illetid: added configuration for the hover color, thanks

src/gui/pane-separator.ts
```diff
@@ -80,68 +74,161 @@ export class PaneSeparator implements IDestroyable {
...
-		this._minPaneHeight = 0;
-		this._maxPaneHeight = 0;
-		this._pixelStretchFactor = 0;
+		if (this._resizeInfo === null || this._handle === null) {
+			return;
+		}
+
+		this._resizeInfo = null;
+		this._handle.backgroundElement.style.display = 'none';
+		this._handle.element.style.backgroundColor = '';
```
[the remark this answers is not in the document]
illetid: removed background color change

src/gui/pane-widget.ts
```diff
@@ -510,6 +510,10 @@ export class PaneWidget implements IDestroyable, MouseEventHandlers {
 		return this._leftPriceAxisWidget;
 	}
 
+	public getPaneCell(): HTMLElement {
+		return this._paneCell;
+	}
+
```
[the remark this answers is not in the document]
illetid: forgot to delete before, removed now, thanks

src/model/chart-model.ts
illetid: added assertion, thanks

src/model/chart-model.ts
illetid: moved to constant, thanks

src/gui/chart-widget.ts
illetid: this method was unused, removed it

src/model/pane.ts
illetid: unfortunately no, we use it in `moveSeriesToPane` and it works with `Series` not `ISeries` and it didn't compile

src/model/pane.ts
illetid: removed

src/views/pane/crosshair-marks-pane-view.ts
illetid: changed method to get series and removed `as`

src/views/pane/crosshair-pane-view.ts
illetid: fixed, thanks

tests/e2e/coverage/test-cases/panes/options.js
illetid: added drag test, thanks

src/gui/chart-widget.ts
```diff
@@ -791,11 +811,15 @@ export class ChartWidget<HorzScaleItem> implements IDestroyable, IChartWidgetBas
 		const hoveredObject = hoveredSource !== null && hoveredSource.object !== undefined
 			? hoveredSource.object.externalId
 			: undefined;
+		const paneIndex = pane
+			? this._paneWidgets.indexOf(pane)
+			: this.model().panes().indexOf(this.model().crosshairSource().pane() as Pane);
```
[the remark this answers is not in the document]
illetid: I cast it to `Pane` to avoid adding one more check if it is `null` because we check for `-1` after so it won't cause any issues

kirchet:
It's something wrong here 🙂

image

SlicedSilver:
> It's something wrong here 🙂
> 
> image

'QA' testing coming soon. 😅 Been focused on the API design review first.

Agree that we should only have a single TV attribution logo (on the bottom pane).

src/gui/chart-widget.ts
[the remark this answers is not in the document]
SlicedSilver: If typescript is saying that the result could be `null` then perhaps it is better to use `ensureNotNull()`.

src/model/pane.ts
SlicedSilver: Is that not a sign that perhaps there has been a public method added to `Series` but not added to `ISeries`?

src/api/ichart-api.ts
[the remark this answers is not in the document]
SlicedSilver: At the `ichart-api` level it is possible to have it as an optional parameter, while the internal API has it specified with a default value.

src/api/iseries-api.ts
```diff
@@ -335,4 +336,20 @@ export interface ISeriesApi<
...
 	 */
 	detachPrimitive(primitive: ISeriesPrimitive<HorzScaleItem>): void;
+
+	/**
+	 * Move the series to another pane.
+	 *
+	 * If the pane with the specified index does not exist, the pane will be created.
+	 *
+	 * @param paneIndex - The index of the pane. Should be a number between 0 and the total number of panes - 1.
+	 */
```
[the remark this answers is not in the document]
SlicedSilver: The new wording is now
```
The index of the pane. Should be a number between 0 and the total number of panes + 1.
```

If you had 2 panes, then the above sentence implies that the following are valid: 0, 1, 2, 3.

So the sentence should probably be:
```
The index of the pane. Should be a number between 0 and the total number of panes.
```

src/gui/chart-widget.ts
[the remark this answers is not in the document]
illetid: fixed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/9d3611d5dbfa11bba4bcce144a4215f88bbe458e

illetid:
> > It's something wrong here 🙂
> > image
> 
> 'QA' testing coming soon. 😅 Been focused on the API design review first.
> 
> Agree that we should only have a single TV attribution logo (on the bottom pane). Also the price label is different. Is it the exact same data? or is it randomised a bit?

fixed logo duplication in https://github.com/tradingview/lightweight-charts/pull/1557/commits/c391ec18accda43087f4bf3a345fe69a7b167f8e

website/tutorials/how_to/panes.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+```js
+chart.removePane(1);
+```
+
+Note: Removing a pane also removes any series contained within it.
+
+## Full Example
+
+import UsageGuidePartial from "./_usage-guide-partial.mdx";
```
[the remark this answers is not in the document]
illetid: thanks, added your previous suggestions and also commited the `UsageGuidePartial` in https://github.com/tradingview/lightweight-charts/pull/1557/commits/aefd079808d60c390a077c1b0808e91c581750e3#diff-53cd3e1dfe2906463ae48d7fdbbf44441a4b0ee796cea74fa9b075a0e1b8200d.

website/tutorials/how_to/panes.mdx
```diff
@@ -0,0 +1,90 @@
...
+        panes: {
+            separatorColor: '#ff0000',
+            separatorHoverColor: '#00ff00',
+            enableResize: false,
+        },
+    },
+});
+```
+
+Lightweight Charts™ includes  [`PaneApi`](/docs/next/api/interfaces/IPaneApi) that allows you to control each pane. The API has methods to get information on the pane such as `getHeight()`, `paneIndex(), and `getSeries()`. It also contains methods to adjust the pane parameters, for example `setHeight(height)` and `moveTo(paneIndex)`.
```
KatPierce, line 58:
```suggestion
Lightweight Charts™ includes  [`PaneApi`](/docs/next/api/interfaces/IPaneApi) that allows you to control each pane. The API has methods to get information on the pane such as `getHeight()`, `paneIndex()`, and `getSeries()`. It also contains methods to adjust the pane parameters, for example `setHeight(height)` and `moveTo(paneIndex)`.
```

website/tutorials/how_to/panes.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+To get the `PaneApi` instance for each pane, you need to call the `panes` method on the `ChartApi` instance.
```
KatPierce, line 60:
```suggestion
To get a `PaneApi` instance for each pane, you need to call the `panes` method on the `ChartApi` instance.
```

website/tutorials/how_to/_usage-guide-partial.mdx
```diff
@@ -0,0 +1,16 @@
+<details>
+<summary>How to use the code sample</summary>
+<strong>The code presented below requires:</strong>
+
+- That `createChart` has already been imported. See [Getting Started](/docs#creating-a-chart) for more information,
+- and that there is an html div element on the page with an `id` of `container`.
```
KatPierce, line 6:
```suggestion
<strong>Before you use the code sample below, you should do the following::</strong>

-  Import `createChart`. Refer to [Getting Started](/docs#creating-a-chart) for more information.
-  Add an HTML div element with the `id` of the `container`.
```

website/tutorials/how_to/panes.js
```diff
@@ -0,0 +1,138 @@
...
+// Lightweight Charts™ Example: Price and Volume
+// https://tradingview.github.io/lightweight-charts/tutorials/how_to/price-and-volume
+
+// remove-end
+const chartOptions = {
+	layout: {
+		textColor: CHART_TEXT_COLOR,
+		background: { type: 'solid', color: CHART_BACKGROUND_COLOR },
+		// highlight-start
+		// setting this to false will disable the resize of the panes by the user
```
KatPierce, line 11: I suppose that this line is missing in the code
`enableResize: false,`

website/docs/panes.md
```diff
@@ -0,0 +1,23 @@
...
+
+# Panes
+
+Panes are essential elements that help segregate data visually within a single chart.
+
+Panes are useful when you have a chart that needs to show more than one kind of data. For example, you might want to see a stock's price over time in one pane and its trading volume in another. This setup helps users get a fuller picture without cluttering the chart.
+
+By default, Lightweight Charts has a single pane, however, you can add more panes to the chart to display different series in separate areas. For detailed examples and code snippets on how to implement panes in your charts [see tutorial](/tutorials/how_to/panes).
+
+While the specific methods to manipulate panes are covered in the detailed [example](/tutorials/how_to/panes), it's important to note that Lightweight Charts provides an [API for pane management](/api/interfaces/IPaneApi.md). This includes adding new panes, moving series between panes, adjusting pane height, and removing panes. The API ensures that developers have full control over the pane lifecycle and organization within their charts.
```
KatPierce, line 13: Sorry, maybe my suggestion was vague. 
This part about manipulating panes was good, let's bring it back to a separate paragraph. So, it will be:

> ## Managing panes
> 
> While the specific methods to manipulate panes are covered in the detailed [example](/tutorials/how_to/panes), it's important to note that Lightweight Charts provides an [API for pane management](/api/interfaces/IPaneApi.md). This includes adding new panes, moving series between panes, adjusting pane height, and removing panes. The API ensures that developers have full control over the pane lifecycle and organization within their charts.

website/docs/panes.md
```diff
@@ -0,0 +1,23 @@
+---
+sidebar_position: 3
+---
+
+# Panes
+
+Panes are essential elements that help segregate data visually within a single chart.
+
```
KatPierce, line 8: Let's remove the empty line between the first sentence and the next paragraph
```suggestion
```

KatPierce (approved):

website/docs/panes.md
```diff
@@ -0,0 +1,24 @@
+---
+sidebar_position: 3
+---
+
+# Panes
+
+Panes are essential elements that help segregate data visually within a single chart.
+Panes are useful when you have a chart that needs to show more than one kind of data. For example, you might want to see a stock's price over time in one pane and its trading volume in another. This setup helps users get a fuller picture without cluttering the chart.
+
+By default, Lightweight Charts has a single pane, however, you can add more panes to the chart to display different series in separate areas. For detailed examples and code snippets on how to implement panes in your charts [see tutorial](/tutorials/how_to/panes).
```
romfrancois, line 10:
```suggestion
By default, Lightweight Charts™ has a single pane, however, you can add more panes to the chart to display different series in separate areas. For detailed examples and code snippets on how to implement panes in your charts [see tutorial](/tutorials/how_to/panes).
```

Could you please add `™` next to all mention of `Lightweight Charts` ?

illetid: added, thanks

romfrancois (approved):

src/model/chart-model.ts
```diff
@@ -607,6 +613,20 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
 	}
 
 	public createPane(index?: number): Pane {
+		if (index !== undefined) {
+			assert(index >= 0, 'Index should be greater or equal to 0');
+			index = Math.min(this._panes.length, index);
+			if (index > this._panes.length) {
+				for (let i = this._panes.length; i < index; i++) {
+					this.createPane(i);
+				}
```
kirchet, line 622: 1. This condition is always false
2. Let's rename the method to `getOrCreatePane`
3. Seems that we can make this method `private`

kirchet (changes requested):

kirchet (approved):

SlicedSilver:
1. After an error of using `moveToPane(-1)` then you can no longer move the pane.
```js
// ...create a normal chart above...
volumeSeries = chart.addHistogramSeries(
    {
        priceFormat: {
            type: 'volume',
        },
    },
    1 // Pane index
);

volumeSeries.setData(generateLineData())

volumeSeries.moveToPane(2); // okay
volumeSeries.moveToPane(5); // okay
volumeSeries.moveToPane(-1); // error, but expected

volumeSeries.moveToPane(2); // error, but should work
```

---

2. Moving a series to a different price scale seems to also move the series to pane 0
```js
// ...create a normal chart above...
volumeSeries = chart.addHistogramSeries(
    {
        priceFormat: {
            type: 'volume',
        },
    },
    1 // Pane index
);

volumeSeries.setData(generateLineData())

// make left price scale visible
chart.applyOptions({leftPriceScale: {visible: true}})

// move volume series to left price scale
volumeSeries.applyOptions({priceScaleId: 'left'})

// !! now appears in pane 0

volumeSeries.moveToPane(1) // does nothing since it thinks it is in pane 1

volumeSeries.moveToPane(0) // hides empty pane 1
volumeSeries.moveToPane(1) // now the volume is back into the correct position
```

---

3. Using the above bug, it is possible to get an empty pane within the middle of the chart
```js
// ...create a normal chart above...
volumeSeries = chart.addHistogramSeries(
    {
        priceFormat: {
            type: 'volume',
        },
    },
    1 // Pane index
);

volumeSeries.setData(generateLineData())

// make left price scale visible
chart.applyOptions({leftPriceScale: {visible: true}})

// move volume series to left price scale
volumeSeries.applyOptions({priceScaleId: 'left'})

// !! now appears in pane 0

volumeSeries.moveToPane(2) // There is now an empty pane 1. Showing 3 panes
```

SlicedSilver:
[Minor] When dragging the pane separator, the time scale width can change. Maybe we would want to 'fix' the time scale width while dragging?

Let's only do this if it is easy, and lightweight to implement.

https://github.com/tradingview/lightweight-charts/assets/3482679/39e30ada-f9eb-4c0e-8339-68ef59572bbd

illetid:
> 1. After an error of using `moveToPane(-1)` then you can no longer move the pane.
> 
> ```js
> // ...create a normal chart above...
> volumeSeries = chart.addHistogramSeries(
>     {
>         priceFormat: {
>             type: 'volume',
>         },
>     },
>     1 // Pane index
> );
> 
> volumeSeries.setData(generateLineData())
> 
> volumeSeries.moveToPane(2); // okay
> volumeSeries.moveToPane(5); // okay
> volumeSeries.moveToPane(-1); // error, but expected
> 
> volumeSeries.moveToPane(2); // error, but should work
> ```
> 
> 2. Moving a series to a different price scale seems to also move the series to pane 0
> 
> ```js
> // ...create a normal chart above...
> volumeSeries = chart.addHistogramSeries(
>     {
>         priceFormat: {
>             type: 'volume',
>         },
>     },
>     1 // Pane index
> );
> 
> volumeSeries.setData(generateLineData())
> 
> // make left price scale visible
> chart.applyOptions({leftPriceScale: {visible: true}})
> 
> // move volume series to left price scale
> volumeSeries.applyOptions({priceScaleId: 'left'})
> 
> // !! now appears in pane 0
> 
> volumeSeries.moveToPane(1) // does nothing since it thinks it is in pane 1
> 
> volumeSeries.moveToPane(0) // hides empty pane 1
> volumeSeries.moveToPane(1) // now the volume is back into the correct position
> ```
> 
> 3. Using the above bug, it is possible to get an empty pane within the middle of the chart
> 
> ```js
> // ...create a normal chart above...
> volumeSeries = chart.addHistogramSeries(
>     {
>         priceFormat: {
>             type: 'volume',
>         },
>     },
>     1 // Pane index
> );
> 
> volumeSeries.setData(generateLineData())
> 
> // make left price scale visible
> chart.applyOptions({leftPriceScale: {visible: true}})
> 
> // move volume series to left price scale
> volumeSeries.applyOptions({priceScaleId: 'left'})
> 
> // !! now appears in pane 0
> 
> volumeSeries.moveToPane(2) // There is now an empty pane 1. Showing 3 panes
> ```

1. we are throwing an error and a developer who uses the library should handle the error
2. and 3 fixed in https://github.com/tradingview/lightweight-charts/pull/1557/commits/745e4bfac02cf29d62fa9bff43f962a4c79b954d

illetid:
> [Minor] When dragging the pane separator, the time scale width can change. Maybe we would want to 'fix' the time scale width while dragging?
> 
> Let's only do this if it is easy, and lightweight to implement.
> 
>  time-scale-width-changing.mov

@SlicedSilver could we move this to future improvements

SlicedSilver (approved):

SigmaU:
can't express how much I am waiting for this feature, guys! So grateful for everyones work ❤️

vishnuc:
I am excited for this feature , love you guys for this work :)

hems:
Can't wait for this : D

has anyone toyed around with doing a MACD panel ? : D

vishnuc:
I will be adding 50+ indicators to my project , once v5 is released , macd is one

PR #1591, opened 2024-05-10
replace day mark with month if month didn't fit

### LINKED ISSUE #1588: Time scale date render bug
**Lightweight Charts™ Version:**  4.1.3

Sometimes timescale displays a date as a number instead of a month. It happens more if there is a gaps in data timestamps
**Steps/code to reproduce:**

```js
function generateData(startValue) {
	const res = [];
	const time = new Date(Date.UTC(2018, 0, 1, 0, 0, 0, 0));
	for (let i = 0; i < 1000; ++i) {
		const item = {
			time: time.getTime() / 1000,
		};
		function getRandomInRange(min, max) {
			return Math.random() * (max - min) + min;
		}
                 // adding random gaps to increase the chance of the bug 
		time.setUTCDate(time.getUTCDate() + getRandomInRange(0, 9) + 1);

		generateBar(i, startValue, item);
		res.push(item);
	}
	return res;
}
const chart = LightweightCharts.createChart(document.getElementById('container'));
const mainSeries = chart.addBarSeries();
const startValue = Math.floor(container.getBoundingClientRect().height / 100) * 100;

mainSeries.setData(generateData(startValue));
```

**Actual behavior:**
As you can see on the screenshot we have displayed `11` instead of `Nov` 

**Expected behavior:**

**Screenshots:**

image

illetid:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #1588
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**
In some cases the month mark can't be placed on the timescale, but some day marks of the current month fit and we have strange behavior when we have on timescale `Jan Feb 21 Apr`. as a fix I've added a check for this specific case and change the weight of the mark before placing it into timescale

### DIFF
--- a/src/model/tick-marks.ts
+++ b/src/model/tick-marks.ts
@@ -1,6 +1,7 @@
 import { lowerBound } from '../helpers/algorithms';
 import { ensureDefined } from '../helpers/assertions';
 
+import { TickMarkWeight } from './horz-scale-behavior-time/types';
 import { InternalHorzScaleItem } from './ihorz-scale-behavior';
 import { TickMarkWeightValue, TimePointIndex, TimeScalePoint } from './time-data';
 
@@ -131,8 +132,26 @@ export class TickMarks<HorzScaleItem> {
 				}
 
 				if (rightIndex - currentIndex >= maxIndexesPerMark && currentIndex - leftIndex >= maxIndexesPerMark) {
-					// TickMark fits. Place it into new array
-					marks.push(mark);
+					// Check if we need to add month mark
+					if (weight === TickMarkWeight.Day && typeof mark.originalTime === 'number') {
+						const startOfTheMonth = new Date(mark.originalTime * 1000);
+						startOfTheMonth.setDate(1);
+						startOfTheMonth.setHours(0);
+						startOfTheMonth.setMinutes(0);
+						startOfTheMonth.setSeconds(0);
+						const startOfTheMonthUtcSeconds = Math.floor(startOfTheMonth.getTime() / 1000);
+						const monthMark = marks.find((m: TickMark) => {
+							return m.weight === TickMarkWeight.Month && (m.originalTime as number) >= startOfTheMonthUtcSeconds;
+						});
+						// If there is no month mark, but current day mark first, promote current mark to month mark
+						if (!monthMark) {
+							mark.weight = TickMarkWeight.Month as TickMarkWeightValue;
+						}
+						marks.push(mark);
+					} else {
+						// TickMark fits. Place it into new array
+						marks.push(mark);
+					}
 					leftIndex = currentIndex;
 				} else {
 					if (this._uniformDistribution) {
--- a/src/model/time-scale.ts
+++ b/src/model/time-scale.ts
@@ -237,7 +237,7 @@ export class TimeScale<HorzScaleItem> implements ITimeScale {
 	private _barSpacing: number;
 	private _scrollStartPoint: Coordinate | null = null;
 	private _scaleStartPoint: Coordinate | null = null;
-	private readonly _tickMarks: TickMarks<HorzScaleItem> = new TickMarks();
+	private readonly _tickMarks: TickMarks<HorzScaleItem> = new TickMarks<HorzScaleItem>();
 	private _formattedByWeight: Map<number, FormattedLabelsCache<HorzScaleItem>> = new Map();
 
 	private _visibleRange: TimeScaleVisibleRange = TimeScaleVisibleRange.invalid();


src/model/tick-marks.ts
```diff
@@ -131,8 +132,26 @@ export class TickMarks<HorzScaleItem> {
 				}
 
 				if (rightIndex - currentIndex >= maxIndexesPerMark && currentIndex - leftIndex >= maxIndexesPerMark) {
-					// TickMark fits. Place it into new array
-					marks.push(mark);
+					// Check if we need to add month mark
+					if (this._marksByWeight.has(TickMarkWeight.Month as TickMarkWeightValue) && weight === TickMarkWeight.Day && i !== currentWeightLength - 1 && typeof mark.originalTime === 'number') {
```
SlicedSilver, line 136: `mark.originalTime === 'number'` originalTime is not always going to be a number. In LWC, dates can be presented in 3 different ways: https://tradingview.github.io/lightweight-charts/docs/api#time

SlicedSilver (changes requested):
The changes work well on the test case, however I'm concerned about implementing this fix within the `tick-marks.ts` file. This fix is specific to the case where the horizontal scale is time based. So I think it would be better to try fix this within the `horz-scale-behavior-time.ts` and `time-scale-point-weight-generator.ts` files since that code is only used for a time based scale and it is only run when new data is provided instead of when the chart is scrolled or zoomed (therefore the performance should be better as well).

src/model/tick-marks.ts
```diff
@@ -131,8 +132,26 @@ export class TickMarks<HorzScaleItem> {
...
+						const startOfTheMonth = new Date(mark.originalTime * 1000);
+						startOfTheMonth.setDate(1);
+						startOfTheMonth.setHours(0);
+						startOfTheMonth.setMinutes(0);
+						startOfTheMonth.setSeconds(0);
+						const startOfTheMonthUtcSeconds = Math.floor(startOfTheMonth.getTime() / 1000);
```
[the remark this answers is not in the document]
illetid: it will make sense if there are lot of day marks, but we can have a 30 at max, and we still need to build a cache key from the timestamp for each item to check if the value in cache, so there is no profit of implementing memo

illetid:
> The changes work well on the test case, however I'm concerned about implementing this fix within the `tick-marks.ts` file. This fix is specific to the case where the horizontal scale is time based. So I think it would be better to try fix this within the `horz-scale-behavior-time.ts` and `time-scale-point-weight-generator.ts` files since that code is only used for a time based scale and it is only run when new data is provided instead of when the chart is scrolled or zoomed (therefore the performance should be better as well).

@SlicedSilver  I addressed all other issues but for this one. it probably make more sense to leave it here. due to all marks are valid and the issue dependent on zoom and/or width of the chart and this check need to be triggered on each buildTickMarks not only on data change and we can't fix it in `time-scale-point-weight-generator.ts`

src/model/horz-scale-behavior-time/time-utils.ts
```diff
@@ -0,0 +1,114 @@
...
+	return data.forEach(convertStringToBusinessDay);
+}
+
+export function fixMonthMarks(marks: readonly TickMark[], mark: TickMark): void {
+	const timePoint = convertTime(mark.originalTime as Time) as unknown as TimePoint;
+	const startOfTheMonth = new Date(timePoint.timestamp * 1000);
+	startOfTheMonth.setDate(1);
+	startOfTheMonth.setHours(0, 0, 0, 0);
+
+	const startOfTheMonthUtcSeconds = Math.floor(startOfTheMonth.getTime() / 1000);
```
SlicedSilver, line 103: If we use `.setHours(0, 0, 0, 0)` then the time should be even multiple of 1000 so we wouldn't need to use `floor` here. However, perhaps this will catch some edge case I'm not aware of so it is probably safer to keep it in.

src/model/tick-marks.ts
```diff
@@ -1,6 +1,8 @@
 import { lowerBound } from '../helpers/algorithms';
 import { ensureDefined } from '../helpers/assertions';
 
+import { fixMonthMarks } from './horz-scale-behavior-time/time-utils';
+import { TickMarkWeight } from './horz-scale-behavior-time/types';
```
SlicedSilver, line 5: I still feel that this file (`tick-marks.ts`) shouldn't 'know' about the type of horizontal scale. The behaviour should apply to all types of scales OR the logic for handling this issue should be within the `horz-scale-behavior-time` code.

As an example, if I implemented a custom horizontal scale and happened to use `50` and `60` as tick weights then this behaviour would be triggered even though my weights have nothing to do with months and days.

illetid:
hi @eugene-korobko, could you pls check this PR and the issue? what do you think about this approach, is it okay? as an alternative instead of changing day tickmark to month we could hide the tickmark entirely

src/model/tick-marks.ts
```diff
@@ -131,8 +132,26 @@ export class TickMarks<HorzScaleItem> {
 				}
 
 				if (rightIndex - currentIndex >= maxIndexesPerMark && currentIndex - leftIndex >= maxIndexesPerMark) {
-					// TickMark fits. Place it into new array
-					marks.push(mark);
+					// Check if we need to add month mark
+					if (weight === TickMarkWeight.Day && typeof mark.originalTime === 'number') {
```
eugene-korobko: `typeof mark.originalTime === 'number' ` check means the fix depends on initial date format, does it? I see no reason why should we make the difference

src/model/horz-scale-behavior-time/time-utils.ts
```diff
@@ -0,0 +1,114 @@
...
+
+	const monthMark = marks.find((m: TickMark) => {
+		return m.weight === TickMarkWeight.Month && (m.originalTime as number) >= startOfTheMonthUtcSeconds;
```
eugene-korobko, line 106: is `m.originalTime as number` conversion correct? Several lines above we call `convertTime(mark.originalTime as Time)`

lets check this code with different format of input data

illetid: Thanks for catching that, added `convertTime`

SlicedSilver (approved):

PR #1663, opened 2024-07-30
Extract watermark into plugin

SlicedSilver:
**Type of PR:** enhancement

**PR checklist:**

- [x] Includes tests
- [x] Documentation update

**Overview of change:**
- Removes the watermark feature from the core library and instead implements it as a built-in Pane Primitive (a.k.a plugin). This allows the watermark feature to be tree-shaken if not used.
- Two new plugins included with the library build: `TextWatermark` and `ImageWatermark`
- TextWatermark extends the behaviour of the old watermark feature by supporting multiple lines of text
- ImageWatermark adds the ability to easily add an image watermark.
- Both plugins work with the new Pane Primitive.


SlicedSilver:
> ## size-limit report 📦
> Path	Size
> ESM	43.07 KB (+1.44% 🔺)
> ESM createChart	41.76 KB (-1.12% 🔽)
> ESM createChartEx	40.49 KB (-0.99% 🔽)
> Standalone-ESM	0 B (-100% 🔽)
> Standalone	44.53 KB (+1.46% 🔺)
> ESM Standalone	44.53 KB (+100% 🔺)
> Plugin: Text Watermark	1.79 KB (+100% 🔺)
> Plugin: Image Watermark	1.7 KB (+100% 🔺)

FYI: I renamed `Standalone-ESM` to `ESM Standalone`

src/plugins/image-watermark/primitive.ts
```diff
@@ -0,0 +1,112 @@
...
+ */
+export class ImageWatermark<T = Time> implements IPanePrimitive<T> {
+	private _requestUpdate?: () => void;
+	private _paneViews: ImageWatermarkPaneView<T>[];
+	private _options: ImageWatermarkOptions;
+	private _imgElement: HTMLImageElement | null = null;
+	private _imageUrl: string;
+
+	public constructor(
+		imageUrl: string,
```
edew, line 50: What do you think about forcing the caller to handle loading the image and accepting an [ImageBitmap](https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmap) here instead?

SlicedSilver: Pros:
- More flexibility for the dev to provide images which generated. However for the more advanced cases we can recommend that they implement the plugin themselves. 🤷‍♂️
- Maybe a few bytes smaller.

Cons:
- More work for the dev, and the API is more confusing (so we need to explain more in the docs).

I went for the easy (less powerful) API.

edew: Makes sense, thanks. Happy to leave it as is.

edew (approved):

PR #1674, opened 2024-08-13
Add Yield Curve and Options Chart Types, Up/Down Markers Plugin

SlicedSilver:
**Type of PR:** new features

**PR checklist:**

- [x] Includes tests
- [x] Documentation update

**Overview of change:**

Add Yield Curve and Options Chart Types, Up/Down Markers Plugin, and Documentation Updates

#### New Features:
1. **Yield Curve Chart**:
   - Added a new chart type specifically designed for displaying yield curves.
   - Introduced the `createYieldCurveChart` function.
   - Implemented specialized horizontal scale behavior for yield curves, including linear spacing and whitespace handling.

2. **Options Chart**:
   - Added a new chart type that uses price values on the horizontal scale.
   - Introduced the `createOptionsChart` function.
   - Implemented horizontal scale behavior for price-based charts.

3. **Up/Down Markers Plugin**:
   - Added the `UpDownMarkersPrimitive` plugin to display markers indicating price changes.
   - Implemented marker management with automatic expiration and update handling.

4. **Codebase Improvements / API additions**:
   - Added the `ignoreWhitespaceIndices` option to the time scale configuration.
   - Enhanced the `updateSeriesData` function to support historical updates.

5. **Documentation**:
   - Updated the documentation to include new chart types and their usage.
   - Added a new page for chart types, detailing the creation and use cases for standard time-based, yield curve, and options charts.


src/plugins/up-down-markers-plugin/types.ts
```diff
@@ -0,0 +1,48 @@
...
+export interface MarkerCoordinates {
+	x: number;
+	y: number;
+	sign: number;
+}
+
+/**
+ * Defines the supported series types for markers.
+ */
+export type SupportedSeriesTypes = 'Line' | 'Area';
```
SlicedSilver, line 48: Since this is exported and thus included in the generated `d.ts` file lets give it a more explicit name

website/tutorials/demos/yield-curve-with-update-markers.js
```diff
@@ -0,0 +1,100 @@
...
+
+const series1 = chart.addLineSeries({
+	lineType: 2,
+	color: '#26c6da',
+	pointMarkersVisible: true,
+	lineWidth: 2,
+});
+const priceChangeMarkers = new UpDownMarkersPrimitive();
+series1.attachPrimitive(priceChangeMarkers);
+priceChangeMarkers.setData(curve1);
```
illetid, line 71: Not a review, just a question. Did you consider creating it using a custom series? From my point of view, managing the series via the series' primitive seems unintuitive. Since it’s a core plugin, we have access to all the source code, so we could potentially wrap the line and area series and add the up/down indicator logic on update. I haven’t thought much about it, so there might be other limitations with the custom series approach

SlicedSilver: As with most things it is a trade-off. In many ways it makes sense to create a custom series but there are some benefits to using a Primitive. The drawing of markers fits better into a primitive (conceptually), and the main reason for using a primitive is that it can be used on multiple types of series (and custom series) so we are not limiting this plugin to only working with one or two series which are copies of existing series types.
Finally, the up down markers plugin can be used without needing to use the `setData` / `update` passthrough methods if you use the manual `setMarkers` method. So my thinking is that using as a Primitive is a more flexible and lower level implementation which can make it useful in more use-cases.

PR #1687, opened 2024-08-29
Extract markers into plugin

illetid:
**Type of PR:** enhancement

- `N/A` ~Addresses an existing issue: fixes #~
- [x] Includes tests
- [x] Documentation update

**Overview of change:**
Moved a series markers into a built-in plugin to enable tree shaking


src/index.ts
```diff
@@ -26,6 +26,7 @@ export { createChart, createChartEx, defaultHorzScaleBehavior } from './api/crea
 */
 export { TextWatermark } from './plugins/text-watermark/primitive';
 export { ImageWatermark } from './plugins/image-watermark/primitive';
+export { createSeriesMarkersPrimitive } from './plugins/series-markers/primitive';
```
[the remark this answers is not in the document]
SlicedSilver: If we are going to use a `createSeriesMarkersPrimitive` function then I think it would be better if it returned an API interface instead of the actual primitive. At the moment the primitive exposes methods like `attached`, `detached`, `hitTest`, ... when it might make sense to only expose the `setMarkers` method.

So perhaps the best approach is to:
- export the primitive classes (so that people can extend if required)
- export `create...` functions for each plugin which adds the primitive, and provides a simplified wrapper API around the primitive which exposes only a few methods (like `setMarkers`, and `remove` which detaches the primitive).

SlicedSilver, line 29: Could we add some type checks tests within the `tests/type-checks` folder?

For example: `series-markers.ts`
```ts
import { createChart, createSeriesMarkersPrimitive } from '../../src';
import { Time } from '../../src/model/horz-scale-behavior-time/types';

const chart = createChart('anything');

const mainSeries = chart.addLineSeries();
mainSeries.setData([]);

const seriesMarkersPrimitive = createSeriesMarkersPrimitive(mainSeries, [
	{
		color: 'green',
		position: 'inBar',
		shape: 'arrowDown',
		time: 1556880900 as Time,
	},
]);

seriesMarkersPrimitive.setMarkers([
	{
		color: 'red',
		position: 'aboveBar',
		shape: 'arrowDown',
		time: 1556880900 as Time,
	},
]);

// and also test with a non-time-based series chart
```

Just to make sure that people using the library in typescript aren't going to get type errors when using it correctly. (See the other examples).

Currently I'm getting an error about not including `originalTime` in the marker object. This is marked as `@internal` so it won't be in the actual release and won't be an error but I don't think we need `originalTime` anyway. It could be an internal type of the primitive / view / renderer. And secondly I think we could remove it completely anyway (I don't see it being used).

src/model/autoscale-info-impl.ts
```diff
@@ -29,11 +29,8 @@ export class AutoscaleInfoImpl {
 	}
 
 	public toRaw(): AutoscaleInfo | null {
-		if (this._priceRange === null) {
-			return null;
-		}
 		return {
-			priceRange: this._priceRange.toRaw(),
+			priceRange: this._priceRange === null ? null : this._priceRange.toRaw(),
```
SlicedSilver, line 33: This return type for `toRaw()` is no longer correct. It now always returns `AutoscaleInfo`.

src/plugins/series-markers/primitive.ts
```diff
@@ -0,0 +1,229 @@
...
+import { SeriesMarkersPaneView } from './pane-view';
+import { InternalSeriesMarker, MarkerPositions, SeriesMarker } from './types';
+import {
+	calculateAdjustedMargin,
+	calculateShapeHeight,
+	shapeMargin as calculateShapeMargin,
+} from './utils';
+
+class SeriesMarkersPrimitive<HorzScaleItem> implements ISeriesPrimitive<HorzScaleItem> {
+	private _attached: SeriesAttachedParameter<HorzScaleItem> | null = null;
```
SlicedSilver, line 23: I don't think we need to keep a reference to this. We only use it to access `this._attached.series` which is also available from `this._series`

src/plugins/series-markers/primitive.ts
```diff
@@ -0,0 +1,229 @@
...
+				this._autoScaleMargins = null;
+			}
+
+			this._autoScaleMarginsInvalidated = false;
+		}
+
+		return this._autoScaleMargins;
+	}
+
+	protected _getMarkerPositions(): MarkerPositions {
```
SlicedSilver, line 125: Does this need to be `protected`? Seems like it could rather be marked as `private`

src/plugins/series-markers/types.ts
```diff
@@ -11,7 +13,7 @@ export type SeriesMarkerShape = 'circle' | 'square' | 'arrowUp' | 'arrowDown';
 /**
  * Represents a series marker.
  */
-export interface SeriesMarker<TimeType> {
+export interface SeriesMarker<TimeType = Time> {
```
SlicedSilver, line 16: Do we need to specify the default for the TimeType generic here? Hopefully it would be correctly inferred.

src/plugins/series-markers/utils.ts
```diff
@@ -48,3 +48,16 @@ export function calculateAdjustedMargin(margin: number, hasSide: boolean, hasInB
 
 	return 0;
 }
+export function convertSeriesMarker<InTimeType, OutTimeType>(sm: SeriesMarker<InTimeType>, newTime: OutTimeType, originalTime?: unknown): SeriesMarker<OutTimeType> {
```
SlicedSilver, line 51: Is this used?

src/index.ts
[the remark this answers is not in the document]
SlicedSilver: Having a look at the `d.ts` file, I'm not a fan of the Plugin classes having their private methods and fields displayed there (it's the same for the other plugins like the watermarks).

So I'm even more in favour of using `create` functions with a simplified public API wrapper around the primitive so that we can hide these implementation details. We are an open-source library so perhaps the benefit of exporting the classes so that they can be extended isn't really required because people can view the source anyway.

So what are your thoughts on no longer exporting the primitive classes for plugins and rather exposing a nice wrapper around them?
There could be a generic wrapper class which exposes some common things like: `detach` and maybe `applyOptions` and then each actual wrapper can extend that and add things like `setMarkers`, `markers` (to read), `series` (to get the attached series)...

I think these would be cleaner and less confusing.

SlicedSilver (commented):
- [x] Could we also add a short documentation page for the plugin?
- [x] I know the graphics tests will fail because the interface changed. Did you manage to run the tests locally before the changes, save those images somewhere, and then run again with the current changes? If so, could you check that the screenshots are the same? It's a bit of pain to manually do this I guess. Downside of our current screenshot testing system.

src/plugins/series-markers/utils.ts
illetid: removed, thanks

src/plugins/series-markers/types.ts
illetid: no, thanks for spotting this

src/plugins/series-markers/primitive.ts
illetid: changed

src/plugins/series-markers/primitive.ts
illetid: removed, thanks

src/index.ts
illetid: added type-checks, thanks, I tested only on javascript examples and miss that

.size-limit.js
```diff
@@ -8,31 +8,31 @@ export default [
...
+		limit: '44.50 KB',
 		import: '*',
 		ignore: ['fancy-canvas'],
 		brotli: true,
 	},
 	{
 		name: 'ESM createChart',
 		path: 'dist/lightweight-charts.production.mjs',
-		limit: '45.00 KB',
+		limit: '41.30 KB',
```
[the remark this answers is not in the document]
illetid: reverted, thanks for letting know, nice idea, changing it every time was horrible

src/index.ts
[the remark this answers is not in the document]
illetid: agree that exporting function instead of primitive is better from a user perspective, and if someone what to extend he can just copy from the sourcecode.
I tried to create a `create` functions, not sure about typings with this `WithOptions` stuff, could you take a looks pls. Also I updated ts to latests, because it didn't see an `applyOptions` methods even it worked in vscode, so bump version solved the issue.

.size-limit.js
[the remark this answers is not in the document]
illetid: and changed docs, we had a series markers page, so I just change the method calls there

src/plugins/pane-primitive-wrapper.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface IPanePrimitiveWithOptions<T, K> extends IPanePrimitive<T> {
+	/**
+	 * @param options - Options to apply. The options are deeply merged with the current options.
+	 */
+	applyOptions?: (options: DeepPartial<K>) => void;
+}
+
+export class PanePrimitiveWrapper<T, Options = unknown, TPrimitive extends IPanePrimitiveWithOptions<T, Options> = IPanePrimitive<T>> implements IPrimitiveWrapper<T, Options> {
+	protected _primitive: TPrimitive;
+	protected _pane: IPaneApi<T>;
```
SlicedSilver, line 20: Lets try use `private` where possible and only use `protected` if we need to.
```suggestion
	private _primitive: TPrimitive;
	private _pane: IPaneApi<T>;
```

src/plugins/image-watermark/primitive.ts
```diff
@@ -109,3 +94,31 @@ export class ImageWatermark<T = Time> implements IPanePrimitive<T> {
...
+ *   alpha: 0.5,
+ *   padding: 20,
+ * });
+ * // to change options
+ * imageWatermark.applyOptions({ padding: 10 });
+ * // to remove watermark from the pane
+ * imageWatermark.detach();
+ * ```
+ */
+export function createImageWatermark<T>(pane: IPaneApi<T>, imageUrl: string, options: DeepPartial<ImageWatermarkOptions>): PanePrimitiveWrapper<T, ImageWatermarkOptions, IPanePrimitiveWithOptions<T, ImageWatermarkOptions>> {
```
SlicedSilver, line 122: [This applies to all the `create...()` functions]
We should rather define the public API for the plugin wrapper by defining an interface like `ITextWatermarkPluginApi` which defines only the public methods so that the generated `d.ts` typings doesn't include any `private` methods.
I'll send you an example code snippet.

website/docs/migrations/from-v4-to-v5.md
```diff
@@ -13,16 +13,17 @@ In the new version of Lightweight Charts, the watermark feature has undergone si
 2. **Re-implementation**: It's now re-implemented as a Pane Primitive (plugin) included within the library.
 3. **Improved Tree-shaking**: This change makes the feature more tree-shakeable, potentially reducing bundle sizes for users who don't need watermarks.
 4. **Added an Image Watermark Primitive**: In addition to the usual text based watermark, there is now
-an image watermark feature provided by the [`ImageWatermark`](/api/classes/ImageWatermark.md) primitive.
+an image watermark feature provided by the [`createImageWatermark`](/api/functions/createImageWatermark.md) primitive.
```
SlicedSilver, line 16: Lets not call it a primitive here, maybe call it a plugin?

tests/type-checks/series-markers.ts
```diff
@@ -0,0 +1,122 @@
...
+]);
+
+markerApi.setMarkers([
+	{
+		color: 'red',
+		position: 'aboveBar',
+		shape: 'arrowDown',
+		time: 1556880900 as Time,
+	},
+]);
```
SlicedSilver, line 122: Using the wrong series here...
```suggestion
const timeValue: HorizontalScaleType = 12345;
const markerApi = createSeriesMarkers(lineSeries, [
	{
		color: 'green',
		position: 'inBar',
		shape: 'arrowDown',
		time: timeValue,
	},
]);

markerApi.setMarkers([
	{
		color: 'red',
		position: 'aboveBar',
		shape: 'arrowDown',
		time: 1556880900,
	},
]);
```

tests/type-checks/non-time-based-custom-series.ts
```diff
@@ -107,7 +107,6 @@ if (dataSet) {
 // eslint-disable-next-line @typescript-eslint/no-unsafe-call
 dataSet.push({ time: 12 });
 
-const textWatermark = new TextWatermark<HorizontalScaleType>({
+createTextWatermark<HorizontalScaleType>(chart.panes()[1], {
```
SlicedSilver, line 110: Lets have the test check that we don't need to specify the generic because it should be automatically inferred. We can use a `satisfies` to then ensure that the type is expected:
```ts
createTextWatermark(chart.panes()[1], {
	lines: [],
}) satisfies ITextWatermarkPluginApi<number>;

createTextWatermark(chart.panes()[1], {
	lines: [],
// @ts-expect-error Time is not the expected Generic here.
}) satisfies ITextWatermarkPluginApi<Time>;
```

src/index.ts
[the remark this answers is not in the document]
SlicedSilver: Thanks. I think the create functions are nicer 👍

src/plugins/image-watermark/primitive.ts
illetid: added suggested changes, now the typings looks more clean, thanks

SlicedSilver:
- There is a failing test within the 'coverage' e2e suite.
- Could you check the graphics e2e tests and ensure that all the tests are generating a `2. test.png` and that the image looks correct. You might need to run the tests locally on golden (v5-candidate) with these line removed: https://github.com/tradingview/lightweight-charts/blob/a20f0846139d10b0752b0d3be26a313a7ec78d2b/tests/e2e/graphics/graphics-test-cases.ts#L213 to get the correct baseline images and store them somewhere. Then run the e2e tests again but this time on this branch (and that line added back in), and then manually compare the images. Perhaps by having a script get the matching images for the baseline folder and the tests folder → place them into a new directory with naming like: `series-markers_test.png` and `series-markers_golden.png` and then you can quickly go through the images one by one and quickly notice if anything looks different. (mac quicklook and arrow keys work well for this)

illetid:
> * There is a failing test within the 'coverage' e2e suite.
> * Could you check the graphics e2e tests and ensure that all the tests are generating a `2. test.png` and that the image looks correct. You might need to run the tests locally on golden (v5-candidate) with these line removed: https://github.com/tradingview/lightweight-charts/blob/a20f0846139d10b0752b0d3be26a313a7ec78d2b/tests/e2e/graphics/graphics-test-cases.ts#L213
>    to get the correct baseline images and store them somewhere. Then run the e2e tests again but this time on this branch (and that line added back in), and then manually compare the images. Perhaps by having a script get the matching images for the baseline folder and the tests folder → place them into a new directory with naming like: `series-markers_test.png` and `series-markers_golden.png` and then you can quickly go through the images one by one and quickly notice if anything looks different. (mac quicklook and arrow keys work well for this)

I've checked all the screenshots. Fixed one small issue with margins that I've spotted in progress, and failing coverage tests are also fixed, I used wrong interface as returned type and ts-transformer renamed properties that should be public.
attached zip with tests results and added in each folder image from reference run to check with 
[devicePixelRatio=2.00.zip](https://github.com/user-attachments/files/16869847/devicePixelRatio.2.00.zip)

src/plugins/series-markers/primitive.ts
```diff
@@ -0,0 +1,189 @@
...
+				return {
+					priceRange: null,
+					margins: margins,
+				};
+			}
+		}
+		return null;
+	}
+
+	public autoScaleMargins(): AutoScaleMargins | null {
```
SlicedSilver, line 101: Could we make this `private`?

illetid: changed, thanks

SlicedSilver (approved):
:shipit:

PR #1704, opened 2024-10-01
fix blurry borders

illetid:
**Type of PR:** bugfix

**Overview of change:**
Fix blurry borders on Firefox

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -228,25 +228,25 @@ jobs:
           productionBuild: true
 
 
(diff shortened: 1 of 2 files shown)


SlicedSilver (approved):
Looks good. Works as expected on FF. 👍

Notes on the CI Checks:
- The single failing graphics test only has a single pixel difference and isn't a regression.
- Disabled the memleak test on master because it is stalling on the CI and timing out. It does work on the v5 branch so it's not a major concern.

PR #1719, opened 2024-11-04
Enhanced color space support

### LINKED ISSUE #1475: Support OKLCH color space
**Lightweight Charts™ Version:** 4.0.1, 4.1.1

The OKLCH color space has recently been introduced into all modern browsers, and while the colors work in some places, some places they do not.  Here is one example of such a place.

**Steps/code to reproduce:**

```js
var series = chart.addAreaSeries({
  topColor: 'oklch(0.6569 0.196 275.75)', // this is fine
  bottomColor: 'rgba(0, 150, 136, 0.04)',
  lineColor: 'oklch(0.6569 0.196 275.75)', // this is NOT fine
  lineWidth: 2,
});
```

**Actual behavior:**

```
Uncaught Error: Cannot parse color: oklch(0.6569 0.196 275.75)
    at colorStringToRgba ...
```

**Expected behavior:**

Handle the new OKLCH color space everywhere.

`colorStringToRgba` in `src/helpers/color.ts` seems to be where the conversion is done.

Here is a fork adding support, however since I have no knowledge around color spaces, the conversion code is directly from ChatGPT, and I don't feel comfortable submitting that as a PR.  I have tested it with a few colors, comparing to the values given by https://oklch.com/ and it does seem to function correctly, even for values outside of the RGB color space.

https://github.com/benphelps/lightweight-charts/tree/oklch

https://jsfiddle.net/ykpg8v7c/

SlicedSilver:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #1475 
- [x] Includes tests
- [x] Documentation update

**Note**
- I'll squash the commits when merging (since I accidentally duplicated some commits)
- ~The graphics e2e test runner will be fixed in a separate PR.~

**Overview of change:**
- Built-in sRGB-based colors (rgb, rgba, hex, named color, hsl) are now supported by the library without specific converters, or maps. Instead the rgb equivalent color string is determined by the browser by using `window.getComputedStyle(element).color`. For most charts, this will only be used for a few colours (like 2 or 3) and the performance effect is insignificant. However this does allow the bundle size to be decreased since we don't need to ship our own parsing converters.
- Custom color parsers can be provided to handle other color (non-sRGB) formats (like p3).
- Added an option to enable p3 expanded color gamut on the canvas (colorSpace)
- It is now possible to use any natively support CSS color format with the library (including expanded colors like display-p3), a custom color parser should be included in the chart options for these additional color formats so that the library can correctly generate contrast colors, and alpha colors as needed (using standard sRGB).

New options:
```ts
	/**
	 * Specifies the color space of the rendering context for the internal
	 * canvas elements.
	 *
	 * Note: this option should only be specified during the chart creation
	 * and not changed at a later stage by using `applyOptions`.
	 *
	 * @defaultValue `srgb`
	 *
	 * See [HTMLCanvasElement: getContext() method - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext#colorspace) for more info
	 */
	colorSpace: ColorSpace;

	/**
	 * Array of custom color parser functions to handle color formats outside of standard sRGB values.
	 * Each parser function takes a string input and should return either:
	 * - An {@link Rgba} array [r,g,b,a] for valid colors (with values 0-255 for rgb and 0-1 for a)
	 * - null if the parser cannot handle that color string, allowing the next parser to attempt it
	 *
	 * Parsers are tried in order until one returns a non-null result. This allows chaining multiple
	 * parsers to handle different color space formats.
	 *
	 * Note: this option should only be specified during the chart creation
	 * and not changed at a later stage by using `applyOptions`.
	 *
	 * The library already supports these color formats by default:
	 * - Hex colors (#RGB, #RGBA, #RRGGBB, #RRGGBBAA)
	 * - RGB/RGBA functions (rgb(), rgba())
	 * - HSL/HSLA functions (hsl(), hsla())
	 * - HWB function (hwb())
	 * - Named colors (red, blue, etc.)
	 * - 'transparent' keyword
	 *
	 * Custom parsers are only needed for other color spaces like:
	 * - Display P3: color(display-p3 r g b)
	 * - CIE Lab: lab(l a b)
	 * - LCH: lch(l c h)
	 * - Oklab: oklab(l a b)
	 * - Oklch: oklch(l c h)
	 * - ...
	 */
	colorParsers: CustomColorParser[];
```


src/model/colors.ts
```diff
@@ -0,0 +1,194 @@
...
+		// special case optimization
+		if (color === 'transparent') {
+			return color;
+		}
+
+		const originRgba = this._parseColor(color);
+		const originAlpha = originRgba[3];
+		return `rgba(${originRgba[0]}, ${originRgba[1]}, ${originRgba[2]}, ${
+			alpha * originAlpha
+		})`;
```
edew, line 119: Should this be added to _rgbaCache?

SlicedSilver: To add caching on the alpha would require something like this:
```ts
public applyAlpha(color: string, alpha: number): string {
		// special case optimization
		if (color === 'transparent') {
			return color;
		}
		const key = `${color}~${alpha}`;
		const cached = this._alphaCache.get(key);
		if (cached) {
			return cached;
		}

		const originRgba = this._parseColor(color);
		const originAlpha = originRgba[3];
		const result = `rgba(${originRgba[0]}, ${originRgba[1]}, ${originRgba[2]}, ${
			alpha * originAlpha
		})`;

		this._alphaCache.set(key, result);
		return result;
	}
```
in addition to reducing the number of possible alpha values like this:
```ts
private _color(seriesLineColor: string, stage: number, startAlpha: number, endAlpha: number): string {
		const alpha = startAlpha + (endAlpha - startAlpha) * stage;

		return this._series
			.model()
			.colorParser()
			.applyAlpha(
				seriesLineColor,
				// 64 is enough values for a smooth looking appearance while still making the calculation cache smaller
				Math.round(alpha * 64) / 64
			);
	}
```

So we are adding:
- string creation via template
- math operation (to limit cache size)
- and a cache check

and this would save doing:
- string creation via template for the result
- math operation (to calculate alpha)
- and a cache check (in parseColor)

So in the end we probably break even anyway. The current approach is less code so I'm thinking that we don't change it.

edew (approved):

PR #1720, opened 2024-11-06
addSeries test and website changes

illetid:
Fix tests and docs series code
**Type of PR:** enhancement

**PR checklist:**

- [x] Includes tests
- [x] Documentation update


SlicedSilver:
Can we add the new series exports as defined entries within: `website/tutorials/how_to/.eslintrc.js` so that we don't need to use `// eslint-disable-next-line no-undef` in files like `website/tutorials/how_to/two-price-scales.js`

illetid:
> Can we add the new series exports as defined entries within: `website/tutorials/how_to/.eslintrc.js` so that we don't need to use `// eslint-disable-next-line no-undef` in files like `website/tutorials/how_to/two-price-scales.js`

added series to globals, thanks

website/docs/migrations/from-v4-to-v5.md
```diff
@@ -3,6 +3,26 @@
...
+const chart = createChart(container, {});
+const lineSeries = chart.addSeries(LineSeries, { color: 'red' });
+```
+
+UMD:
+
+```js
+const chart = LightweightCharts.createChart(container, {});
+const lineSeries = chart.addSeries(LightweightCharts.LineSeries, { color: 'red' });
+```
```
SlicedSilver, line 24: Could we make this a bit more explicit? I feel that using the term SeriesDefinition might confuse some people (and that they would expect there to be an export named `SeriesDefinition`).

I've attached a md file with a suggestion.
[series-changes-section.md](https://github.com/user-attachments/files/17704849/series-changes-section.md)

illetid: changed, thanks for the suggestion

SlicedSilver (approved):

PR #1723, opened 2024-11-07
Graphics e2e tests syntax mode

SlicedSilver:
**Type of PR:** infrastructure

**Overview of change:**
Adds a mode for running the graphics test such that the golden case uses the test case code from the golden branch. This will be useful testing changes to API (such that the test case code is changed and can only run in the new branch).

Additionally fixes the graphics pipeline when running on the CI. Previously a single failing test would fail all the remaining test cases as well.

**How to run:**
Authorised users can trigger the pipeline on a specific branch with the `run_branch_specific_graphics_tests` parameter set to `true` through the CircleCI site.
Note: it will only work if the 'golden' branch has the code in this PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,11 @@
 
 version: 2.1
 
+parameters:
+  run_branch_specific_graphics_tests:
+    type: boolean
+    default: false
+
 aliases:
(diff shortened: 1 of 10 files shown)


edew (approved):
Nice. LGTM 👍

PR #1731, opened 2024-11-18
do full update on primitive detach

### LINKED ISSUE #1594: Chart isn't updated when primitive is detached using plugins-example
**Lightweight Charts™ Version:** 4.1.4 

Importing the code from the [band-indicator example](https://github.com/tradingview/lightweight-charts/blob/master/plugin-examples/src/plugins/bands-indicator/example/example.ts) and using in a React component there's an issue with detaching the primitive using a controlled state, as shown in the code below, the chart isn't updated after executing the function `detachPrimitive`.

The problem relies on the file `plugin-base.ts` that implements the method for detaching the primitive. The method doesn't call the `requestUpdate` function before destroying the instance data. If we add the line `this.requestUpdate();` before `this._requestUpdate = undefined;` the feature works as expected.

```ts
const colors = {
  backgroundColor: "black",
  lineColor: "#2962FF",
  textColor: "white",
  areaTopColor: "#2962FF",
  areaBottomColor: "rgba(41, 98, 255, 0.28)",
};

const bandIndicator = new BandsIndicator();

const ChartComponent = ({
  data,
  showBands,
}: {
  data: LineData[];
  showBands: boolean;
}) => {
  const chartContainerRef = useRef<HTMLDivElement>(null);
  const lineSeries = useRef<ReturnType<IChartApi["addLineSeries"]>>(null);

  useEffect(() => {
    const { backgroundColor, textColor } = colors;
    const handleResize = () => {
      chart.applyOptions({ width: chartContainerRef.current.clientWidth });
    };

    const chart = createChart(chartContainerRef.current, {
      layout: {
        background: { type: ColorType.Solid, color: backgroundColor },
        textColor,
      },
      width: chartContainerRef.current.clientWidth,
      height: 300,
    });

    lineSeries.current = chart.addLineSeries();
    lineSeries.current.setData(data);

    window.addEventListener("resize", handleResize);

    return () => {
      window.removeEventListener("resize", handleResize);
      chart.remove();
    };
  }, [data]);

  useEffect(() => {
    if (!lineSeries.current) return;
    if (showBands) {
      lineSeries.current.attachPrimitive(bandIndicator);
    } else {
      lineSeries.current.detachPrimitive(bandIndicator);
    }
  }, [showBands]);

  return <div ref={chartContainerRef} />;
};

const ChartWidget = (props: any) => {
  const [showBands, setShowBands] = React.useState(false);
  const sampleData = React.useMemo(() => generateLineData(), []);
  return (
    <>
      <ChartComponent data={sampleData} showBands={showBands} />
      <button onClick={() => setShowBands(!showBands)}>Toggle Bands</button>
    </>
  );
};
```

**Actual behavior:**

The chat isn't updated when the method `detachPrimitive` is executed.

**Expected behavior:**

The chart should update and remove the primitive draw.

https://codesandbox.io/p/sandbox/jolly-christian-lkmqnh

signatenkov:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #1594

**Overview of change:**

if primitive is detached chart will perform full update.

### DIFF
--- a/src/api/series-api.ts
+++ b/src/api/series-api.ts
@@ -242,6 +242,7 @@ export class SeriesApi<
 		this._series.detachPrimitive(primitive as ISeriesPrimitiveBase<unknown>);
 		if (primitive.detached) {
 			primitive.detached();
+			this._series.model().fullUpdate();
 		}
 	}
 


SlicedSilver:
@signatenkov Could we add an e2e graphics test case for this issue?
You could use one of the existing plugins e2e test cases, and then call `detachPrimitive`.
- On the old version, we would expect that the plugin is still visible, and on the new version that it is no longer visible.
- So the graphics test should be sufficient for testing this behaviour

signatenkov:
#1738

PR #1732, opened 2024-11-18
playground

signatenkov:
**Type of PR:** enhancement

**Overview of change:**

It's just basic vite server running blitzstack/codepen snippet against local build of lightweight-chart 

`npm run playground` to start playground

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -16,4 +16,5 @@
 /website/src/theme/**/*.js
 
 /plugin-examples
+/playground
 /packages/create-lwc-plugin
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
     "size-limit": "size-limit",
     "verify": "npm-run-all clean -p build:prod check-markdown-links -p lint check-dts-docs tsc-verify test size-limit -p type-tests",
     "test": "mocha tests/unittests/**/*.spec.ts",
-    "type-tests": "tsc -p ./tests/type-checks/tsconfig.composite.json --noEmit"
+    "type-tests": "tsc -p ./tests/type-checks/tsconfig.composite.json --noEmit",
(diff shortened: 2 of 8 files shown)


signatenkov:
@w0rse @SlicedSilver what do you think?

SlicedSilver:
The `v5-candidate` branch has a new `debug` folder which serves a similar purpose: https://github.com/tradingview/lightweight-charts/tree/v5-candidate/debug

I would suggest having a look at that. And if you want to make some improvements then lets do it on top of that folder.

signatenkov:
@SlicedSilver I've checked and my implementation provides debug environment similar to what codepen/blitzstack provides (modules, react, typescript, live reload) which in my opinion works better for development

Implementation is v5 if I understood it right more aimed towards users of lwc rather than devs

Perhaps we can combine both solutions under "debug" folder

SlicedSilver:
> @SlicedSilver I've checked and my implementation provides debug environment similar to what codepen/blitzstack provides (modules, react, typescript, live reload) which in my opinion works better for development
> 
> Implementation is v5 if I understood it right more aimed towards users of lwc rather than devs
> 
> Perhaps we can combine both solutions under "debug" folder

👍 Cool, lets change the merge target to `v5-candidate` and add the `playground` folder within `debug` folder. It then allows the dev to pick whichever approach they prefer and keeps all development related test files within a single folder.

Also make sure that the `.gitignore` will ignore any files used / created by the dev. So rename `main.ts` → `main.ts.example` and then make it ignore any ts, js, etc... files within that folder.

signatenkov:
@SlicedSilver yeah I already handled gitignore by ignoring whole src folder and package.json so you can modify files directly without worrying that they might leak into the commit

PR #1733, opened 2024-11-18
Fix priceLine rendering when having LastPriceAnimation indicator

### LINKED ISSUE #1711: LastPriceAnimation indicator overlaps priceLine axis label
**Lightweight Charts™ Version:** 4.1.1

Add a price line with axis label 
Add LastPriceAnimationMode.OnDataUpdate and see the pulsing dot indicator overlaps the priceLine axis label

signatenkov:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #1711

Note:
- Not adding an e2e graphics test case because it is unreliable to test the Last Price Animation during it the timing of the screenshot within the animation effect.

### DIFF
--- a/src/gui/pane-widget.ts
+++ b/src/gui/pane-widget.ts
@@ -506,6 +506,7 @@ export class PaneWidget implements IDestroyable, MouseEventHandlers {
 			});
 			this._drawCrosshair(topTarget);
 			this._drawSources(topTarget, sourceTopPaneViews);
+			this._drawSources(topTarget, sourceLabelPaneViews);
 		}
 	}
 


SlicedSilver:
Could we change the merge target to `v5-candidate`?

I think at this stage it is easier to just work with the v5 branch because it is unlikely there will be another v4 release. If we merge this into master then we also need to cherry pick it into the v5 branch anyway. So it is easier to just put it into v5, and then we don't need to worry about maintaining two branches and ensuring we haven't forgotten to move something across.

signatenkov:
v5: #1737

PR #1735, opened 2024-11-18
debug/playground

signatenkov:
**Type of PR:** enhancement 

as discussed I moved playground into debug folder on v5-candidate

Old PR: #1732

### DIFF
--- a/debug/playground/package.json
+++ b/debug/playground/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "lwc-playground",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "dependencies": {
+    "eslint": "^9.15.0",
+    "typescript": "^5.6.3",
+    "vite": "^5.4.11"
+  },
+  "scripts": {
+    "dev": "/bin/sh ./playground.sh",
+    "dev:clean": "/usr/bin/env CLEAN=1 /bin/sh ./playground.sh"
+  }
+}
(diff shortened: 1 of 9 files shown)


signatenkov:
@SlicedSilver make it in v5-candidate

SlicedSilver:
I think we could simplify this by doing the following:
- specify lightweight-charts within the package.json like this: `"lightweight-charts": "file:../.."` so that we don't need to use `npm link`
- change the `dev` script in the `package.json` to `"dev": "vite",`
- remove the `playground.sh` file
- move `debug/playground/wwwroot/*` to `debug/playground/` (so we don't need the `wwwroot` folder)

We have actually got something similar in the `plugins-examples` folder (we can skip the building part).

---

Anyway, this approach is probably nicer than the esm files in the debug folder. So perhaps you could move all of these playground files into the debug folder itself but still keep the `debug-standalone.html` file (because it would still be useful to test with the standalone build as well).

debug/playground/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "lwc-playground",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "dependencies": {
+    "eslint": "^9.15.0",
```
SlicedSilver, line 7: To get this working on a fresh install, I needed to add this:
```suggestion
  "dependencies": {
    "lightweight-charts": "file:../..",
    "eslint": "^9.15.0",
```

Issue #1312, opened 2023-03-21
Support for limiting min/max zoom/scale level


ljuborados:
**Is your feature request related to a problem? Please describe.**

We have a somewhat basic requirement - limiting how much the chart can zoom, 
e.g. this universally looks bad and we should never allow zooming into that much detail: 

<img width="507" alt="image" src="https://user-images.githubusercontent.com/18645781/226592110-f62464a1-2b11-4d54-b44d-9f614a9adffe.png">

**Describe the solution you'd like**

Ideally a config option limiting max diff between from & to - either in logical representation or in time diff.
Less ideally, being able to return `null` or `false` from `subscribeVisibleTimeRangeChange` to prevent changing the visible time range.

**Additional context**

Tried calling `setVisibleRange` from inside `subscribeVisibleTimeRangeChange` but unfortunately UI 'jumps' to the scrolled to position first, and only after applies the range in `setVisibleRange` which comes with obvious UX issues.

Willing to lend a hand in a PR if this is something you guys would consider adding.

Thanks!

SlicedSilver:
Linking related issue: #1279

Would it be right to summarise your request as the ability to set the following options:
- Minimum Bar Spacing
- Maximum Bar Spacing

Bar spacing is the gap between the centre points of bars. It is essentially the 'zoom' level. So Essentially, options for max and min zoom. 

Is there anything else?

ljuborados:
Absolutely would, that would handle the use-case.

Thanks!

PR #1740, opened 2024-11-21
Max bar spacing option v5

### LINKED ISSUE #1279: Supporting of Max Bar Spacing
Does the library support the ability to set maxBarSpacing?
Based on the documentation, I only see minBarSpacing. If the library does not support this, are there alternative ways to set maxBarSpacing?

signatenkov:
**Type of PR:** enhancement

**PR checklist:**

- [ ] Addresses an existing issue: fixes #1279 #1312
- [ ] Includes tests
- [ ] Documentation update


src/model/time-scale.ts
```diff
@@ -845,19 +854,23 @@ export class TimeScale<HorzScaleItem> implements ITimeScale {
...
-
-		if (this._barSpacing < minBarSpacing) {
+		const maxBarSpacing = this._maxBarSpacing();
+		if (maxBarSpacing > 0 && this._barSpacing > maxBarSpacing) {
+			this._barSpacing = maxBarSpacing;
+			this._visibleRangeInvalidated = true;
+		} else if (this._barSpacing < minBarSpacing) {
 			this._barSpacing = minBarSpacing;
 			this._visibleRangeInvalidated = true;
 		}
```
SlicedSilver, line 864: We could probably simplify this by using the `clamp` function from `src/helpers/mathex.ts` to get an 'adjusted value' and then check with only a single `if` if it has changed from the current value.

tests/e2e/coverage/test-cases/time-scale/bar-width.js
```diff
@@ -28,10 +34,25 @@ function beforeInteractions(container) {
...
+		maxBarSpacing: 24,
+	});
+
+	await awaitNewFrame();
+
+	chart.timeScale().applyOptions({
+		minBarSpacing: 2,
+		maxBarSpacing: 1,
+	});
+
```
SlicedSilver, line 55: I would prefer that we don't modify existing test cases but rather add new test cases. It is better to have more test cases then to try test multiple things in a single test.

For example, a separate test for the logic when min is larger than max for the bar spacing options.

tests/e2e/graphics/test-cases/time-scale/max-bar-spacing.js
```diff
@@ -0,0 +1,41 @@
...
+	const chart = window.chart = LightweightCharts.createChart(container, {
+		timeScale: {
+			minBarSpacing: 10,
+			maxBarSpacing: 10,
+		},
+		layout: { attributionLogo: false },
+		width: 500,
+	});
+
+	const mainSeries = chart.addBarSeries();
```
SlicedSilver, line 35: This is the old v4 syntax. It will fail the test with a syntax / runtime error.

It needs to use the new `chart.addSeries(BarSeries)` syntax

SlicedSilver (changes requested):

PR #1758, opened 2024-12-17
CLL-314: add prototype pollution protection for merge helper function

SlicedSilver:
**Type of PR:** bugfix

**PR checklist:**

- [x] Includes tests

**Overview of change:**
Prevents prototype pollution via the internal `merge` helper function in the unlikely case that user input or api responses are used directly for options and have been maliciously compromised.

### DIFF
--- a/tests/unittests/helpers.spec.ts
+++ b/tests/unittests/helpers.spec.ts
@@ -1,0 +1,173 @@
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+
+import { clone, merge } from '../../src/helpers/strict-type-checks';
+
+describe('Helpers', () => {
+	/* eslint-disable
+		@typescript-eslint/unbound-method,
+		@typescript-eslint/no-explicit-any,
+		@typescript-eslint/no-unsafe-member-access,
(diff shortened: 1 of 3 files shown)


tests/unittests/helpers.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+				g: 5,
+			});
+		});
+
+		it('should handle arrays correctly', () => {
+			const dst = { arr: [1, 2, 3] };
+			const src = { arr: [4, 5, 6] };
+
+			const result = merge(dst, src);
+			expect(result.arr).to.deep.equal([4, 5, 6]);
```
edew, line 54: I know this is asserting the existing behaviour but I would have assumed the result would be `[1,2,3,4,5,6]`?

edew (approved):

tests/unittests/helpers.spec.ts
SlicedSilver: Yeah, in our case we wouldn't want to be merging the contents of an array. However, I'm not sure we even use arrays in our options anywhere (off the top of my head). The correct behaviour we want is that an array overrides instead of merges it's contents

PR #1768, opened 2025-01-02
fix point markers out of the view

### LINKED ISSUE #1755: drawSeriesPointMarkers function does not work correctly
**Lightweight Charts™ Version:** 4.2.2

**Steps/code to reproduce:**

1. Add two series of data (one must start later than the other)
2. Add pointMarkersVisible = true for shorter series
3. Scroll until the shorter series goes out of view

**Actual behavior:**

Black fill appears

**Expected behavior:**

Black fill does not appears

**Screenshots:**

https://github.com/user-attachments/assets/2044e014-4a9c-479d-9409-0ece95aa167c

**JSFiddle link:**

https://jsfiddle.net/x9yh56qm/

I think this behavior occurs because the **drawSeriesPointMarkers** function always does **context.fill()**, even if the series is out of scope

illetid:
**Type of PR:**  bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #1755
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

### DIFF
--- a/src/renderers/draw-series-point-markers.ts
+++ b/src/renderers/draw-series-point-markers.ts
@@ -13,6 +13,10 @@ export function drawSeriesPointMarkers<TItem extends LinePoint, TStyle extends C
 	// so if styleGetter returns objects, then styleGetter should return the same object for equal styles
 	styleGetter: (renderingScope: BitmapCoordinatesRenderingScope, item: TItem) => TStyle
 ): void {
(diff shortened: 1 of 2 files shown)


SlicedSilver (approved):
LGTM 👍

Seems that we were always calling `fill()` (at the bottom of the function) even if the for loop never drew anything. Good catch 👍

PR #1774, opened 2025-01-15
CLL-303: Assert that open, close, high, low, value can safely represe…

### LINKED ISSUE #1500: Browser freezes when updating chart data.
**Lightweight Charts™ Version:** 4.1.2

**Steps/code to reproduce:**

```js
const chart = LightweightCharts.createChart(document.getElementById("chart"));
let series = chart.addLineSeries();
series.update({
	"time": "2023-12-27",
	"value": 459761321634093
});
```

**Actual behavior:**

Browser freezes.

**Expected behavior:**

**Screenshots:**

https://jsfiddle.net/L3rvudpc/

signatenkov:
**Type of PR:** Fixes #1500 

**Overview of change:**

This PR adds validation for values in time series to prevent issue with browser freezing.
Browser freezing happens when price value cannot be safely represented as decimal with 2 digits after the dot.

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -79,7 +79,7 @@ export default [
 		path: 'dist/lightweight-charts.production.mjs',
 		import: '{ createSeriesMarkers }',
 		ignore: ['fancy-canvas'],
-		limit: '4.123 KB',
+		limit: '4.5 KB',
 		brotli: true,
 	},
 	{
@@ -119,7 +119,7 @@ export default [
 		path: 'dist/lightweight-charts.production.mjs',
 		import: '{ CandlestickSeries }',
 		ignore: ['fancy-canvas'],
-		limit: '2.54 KB',
+		limit: '3.00 KB',
 		brotli: true,
 	},
 	{
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -79,7 +79,7 @@ export default [
 		path: 'dist/lightweight-charts.production.mjs',
 		import: '{ createSeriesMarkers }',
 		ignore: ['fancy-canvas'],
-		limit: '4.08 KB',
+		limit: '4.123 KB',
 		brotli: true,
 	},
 	{
@@ -119,7 +119,7 @@ export default [
 		path: 'dist/lightweight-charts.production.mjs',
 		import: '{ CandlestickSeries }',
 		ignore: ['fancy-canvas'],
-		limit: '2.5 KB',
+		limit: '2.54 KB',
 		brotli: true,
 	},
 	{
--- a/src/model/data-validators.ts
+++ b/src/model/data-validators.ts
@@ -115,8 +115,8 @@ function checkCustomItem(
 	return;
 }
 
-const MIN_SAVE_VALUE = Number.MIN_SAFE_INTEGER / 100;
-const MAX_SAVE_VALUE = Number.MAX_SAFE_INTEGER / 100;
+const MIN_SAFE_VALUE = Number.MIN_SAFE_INTEGER / 100;
+const MAX_SAFE_VALUE = Number.MAX_SAFE_INTEGER / 100;
 
 function isSafeValue(value: number): boolean {
 	return value >= MIN_SAVE_VALUE && value <= MAX_SAVE_VALUE;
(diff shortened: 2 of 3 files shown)


SlicedSilver (changes requested):

.size-limit.js
```diff
@@ -119,7 +119,7 @@ export default [
 		path: 'dist/lightweight-charts.production.mjs',
 		import: '{ CandlestickSeries }',
 		ignore: ['fancy-canvas'],
-		limit: '2.5 KB',
+		limit: '2.54 KB',
```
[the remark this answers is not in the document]
signatenkov: I think for simplicity let's not overthink it and just keep quantised to 0.5 so `3.00 KB` in this case

.size-limit.js
```diff
@@ -79,7 +79,7 @@ export default [
 		path: 'dist/lightweight-charts.production.mjs',
 		import: '{ createSeriesMarkers }',
 		ignore: ['fancy-canvas'],
-		limit: '4.08 KB',
+		limit: '4.123 KB',
```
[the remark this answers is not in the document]
signatenkov: Set to `4.5 KB`

src/model/data-validators.ts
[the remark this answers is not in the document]
signatenkov: Sure thing, it's just a bit weird that it can't handle function defined and used only locally...

```
grep isSafeValue ./dist/* | cut -d: -f1 | sort | uniq
./dist/lightweight-charts.development.mjs
./dist/lightweight-charts.standalone.development.js
./dist/lightweight-charts.standalone.development.mjs
```

and actually it handles it well, but fails on other thing which is `VALUE_SAFE_RANGE`, seems like it doesn't know how to handle template strings.

Actually the only reason I've created VALUE_SAFE_RANGE is because I didn't want to stress production build with extra computations.


ps. here is how I investigated the thing
```
index 7a2563809..66ac597cc 100644
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -57,6 +57,9 @@ function getConfig(inputFile, { format, isProd, isStandalone }) {
 					// eslint-disable-next-line camelcase
 					inline_script: true,
 				},
+				compress: false,
+				// eslint-disable-next-line camelcase
+				keep_fnames: true,
 				mangle: {
 					module: format === 'esm',
 					properties: {
```

signatenkov:
I've also changed `npm run size-limit` a bit so it now highlights when bundle is less than a limit + it would build prod automatically in case if there is none or after changing a branch

src/model/data-validators.ts
```diff
@@ -117,3 +114,10 @@ function checkCustomItem(
 	// Nothing to check yet...
 	return;
 }
+
+const MIN_SAVE_VALUE = Number.MIN_SAFE_INTEGER / 100;
+const MAX_SAVE_VALUE = Number.MAX_SAFE_INTEGER / 100;
```
edew, line 119:
```suggestion
const MIN_SAFE_VALUE = Number.MIN_SAFE_INTEGER / 100;
const MAX_SAFE_VALUE = Number.MAX_SAFE_INTEGER / 100;
```

edew (changes requested):

src/model/data-validators.ts
signatenkov: @edew I know it hurts, sorry about that!

SlicedSilver: 🙈 I thought I commented on that already.

Thanks for catching it @edew

src/model/data-validators.ts
```diff
@@ -108,6 +98,13 @@ function checkLineItem<HorzScaleItem>(
 			lineItem.value
 		}`
 	);
+
+	assert(
+		isSafeValue(lineItem.value),
+		`${type} series item data value must be between ${MIN_SAVE_VALUE.toPrecision(16)} and ${MAX_SAVE_VALUE.toPrecision(16)}, got=${typeof lineItem.value}, value=${
```
SlicedSilver, line 104:
```suggestion
		`${type} series item data value must be between ${MIN_SAFE_VALUE.toPrecision(16)} and ${MAX_SAFE_VALUE.toPrecision(16)}, got=${typeof lineItem.value}, value=${
```

signatenkov (commented):
Yeah, I've fixed SAVE/SAFE in other places straight after the first commit, but pushed just now.

PR #1776, opened 2025-01-20
don't draw crosshair lines after mouse leaves chart pane

SlicedSilver:
**Type of PR:** bugfix (regression added during v4 to v5 changes)

**PR checklist:**

- [ ] Addresses an existing issue: (no issue created)
- [x] Includes tests
- `N/A` Documentation update

**Overview of change:**
When the mouse (pointer) leaves the chart pane then it is expected that the chart should stop drawing the crosshair.

A one-line code change to trigger a `updateAllViews` was all that was required to fix the issue.


illetid (approved):
🚀

PR #1790, opened 2025-01-30
ensure resize observer uses last entry

SlicedSilver:

**Overview of change:**
The issue with using ⁠`find()` instead of the last entry is that `ResizeObserver` may batch multiple resize notifications for the same element in a single callback. The last entry in the array represents the most recent size change, making it more accurate for the current state. Using `⁠find()` could potentially return an outdated intermediate size value.

Additionally, since this usage of `ResizeObserver` is configured to observe only one element (`this._container`), there's no need to search for the matching entry - the callback will only receive entries for that element, making ⁠`entries[entries.length - 1]` the most efficient approach.


edew (commented):
Nice 👍

PR #1800, opened 2025-02-04
suppress ResizeObserver error overlay in dev mode

SlicedSilver:
**Note: This doesn't affect the library code, or built version of the documentation. Purely for development mode of the website.**

Suppresses the "ResizeObserver loop completed with undelivered notification" error overlay that appears during development. This is a harmless error that occurs during React development mode and doesn't affect production builds.

The error will still be logged to the console but won't disrupt development by showing the error overlay.


romfrancois (approved):
LGTM - joy of using React :)

PR #1818, opened 2025-02-17
updated version of the sandbox

signatenkov:
This PR replaces existing `/debug` folder with alternative version for more convenient and streamlined experimentation with lightweight-charts.

The main difference with my first attempt is focus on simplicity, ability to have unlimited amount of sandboxes.

Performance considerations was also taken into account and every sandbox is sharing the same common root `/debug/playground` and their packages. Any dependency managed under sandbox would happen under common root.


debug/README.md
SlicedSilver, line 1: Could we add something about how to test using the production (minified) version of the library?
Since we use the following in our library package:
```json
"exports": {
    "./package.json": "./package.json",
    ".": {
      "development": {
        "types": "./dist/typings.d.ts",
        "import": "./dist/lightweight-charts.development.mjs"
      },
      "production": {
        "types": "./dist/typings.d.ts",
        "import": "./dist/lightweight-charts.production.mjs"
      },
      "default": {
        "types": "./dist/typings.d.ts",
        "import": "./dist/lightweight-charts.production.mjs"
      }
    }
  },
```
It means all that we need to do to tell vite to use the production version is to set the `NODE_ENV` to `production`.

So if I want to run my playground with the production then I just do this:
```bash
NODE_ENV=production npm run serve my-playground
```

debug/default/index.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const chartOptions = {
+	layout: {
+		textColor: "black",
+		background: {
+			type: "solid",
+			color: "white",
+		},
+	},
+};
```
SlicedSilver, line 15: I got a typescript warning when I opened the file in my playground.
Lets add some types to make TS happy.
```suggestion
import {
	AreaSeries,
	CandlestickSeries,
	ChartOptions,
	ColorType,
	createChart,
	DeepPartial,
} from "lightweight-charts";

const chartOptions = {
	layout: {
		textColor: "black",
		background: {
			type: ColorType.Solid,
			color: "white",
		},
	},
} satisfies DeepPartial<ChartOptions>;
```

SlicedSilver (commented):
This is really cool 🤩 

Very simple and fast to create new examples. Great work.

signatenkov:
@SlicedSilver Thanks for the review, I've added `serve:prod` to serve sandbox using production build.

SlicedSilver (approved):

signatenkov:
@SlicedSilver `size-limit` fail seems unrelated

SlicedSilver:
> @SlicedSilver `size-limit` fail seems unrelated

Yes, it is fixed in #1823

PR #1823, opened 2025-02-20
Set cursor on primitives

illetid:
**Type of PR:**  bugfix

**PR checklist:**

- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**
Fix the issue with primitive's `cursorStyle`. Added it to the hovered source instead of checking hit test on mouse move to fix the issue when mouse on the same spot and we change price/time scale (video of the bug is attached) 

https://github.com/user-attachments/assets/883e251e-ee7c-4113-aea7-ed31ba2fc52a

### DIFF
--- a/src/model/chart-model.ts
+++ b/src/model/chart-model.ts
@@ -173,6 +173,7 @@ export interface HoveredObject {
 export interface HoveredSource {
 	source: IPriceDataSource | IPrimitiveHitTestSource;
 	object?: HoveredObject;
+	cursorStyle?: string | null;
 }
 
 export interface PriceScaleOnPane {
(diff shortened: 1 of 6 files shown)


src/model/chart-model.ts
```diff
@@ -796,7 +797,8 @@ export class ChartModel<HorzScaleItem> implements IDestroyable, IChartModelBase
 
 		this.cursorUpdate();
 		if (!skipEvent) {
-			this._updateHoveredSourceOnChange(pane, x, y);
+			const hitTest = hitTestPane(pane, x, y);
+			this.setHoveredSource(hitTest && { source: hitTest.source, object: hitTest.object, cursorStyle: hitTest.cursorStyle || null });
```
itumanov, line 801: @illetid those 2 lines are causing major performance issues on my side (I did my custom build locally without them and it's fine). Any way to optimize it without breaking anything?
image

itumanov: fixed by excluding hover detection from "pressedMouseMove" event, since it doesn't make any sense, because mouse position relative to pane doesn't change

PR #1826, opened 2025-02-27
Add Price-Based Positioning for Series Markers

### LINKED ISSUE #1508: Markers should have a price value/y-coordinate to point to the correct level of a buy/sell on the y axis (with positioning inBar)

Covered in the following issues:
https://github.com/tradingview/lightweight-charts/issues/636

"Arrows should point correctly at the pricelevel of buy/sell. The down-tip of the arrowDown should be placed at the price-level of the sell. The upwards tip of the arrowUp should be placed at the pricelevel of the buy.

I expect the arrow to point to the correct price level if positioning is "inBar", or expect that there would be another positioning option that would make the marker handle this case."

**Additional context**

In below example, current behavior results in a marker below $71, even though a sell occurred at $71. 

image

### LINKED ISSUE #1648: Markers @ price
Currently markers can only be set above, below or at the candle. This can lead to misleading visualization. 

I would like to be able to specify exactly where I want the marker.  It can be within the candle itself or outside.

EranGrin:
# Add Price-Based Positioning for Series Markers

This PR adds a new feature to series markers that allows them to be positioned at exact price levels, providing more precise control over marker placement compared to the traditional bar-relative positioning.

Closes #1648
Closes #1508

## Changes

- Added three new position types for markers:
  - `atPriceTop` - Places the marker so its bottom edge is exactly at the specified price level
  - `atPriceBottom` - Places the marker so its top edge is exactly at the specified price level
  - `atPriceMiddle` - Places the marker centered at the specified price level

- Added support for the `price` property in markers to specify the exact price level
- Implemented fallback behavior when price is missing or cannot be calculated
- Fixed sizing issues with markers when using different size values
- Improved text positioning for markers to maintain consistent spacing
- Added comprehensive documentation for the new feature
- Added tests to verify the functionality

## Implementation Details

The implementation maintains full backward compatibility with the existing API while adding new functionality:

1. The `positionMarkerAtPrice` function handles the price-based positioning logic
2. The `fillSizeAndY` function prioritizes price-based positioning when applicable
3. Fallback behavior ensures graceful degradation when price data is missing
4. Text positioning follows the same logic as the traditional positioning

## Documentation

- Added a new documentation file for series markers
- Updated the release notes to include information about the new feature
- Added JSDoc comments to the code

## Testing

- Added a new test case for price-based positioning
- Tested with different marker sizes to ensure consistent appearance
- Verified fallback behavior works as expected

### DIFF
--- a/src/plugins/series-markers/pane-view.ts
+++ b/src/plugins/series-markers/pane-view.ts
@@ -17,7 +17,7 @@ import {
 	SeriesMarkerRendererDataItem,
 	SeriesMarkersRenderer,
 } from './renderer';
-import { InternalSeriesMarker } from './types';
+import { InternalSeriesMarker, SeriesMarkerPosition, SeriesMarkerPricePosition } from './types';
 import {
 	calculateShapeHeight,
 	shapeMargin as calculateShapeMargin,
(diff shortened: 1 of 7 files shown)


SlicedSilver:
@EranGrin Thanks for the pull request. We will have a look and review this next week.

SlicedSilver:
@EranGrin I've pushed some changes into the PR which simplifies the code and reduces the amount of line changes.
- The typings have changed such that `price` is required if you use one of the new Price Based Positions.
- Removed the changes to the sizing of markers. I think this can be tackled in a separate PR.
- Simplified the logic in the Primitive. The library will already refresh everything when required so it wasn't required to add the listeners for mouse, price-scale, and data changes.
- Simplified the pane view logic since the main required change is selecting a different price based on the `position` and removing the offset in those cases.

Could you please check that your use case still works with these changes?

src/plugins/series-markers/pane-view.ts
```diff
@@ -32,7 +32,32 @@ interface Offsets {
...
+			return seriesData.close;
+		}
+		if (marker.position === 'aboveBar') {
+			return seriesData.high;
+		}
+		if (marker.position === 'belowBar') {
+			return seriesData.low;
+		}
+	}
+	return;
```
EranGrin, line 57: Might be better to return `undefined`

EranGrin (commented):
Looks promising, I'll do some local tests tmr.

src/plugins/series-markers/pane-view.ts
SlicedSilver: `return;` and `return undefined;` are exactly the same thing and when it is minified they will both be converted into the same code of `return;`
So it is already returning `undefined` here.

EranGrin:
@SlicedSilver I did some local tests and, until now, everything looked great!

tests/e2e/graphics/test-cases/series-markers/series-markers-price-positioned.js
```diff
@@ -0,0 +1,31 @@
...
+	]);
+
+	LightweightCharts.createSeriesMarkers(
+		line,
+		[
+			{ time: '2017-04-11', position: 'atPriceBottom', color: 'orange', shape: 'arrowUp', price: 80.01 },
+			{ time: '2017-04-11', position: 'atPriceTop', color: 'orange', shape: 'arrowDown', price: 80.01 },
+			{ time: '2017-04-15', position: 'atPriceMiddle', color: 'orange', shape: 'circle', price: 74.43 },
+			{ time: '2017-04-22', position: 'atPriceBottom', color: 'orange', shape: 'arrowUp', price: 76.0 },
+			{ time: '2017-04-22', position: 'atPriceTop', color: 'orange', shape: 'arrowDown', price: 86.0 },
```
illetid, line 25: I think we need to add a case with text and test with `id` to check that `hoveredSourceId` is still working for these markers

illetid (approved):
LGTM

Issue #1697, opened 2024-09-18
SubscribeCrosshairMove incorrect cordinates


Skiv1989:
**Is your feature request related to a problem? Please describe.**

I have a chart with candle stick data. Crosshair option is magnet

Then I'm trying to draw a line on the click, coordinates for the line I've got from SubscribeCrosshairMove subscription.
Expected behavior would be draw the line on crosshair position (candle close level), but actual it will be on the mouse position. 
When magnet is on mouse position and crosshair position can be different that's where the issue is come from

**Describe the solution you'd like**

Rename method or provide correct coordinates

PR #1831, opened 2025-03-06
Magnet mod should align the OHLC instead of close

### LINKED ISSUE #1726: Why doesn't `CrosshairOptions` provide an option to snap to the nearest `open/high/low/close` data points?
Hi,

Currently, in lightweight-charts, the CrosshairMode.Magnet only allows the crosshair to snap to the close price in OHLC-based charts. However, it would be very useful if there were an option in CrosshairOptions to snap the crosshair to the nearest open, high, low, or close data points, depending on the user's preference.

This feature would greatly improve the interactivity and precision when analyzing different price levels on the chart.

Could you please consider adding this feature in a future release?

Thanks!

Jonney:
**Type of PR:**
enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #1726
- [ ] ~~Includes tests~~
- [x] Documentation update

**Overview of change:**

### DIFF
--- a/src/model/crosshair.ts
+++ b/src/model/crosshair.ts
@@ -44,7 +44,7 @@ export const enum CrosshairMode {
 	 */
 	Normal,
 	/**
-	 * This mode sticks crosshair's horizontal line to the price value of a single-value series or to the close price of OHLC-based series.
+	 * This mode sticks crosshair's horizontal line to the price value of a single-value series or to the open/high/low/close price of OHLC-based series.
 	 */
 	Magnet,
 	/**
--- a/src/model/crosshair.ts
+++ b/src/model/crosshair.ts
@@ -44,13 +44,17 @@ export const enum CrosshairMode {
 	 */
 	Normal,
 	/**
-	 * This mode sticks crosshair's horizontal line to the price value of a single-value series or to the open/high/low/close price of OHLC-based series.
+	 * This mode sticks crosshair's horizontal line to the price value of a single-value series or to the close price of OHLC-based series.
 	 */
 	Magnet,
 	/**
 	 * This mode disables rendering of the crosshair.
 	 */
 	Hidden,
+	/**
+	 * This mode sticks crosshair's horizontal line to the price value of a single-value series or to the open/high/low/close price of OHLC-based series.
+	 */
+	MagnetOHLC,
 }
 
 /** Structure describing a crosshair line (vertical or horizontal) */
--- a/src/model/magnet.ts
+++ b/src/model/magnet.ts
@@ -12,8 +12,6 @@ import { TimePointIndex } from './time-data';
 
 export class Magnet {
 	private readonly _options: CrosshairOptions;
-	private static readonly MAGNET_THRESHOLD: number = 15;
-
 	public constructor(options: CrosshairOptions) {
 		this._options = options;
 	}
@@ -71,9 +69,6 @@ export class Magnet {
 		candidates.sort((y1: Coordinate, y2: Coordinate) => Math.abs(y1 - y) - Math.abs(y2 - y));
 
 		const nearest = candidates[0];
-		if (this._options.mode === CrosshairMode.MagnetOHLC && Math.abs(nearest - y) > Magnet.MAGNET_THRESHOLD) {
-			return res;
-		}
 		res = defaultPriceScale.coordinateToPrice(nearest, firstValue);
 
 		return res;
--- a/src/model/magnet.ts
+++ b/src/model/magnet.ts
@@ -54,7 +54,12 @@ export class Magnet {
 
 				// convert bar to pixels
 				const firstPrice = ensure(series.firstValue());
-				return acc.concat([ps.priceToCoordinate(bar.value[PlotRowValueIndex.Close], firstPrice.value)]);
+				return acc.concat([
+					ps.priceToCoordinate(bar.value[PlotRowValueIndex.Open], firstPrice.value),
+					ps.priceToCoordinate(bar.value[PlotRowValueIndex.High], firstPrice.value),
+					ps.priceToCoordinate(bar.value[PlotRowValueIndex.Low], firstPrice.value),
+					ps.priceToCoordinate(bar.value[PlotRowValueIndex.Close], firstPrice.value),
+				]);
 			},
 			[] as Coordinate[]);
 
--- a/src/model/magnet.ts
+++ b/src/model/magnet.ts
@@ -12,6 +12,7 @@ import { TimePointIndex } from './time-data';
 
 export class Magnet {
 	private readonly _options: CrosshairOptions;
+
 	public constructor(options: CrosshairOptions) {
 		this._options = options;
 	}
--- a/src/model/magnet.ts
+++ b/src/model/magnet.ts
@@ -10,6 +10,16 @@ import { Series } from './series';
 import { SeriesType } from './series-options';
 import { TimePointIndex } from './time-data';
 
+const magnetPlotRowKeys: readonly PlotRowValueIndex[] = [
+	PlotRowValueIndex.Close,
+] as const;
+const magnetOHLCPlotRowKeys: readonly PlotRowValueIndex[] = [
+	PlotRowValueIndex.Open,
+	PlotRowValueIndex.High,
+	PlotRowValueIndex.Low,
+	PlotRowValueIndex.Close,
+] as const;
+
 export class Magnet {
 	private readonly _options: CrosshairOptions;
 
@@ -55,8 +65,8 @@ export class Magnet {
 				// convert bar to pixels
 				const firstPrice = ensure(series.firstValue());
 				const plotRowKeys = this._options.mode === CrosshairMode.MagnetOHLC
-					? [PlotRowValueIndex.Open, PlotRowValueIndex.High, PlotRowValueIndex.Low, PlotRowValueIndex.Close]
-					: [PlotRowValueIndex.Close];
+					? magnetOHLCPlotRowKeys
+					: magnetPlotRowKeys;
 				return acc.concat(
 					plotRowKeys.map((key: PlotRowValueIndex) => ps.priceToCoordinate(bar.value[key], firstPrice.value))
 				);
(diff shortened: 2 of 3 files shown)


src/model/crosshair.ts
```diff
@@ -44,7 +44,7 @@ export const enum CrosshairMode {
 	 */
 	Normal,
 	/**
-	 * This mode sticks crosshair's horizontal line to the price value of a single-value series or to the close price of OHLC-based series.
+	 * This mode sticks crosshair's horizontal line to the price value of a single-value series or to the open/high/low/close price of OHLC-based series.
 	 */
 	Magnet,
```
SlicedSilver, line 49: I would suggest that we rather add another `CrosshairMode` called `MagnetOHLC` (added to the end of the enum). This means that we are not introducing a breaking change to the library.

src/model/magnet.ts
```diff
@@ -54,7 +54,12 @@ export class Magnet {
 
 				// convert bar to pixels
 				const firstPrice = ensure(series.firstValue());
-				return acc.concat([ps.priceToCoordinate(bar.value[PlotRowValueIndex.Close], firstPrice.value)]);
+				return acc.concat([
+					ps.priceToCoordinate(bar.value[PlotRowValueIndex.Open], firstPrice.value),
+					ps.priceToCoordinate(bar.value[PlotRowValueIndex.High], firstPrice.value),
+					ps.priceToCoordinate(bar.value[PlotRowValueIndex.Low], firstPrice.value),
+					ps.priceToCoordinate(bar.value[PlotRowValueIndex.Close], firstPrice.value),
+				]);
```
SlicedSilver, line 62: We could probably refactor this since it has a lot of repeated code:

```ts
const plotRowKeys = this._options.mode === CrosshairMode. MagnetOHLC ? [PlotRowValueIndex.Open, PlotRowValueIndex.High, PlotRowValueIndex.Low, PlotRowValueIndex.Close] : [PlotRowValueIndex.Close];
const plotRowValues = plotRowKeys.map((key: PlotRowValueIndex) => ps.priceToCoordinate(bar.value[key], firstPrice.value));
return acc.concat(plotRowValues);
```

SlicedSilver:
Thanks @Jonney for the PR.

Jonney:
Is it a good idea to add Magnet.MAGNET_THRESHOLD to control the intensity of MagnetOHLC? @SlicedSilver

SlicedSilver:
> Is it a good idea to add Magnet.MAGNET_THRESHOLD to control the intensity of MagnetOHLC? @SlicedSilver

I personally don't think it is needed. If we are in magnet mode then I would expect that the crosshair always jumps to a price value within the bar data.

src/model/magnet.ts
```diff
@@ -54,7 +55,12 @@ export class Magnet {
 
 				// convert bar to pixels
 				const firstPrice = ensure(series.firstValue());
-				return acc.concat([ps.priceToCoordinate(bar.value[PlotRowValueIndex.Close], firstPrice.value)]);
+				const plotRowKeys = this._options.mode === CrosshairMode.MagnetOHLC
+					? [PlotRowValueIndex.Open, PlotRowValueIndex.High, PlotRowValueIndex.Low, PlotRowValueIndex.Close]
+					: [PlotRowValueIndex.Close];
+				return acc.concat(
+					plotRowKeys.map(key: PlotRowValueIndex => ps.priceToCoordinate(bar.value[key], firstPrice.value))
```
SlicedSilver, line 62: This isn't valid JS code and a build error is occurring. It's missing the brackets around the `key: PlotRowValueIndex` and should be `(key: PlotRowValueIndex)`.

Locally, before committing changes you can run `npm run verify` and it should catch issues like this.

Jonney: Sorry, I fixed it.

Jonney:
> > Is it a good idea to add Magnet.MAGNET_THRESHOLD to control the intensity of MagnetOHLC? @SlicedSilver
> 
> I personally don't think it is needed. If we are in magnet mode then I would expect that the crosshair always jumps to a price value within the bar data.

Please, I really need this MAGNET_THRESHOLD. Without MAGNET_THRESHOLD, my project is difficult to write a drawing plugin for lightweight-charts. Then I have to use the development version to overwrite the original method, just like [#1726](https://github.com/tradingview/lightweight-charts/issues/1726#issuecomment-2704402761).

SlicedSilver:
> > Is it a good idea to add Magnet.MAGNET_THRESHOLD to control the intensity of MagnetOHLC? @SlicedSilver
> 
> I personally don't think it is needed. If we are in magnet mode then I would expect that the crosshair always jumps to a price value within the bar data.

@Jonney I'm not sure if you deleted your earlier comment because I don't see it now. If you have a need for a magnet threshold setting and feature then could you please explain your use case for needing it. We can then create a new issue for that feature request and consider adding it.
In principle we try to keep the feature set of the library as small as required so that the bundle size remains small. However we will add new features if they make sense and would benefit many users.

Jonney:
After consideration, I found that I could achieve the functionality I needed without adding MAGNET_THRESHOLD, so I deleted both the code and the comments.

Besides that, there is another problem.

The `align()` function converts the OHLC price to coordinates. After comparison, it finds the nearest coordinate and then converts the coordinate back to a price. This conversion causes the price to differ from the OHLC price because the price is of type float, leading to precision issues after conversion. For example, the close price is 35.8, the price returned is 35.8000001. The user will never get the exact price.

I can refactor this part of the code so that there's no need to convert the price to coordinates. We can directly compare using the price and return the OHLC price. Would you like me to make this change in this PR, or should I open a new PR? @SlicedSilver

SlicedSilver:
> After consideration, I found that I could achieve the functionality I needed without adding MAGNET_THRESHOLD, so I deleted both the code and the comments.
> 
> Besides that, there is another problem.
> 
> The `align()` function converts the OHLC price to coordinates. After comparison, it finds the nearest coordinate and then converts the coordinate back to a price. This conversion causes the price to differ from the OHLC price because the price is of type float, leading to precision issues after conversion. For example, the close price is 35.8, the price returned is 35.8000001. The user will never get the exact price.
> 
> I can refactor this part of the code so that there's no need to convert the price to coordinates. We can directly compare using the price and return the OHLC price. Would you like me to make this change in this PR, or should I open a new PR? @SlicedSilver

I think this is fine because it is possible to get the actual exact price value from the [`subscribeCrosshairMove`](https://tradingview.github.io/lightweight-charts/docs/api/interfaces/IChartApi#subscribecrosshairmove) event using [`seriesData`](https://tradingview.github.io/lightweight-charts/docs/api/interfaces/MouseEventParams#seriesdata).

This example shows some code which does that: https://tradingview.github.io/lightweight-charts/tutorials/how_to/legends#how-to

Jonney:
> I think this is fine because it is possible to get the actual exact price value from the [`subscribeCrosshairMove`](https://tradingview.github.io/lightweight-charts/docs/api/interfaces/IChartApi#subscribecrosshairmove) event using [`seriesData`](https://tradingview.github.io/lightweight-charts/docs/api/interfaces/MouseEventParams#seriesdata).
> 
> This example shows some code which does that: https://tradingview.github.io/lightweight-charts/tutorials/how_to/legends#how-to

Look here #1697, what we got from subscribeCrosshairMove() are the mouse coordinates, not the crosshair coordinates. In Magnet mode or MagnetOHLC mode, the mouse coordinates are different from the crosshair coordinates. In Magnet mode, using seriesData will get the close price, but in MagnetOHLC mode, there are 4 values, and we don't know which point the crosshair is over.

So we have to add param.price in subscribeCrosshairMove(function(param){}), or add chart.getCrosshairPosition(). No matter which method is used, the price must come from Magnet.align() returns. This is why I suggest not converting the price to coordinates in Magnet.align(), as it will result in a loss of floating-point precision.

SlicedSilver:
> > I think this is fine because it is possible to get the actual exact price value from the [`subscribeCrosshairMove`](https://tradingview.github.io/lightweight-charts/docs/api/interfaces/IChartApi#subscribecrosshairmove) event using [`seriesData`](https://tradingview.github.io/lightweight-charts/docs/api/interfaces/MouseEventParams#seriesdata).
> > This example shows some code which does that: https://tradingview.github.io/lightweight-charts/tutorials/how_to/legends#how-to
> 
> Look here #1697, what we got from subscribeCrosshairMove() are the mouse coordinates, not the crosshair coordinates. In Magnet mode or MagnetOHLC mode, the mouse coordinates are different from the crosshair coordinates. In Magnet mode, using seriesData will get the close price, but in MagnetOHLC mode, there are 4 values, and we don't know which point the crosshair is over.
> 
> So we have to add param.price in subscribeCrosshairMove(function(param){}), or add chart.getCrosshairPosition(). No matter which method is used, the price must come from Magnet.align() returns. This is why I suggest not converting the price to coordinates in Magnet.align(), as it will result in a loss of float precision.

I think we can avoid the complexity of adding new methods or properties to the library and rather determine the magnet price in the external code. Something like this:
```ts
function nearestPrice(targetPrice: number, possiblePrices: number[]): number {
	return possiblePrices.reduce(
		(nearest, price) =>
			Math.abs(price - targetPrice) < Math.abs(nearest - targetPrice)
				? price
				: nearest,
		possiblePrices[0]
	);
}

chart.subscribeCrosshairMove(function (param) {
	const data = param.seriesData.get(candlestickSeries) as BarData<Time>;
	if (!data) {
		return;
	}
	const mode = chart.options().crosshair.mode;
	if (mode === CrosshairMode.Magnet) {
		console.log('Magnet Price is: ', data.close);
		return;
	}
	if (mode === CrosshairMode.MagnetOHLC) {
		if (param.point) {
			const crosshairPrice = candlestickSeries.coordinateToPrice(param.point.y);
			if (crosshairPrice !== null) {
				console.log(
					'Magnet Price is: ',
					nearestPrice(crosshairPrice, [
						data.close,
						data.open,
						data.high,
						data.low,
					])
				);
				return;
			}
		}
		console.log('Magnet Price is: ', data.close);
	}
});
```

SlicedSilver (approved):

PR #1835, opened 2025-03-12
Performance Optimization for Charts with Markers and Large Datasets

### LINKED ISSUE #1808: Adding markers to chart with many datapoints causes significant performance issues.
**Lightweight Charts™ Version:** 5.0.X (lastet version via CDN)

**Steps/code to reproduce:**
I am using lightweight charts to produce a chart that plots candles for an index at the 5-minute interval for the course of whole year.  This is around 60,0000+ candles on one chart, and it performs just fine with all candles loaded (meaning that I can scroll, zoom, move my cursor with no noticeable lag).  However, adding just one marker to this chart makes it unusably slow to react to input (scrolling, zooming, moving the cursor is extremely laggy.

Are there any recommendations for how to deal with large data sets with markers, without tanking performance?  I have done some rough experimentation with this, and the performance seems to begin to drop off around 15000 data points (with one marker added to the chart).

**Actual behavior:**

Adding one marker to a chart with a large data set (10s of thousands) causes significant performance issues compared to a chart with the same data set, but no markers.

**Expected behavior:**

Adding just one marker to a chart should not cause performance issues.  Scrolling, zooming, and using the cursor should be as responsive as using the same chart with no markers.

SlicedSilver:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #1808
- `N/A` ~Includes tests~

## Summary
This PR addresses a significant performance degradation issue when adding markers to charts with large datasets (15,000+ data points) in Lightweight Charts™ v5.0.X.

## Problem
When rendering charts with large datasets (particularly 5-minute interval candles over a year, ~60,000+ data points), adding even a single marker causes severe performance issues:
- Extremely laggy scrolling
- Significantly impacted zooming responsiveness
- Delayed and unresponsive cursor movement

The issue becomes noticeable at approximately 15,000 data points with just one marker added.

## Root Cause
The performance bottleneck was identified in the `_recalculateMarkers` function running on every pane update within the series-markers plugin, regardless of whether recalculation was necessary.

## Solution
This PR optimizes marker calculations by:
1. Adding a `_recalculationRequired` flag to track when recalculation is needed
2. Only running `_recalculateMarkers` when this flag is true
3. Setting the flag to true only when necessary:
   - Data changes
   - Markers are set/updated
   - Component is attached

## Testing
- Verified smooth performance with 60,000+ data points and markers
- Confirmed all chart interactions (scroll, zoom, cursor movement) remain responsive
- Tested with various marker configurations to ensure correct rendering

## References
- Original issue: [#1808](https://github.com/tradingview/lightweight-charts/issues/1808)
- Reproduction example: [Glitch demo](https://glitch.com/edit/#!/lwc-1808-issue)

### DIFF
--- a/src/plugins/series-markers/primitive.ts
+++ b/src/plugins/series-markers/primitive.ts
@@ -31,6 +31,7 @@ export class SeriesMarkersPrimitive<HorzScaleItem> implements ISeriesPrimitive<H
 	private _autoScaleMargins: AutoScaleMargins | null = null;
 	private _markersPositions: MarkerPositions | null = null;
 	private _cachedBarSpacing: number | null = null;
+	private _recalculationRequired: boolean = true;
 
 	public attached(param: SeriesAttachedParameter<HorzScaleItem>): void {
 		this._recalculateMarkers();
@@ -39,6 +40,7 @@ export class SeriesMarkersPrimitive<HorzScaleItem> implements ISeriesPrimitive<H
 		this._paneView = new SeriesMarkersPaneView(this._series, ensureNotNull(this._chart));
 		this._requestUpdate = param.requestUpdate;
 		this._series.subscribeDataChanged((scope: DataChangedScope) => this._onDataChanged(scope));
+		this._recalculationRequired = true;
 		this.requestUpdate();
 	}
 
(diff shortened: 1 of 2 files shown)


romfrancois (approved):
LGTM

illetid (approved):
lgtm

PR #1856, opened 2025-04-07
feat: boundaries price marks

AStaroverov:
**Type of PR:** 

**PR checklist:**

- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**
Added feature with bounded price marks.

Feature can be enabled with PriceScaleOptions flag boundaryMarksVisible(boolean).
After enabling this feature Price marks will constantly show min and max values on cavas.

### DIFF
--- a/src/model/price-scale.ts
+++ b/src/model/price-scale.ts
@@ -31,7 +31,7 @@ import {
 	toPercent,
 	toPercentRange,
 } from './price-scale-conversions';
-import { PriceTickMarkBuilder } from './price-tick-mark-builder';
+import { EdgeMarksOptions, PriceTickMarkBuilder } from './price-tick-mark-builder';
 import { RangeImpl } from './range-impl';
 import { sortSources } from './sort-sources';
 import { SeriesItemsIndexesRange, TimePointIndex } from './time-data';
@@ -196,7 +196,7 @@ export interface PriceScaleOptions {
 	/**
 	 * Boundary marks are the marks that are drawn at the top and bottom of the price scale.
 	 */
-	boundaryMarksVisible: boolean;
+	ensureEdgeTickMarksVisible: boolean;
 }
 
 interface RangeCache {
@@ -261,9 +261,7 @@ export class PriceScale {
 			100,
 			this._coordinateToLogical.bind(this),
 			this._logicalToCoordinate.bind(this),
-			options.boundaryMarksVisible ? {
-				getPadding: () => this.fontSize() / 2,
-			} : undefined
+			this._getEdgeMarksOptions()
 		);
 	}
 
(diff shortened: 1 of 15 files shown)


src/model/price-scale.ts
```diff
@@ -192,6 +192,11 @@ export interface PriceScaleOptions {
 	 * @defaultValue 0
 	 */
 	minimumWidth: number;
+
+	/**
+	 * Boundary marks are the marks that are drawn at the top and bottom of the price scale.
+	 */
+	ensureEdgeTickMarksVisible: boolean;
```
SlicedSilver, line 199:
```suggestion
	/**
	 * Ensures that tick marks are always visible at the very top and bottom of the price scale,
	 * regardless of the data range. When enabled, a tick mark will be drawn at both edges of the scale, 
	 * providing clear boundary indicators.
	 *
	 * @default false
	 */
	ensureEdgeTickMarksVisible: boolean;
```

src/model/price-tick-mark-builder.ts
SlicedSilver, line 1: Its annoying that the git diff on this file is scrambled (not your fault).

SlicedSilver, line 1: For the `PriceTickMarkBuilder` we are already passing `this` (PriceScale) as the first argument so I don't think we need to pass an `EdgeMarksOptions` argument at the end. Instead the code in `src/model/price-tick-mark-builder.ts` could read from a new public method on the price scale called 'getEdgeMarksPadding' (it will get minified anyway) which returns a number.

Ideally if we could remove some of the method calls and options passing then we will save a few bytes and performance.

SlicedSilver (changes requested):
Could we add some graphics tests cases:
- Enabling the option
- Enable option, then set autoscale true
- Adjust price scale, then enable the option
- Logarithmic scale option
- ...

Lets try cover the various use cases so we can check the behaviour and also know if if breaks in the future.

src/model/price-tick-mark-builder.ts
AStaroverov: resolved

AStaroverov:
@SlicedSilver added test for different cases. How do you think it will be enough?

SlicedSilver:
@AStaroverov Could we add some e2e graphics tests to cover the following additions as well:
- `setVisibleRange` 
- `getVisibleRange`
- `setAutoScale`

Please also add these to the [coverage test cases](https://github.com/tradingview/lightweight-charts/tree/master/tests/e2e/coverage). Just code that uses the syntax so that the functions are executed (and thus if they throw an error in the future then we will know).

SlicedSilver (approved):

PR #1868, opened 2025-04-18
Implemented series order functionality #136

### LINKED ISSUE #136: Option to specify zIndex of the series
i would like to control the layers of the chart elements via z-index..  is there already something for this?  i may randomly add a line and i would like to group them somehow

i believe something like 

z-indes:0  is the middle layer 
-1 is behind layer, 1 is infront layer

higher the number, closer to the eye

kirchet:
**Type of PR:**  enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #136
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Added possibility to change series order and get current order value

```TS
const series1 = chart.addSeries(...);
const series2 = chart.addSeries(...);

series1.order(); // 0
series2.order(); // 1

series2.setOrder(0);

series1.order(); // 1
series2.order(); // 0

chart.removeSeries(series2);

series1.order(); // 0
```

Series are painted on a pane by their order - than more the order the more top the series.

A series order value can be recalculated by the chart automatically after some operations
like removing other series or moving the series to another pane.

Also notice that price scales use formatters from the series with the lowest order, so moving series
to a new position may change a corresponding price scale formatter.

### DIFF
--- a/src/api/iseries-api.ts
+++ b/src/api/iseries-api.ts
@@ -315,21 +315,21 @@ export interface ISeriesApi<
 	moveToPane(paneIndex: number): void;
 
 	/**
-	 * Returns the index of the series in all series of the pane.
+	 * Gets the zero-based index of this series within the list of all series on the current pane.
 	 *
-	 * @returns The index of the series.
+	 * @returns The current index of the series in the pane's series collection.
 	 */
 	seriesOrder(): number;
 
 	/**
-	 * Sets the index of the series in all series of the pane.
-	 * This value can be recalculated by the chart automatically after some operations
-	 * like removing other series or moving the series to another pane.
-	 * Notice that values like -1 or actual series count or greater are adjusted to correct values.
-	 * Also notice that price scales use formatters from the series with the lowest index, so moving series
-	 * to a new position may change a corresponding price scale formatter.
+	 * Sets the zero-based index of this series within the pane's series collection, thereby adjusting its rendering order.
 	 *
-	 * @param order - The index of the series.
+	 * Note:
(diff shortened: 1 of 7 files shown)


src/api/iseries-api.ts
```diff
@@ -314,6 +314,25 @@ export interface ISeriesApi<
...
+	 * Sets the index of the series in all series of the pane.
+	 * This value can be recalculated by the chart automatically after some operations
+	 * like removing other series or moving the series to another pane.
+	 * Notice that values like -1 or actual series count or greater are adjusted to correct values.
+	 * Also notice that price scales use formatters from the series with the lowest index, so moving series
+	 * to a new position may change a corresponding price scale formatter.
+	 *
+	 * @param order - The index of the series.
+	 */
+	setSeriesOrder(order: number): void;
```
SlicedSilver, line 334:
```suggestion
	/**
	 * Gets the zero-based index of this series within the list of all series on the current pane.
	 *
	 * @returns The current index of the series in the pane's series collection.
	 */
	seriesOrder(): number;

	/**
	 * Sets the zero-based index of this series within the pane's series collection, thereby adjusting its rendering order.
	 *
	 * Note:
	 * - The chart may automatically recalculate this index after operations such as removing other series or moving this series to a different pane.
	 * - If the provided index is less than 0, equal to, or greater than the number of series, it will be clamped to a valid range.
	 * - Price scales derive their formatters from the series with the lowest index; changing the order may affect the price scale's formatting
	 *
	 * @param order - The desired zero-based index to set for this series within the pane.
	 */
	setSeriesOrder(order: number): void;
```

SlicedSilver (approved):

Issue #1872, opened 2025-04-21
Chart api methods for direct panes control


ukorvl:
**Is your feature request related to a problem? Please describe.**

Hey, i'm developing React [wrapper](https://github.com/ukorvl/lightweight-charts-react-components) for lightweight-charts library (yes, another one).
The idea is to add a declarative way of creating panes by `<Pane>` component making all children inside (series, primitives, price scales, etc) utilize this pane. For instance:
```tsx
<Pane>
  <Series>
  <Primitive>
</Pane>
```
The problem of this approach is that i didn't find a way to programmatically add a pane to the chart, except a series creation which implicitly creates a new pane, if `pandeIndex` arg is provided. This fact in its turn makes parent `<Pane>` dependable on its child `<Series>`, which produces side effects.
Ideally, `Series` component should first wait for the pane to be initialized and then render.
It can be achieved easily if we can add panes to the chart directly.

The other problem is that if i remove, let's say, pane with `1` index, having 0,1,2 panes initially, the `2` pane will change its index to `1`. But it contradicts the general "controlled component" approach to jsx components. If i pass `order={2}` or `paneIndex={2}` prop to this pane, it ideally should not change the index under the hood (just because user removed some other pane).

**Describe the solution you'd like**

A direct pane creation method like `chartApi.addPane(index, opts?)`
Also, maybe an option to disable automatic panes removal.

**Additional context**

I have on mind a "last-resort" way to achieve that just by using dummy "always-exisitng" series for each pane (suggesting it will not affect performance too much) to make panes behavior more predictable. But want to ensure that it is not planned to add it to the core library.

SlicedSilver:
Thank you for your detailed feedback and for sharing your use case—it's really helpful to understand how developers are using the library in more complex, declarative scenarios like React wrappers.

**Current API Limitations**

At present, the API doesn't provide a direct way to add or remove panes independently of series, as you've discovered. The current design is intentional, aiming for simplicity, but we understand that it can lead to less predictable behaviour when building controlled components, especially when pane indices shift after removal.

**Feature Request & Roadmap**

[...]

PR #1894, opened 2025-06-09
Improve pane api

illetid:
**Type of PR:**  enhancement

**PR checklist:**

- [x] Addresses an existing issue
- [x] Includes coverage tests
- [ ] Documentation update

**Overview of change:**
Added few requested methods to improve usage of the pane's API

Fixes: 
-  #1898 by exposing stretchFactor. so users can easily set relative pane's height
- #1896 fix movePane behaviour 
- #1872 ability to create empty pane 
 - ##1907 return null when there is no pane yet

### DIFF
--- a/src/api/ipane-api.ts
+++ b/src/api/ipane-api.ts
@@ -49,9 +49,9 @@ export interface IPaneApi<HorzScaleItem> {
 	/**
 	 * Retrieves the HTML element of the pane.
 	 *
-	 * @returns The HTML element of the pane.
+	 * @returns The HTML element of the pane or null if pane wasn't created yet.
 	 */
-	getHTMLElement(): HTMLElement;
+	getHTMLElement(): HTMLElement | null;
 
 	/**
 	 * Attaches additional drawing primitive to the pane
--- a/src/api/pane-api.ts
+++ b/src/api/pane-api.ts
@@ -73,8 +73,12 @@ export class PaneApi<HorzScaleItem> implements IPaneApi<HorzScaleItem> {
 		return this._pane.series().map((source: Series<SeriesType>) => this._seriesApiGetter(source)) ?? [];
 	}
 
-	public getHTMLElement(): HTMLElement {
-		return this._chartWidget.paneWidgets()[this.paneIndex()].getElement();
+	public getHTMLElement(): HTMLElement | null {
+		const widgets = this._chartWidget.paneWidgets();
+		if (!widgets || widgets.length === 0 || !widgets[this.paneIndex()]) {
+			return null;
+		}
+		return widgets[this.paneIndex()].getElement();
 	}
 
 	public attachPrimitive(primitive: IPanePrimitive<HorzScaleItem>): void {


SlicedSilver:
Could we add a few e2e graphics tests (and have each test check one thing)
- stretchFactor
- preservePane
- Skip default pane, add 2 panes manually, only add a series to second one
- ...

I think we should also update the coverage tests because we have increased the API.

src/api/ipane-api.ts
```diff
@@ -72,4 +72,32 @@ export interface IPaneApi<HorzScaleItem> {
...
+	 * @returns The stretch factor of the pane. Default is 1000
+	 */
+	getStretchFactor(): number;
+
+	/**
+	 * Sets the stretch factor of the pane.
+	 *
+	 * @param stretchFactor - The stretch factor of the pane.
+	 */
+	setStretchFactor(stretchFactor: number): void;
```
SlicedSilver, line 102: It would be good if 'stretch factor' was explained somewhere in these comments, and potentially a documentation explanation as well.

src/api/ipane-api.ts
```diff
@@ -72,4 +72,32 @@ export interface IPaneApi<HorzScaleItem> {
...
+	 * @param preserve - Whether to preserve the empty pane
+	 */
+	setPreserveEmptyPane(preserve: boolean): void;
+
+	/**
+	 * Returns whether to preserve the empty pane
+	 *
+	 * @returns Whether to preserve the empty pane
+	 */
+	preserveEmptyPane(): boolean;
```
SlicedSilver, line 88: What is the main benefit to being able to preserve empty panes? Is this not something that could be worked around by correctly adding the new series to a pane before removing the existing one?

Or is there a need to literally have an empty pane in the chart? Adding a whitespace series might give the same outcome.

I'm asking because if it is not strictly required then perhaps we don't add it to the API so their is less to maintain and support in the future.

src/model/chart-model.ts
```diff
@@ -327,6 +327,13 @@ export interface ChartOptionsBase {
 	 * Basic localization options
 	 */
 	localization: LocalizationOptionsBase;
+
+	/**
+	 * Whether to add a default pane to the chart
+	 * Disable this option when you want to create a chart with no panes and add them manually
+	 * @defaultValue `true`
+	 */
+	addDefaultPane: boolean;
```
SlicedSilver, line 336: Whats the use case for creating a chart with no panes or series?

Couldn't the developer rather only create the chart when they are ready to show something?

tests/e2e/coverage/test-cases/panes/add-pane.js
```diff
@@ -12,13 +12,20 @@ function interactionsToPerform() {
...
 function beforeInteractions(container) {
-	chart = LightweightCharts.createChart(container);
+	chart = LightweightCharts.createChart(container, {
+		addDefaultPane: false,
+	});
 
-	const mainSeries = chart.addSeries(LightweightCharts.LineSeries);
+	const pane0 = chart.addPane(true);
+
+	const mainSeries = chart.addSeries(LightweightCharts.LineSeries, {}, pane0.paneIndex());
```
SlicedSilver, line 21: Seems like having a `pane.addSeries()` method might be useful. The pane has access to the ChartApi so it would be able to do this.

src/api/ipane-api.ts
illetid: Added short explanation with example

vishnuc (commented):
* So if you have three panes, and you want to make the first pane twice as big as the second and third panes, you can set the stretch factor of the first pane to 2000.

still its at 2000

illetid:
@vishnuc Hi I've added better example in tsdoc https://github.com/tradingview/lightweight-charts/pull/1894/files#diff-c8cf914ef60b765fed227cad366e10a5401ac8d321e695bf684c3463d2ac6093R107-R119

We changed the default to 1, to be less confusing. Sum of stretches shouldn't equal 1 or any other particular number, what's matter is a total sum of stretches of all panes and their difference between each others. You can check `_adjustSizeImpl` https://github.com/tradingview/lightweight-charts/blob/master/src/gui/chart-widget.ts#L455 and check how the library is calculating height of panes

vishnuc:
> @vishnuc Hi I've added better example in tsdoc https://github.com/tradingview/lightweight-charts/pull/1894/files#diff-c8cf914ef60b765fed227cad366e10a5401ac8d321e695bf684c3463d2ac6093R107-R119
> 
> We changed the default to 1, to be less confusing. Sum of stretches shouldn't equal 1 or any other particular number, what's matter is a total sum of stretches of all panes and their difference between each others. You can check `_adjustSizeImpl` https://github.com/tradingview/lightweight-charts/blob/master/src/gui/chart-widget.ts#L455 and check how the library is calculating height of panes

thanks makes sense.. on users drag and resizing does this stretchFactor change.. and can we get the changed via getsetfactor ?

ukorvl:
Hey @illetid, thanks a lot for that long-awaited update!
Just to clarify, if a user decides to preserve 0 pane and 2 pane, with 0 and 2 indexes accordingly, how it would assign indexes if a user removes, for instance, pane with index `1`? Will it shift them, making previously 2nd pane having index of 1 or there will be a "gap"?

illetid:
> Hey @illetid, thanks a lot for that long-awaited update! Just to clarify, if a user decides to preserve 0 pane and 2 pane, with 0 and 2 indexes accordingly, how it would assign indexes if a user removes, for instance, pane with index `1`? Will it shift them, making previously 2nd pane having index of 1 or there will be a "gap"?

my assumption was if you want to build a component based wrapper you should always use preserve empty pane and `addDefaultPane: false`. and if you want to remove pane you just not reder the component in react.

Answering your question yes indexes will be recalculated since it's an array indexes

ukorvl:
> > Hey @illetid, thanks a lot for that long-awaited update! Just to clarify, if a user decides to preserve 0 pane and 2 pane, with 0 and 2 indexes accordingly, how it would assign indexes if a user removes, for instance, pane with index `1`? Will it shift them, making previously 2nd pane having index of 1 or there will be a "gap"?
> 
> my assumption was if you want to build a component based wrapper you should always use preserve empty pane and `addDefaultPane: false`. and if you want to remove pane you just not reder the component in react.
> 
> Answering your question yes indexes will be recalculated since it's an array indexes

Agree with the assumption, it solves the base problem of wrapper component. I'm currently looking a bit further onto the controllable pane indexes, but it is not related to the current scope.

SlicedSilver (approved):

PR #1940, opened 2025-08-04
Adds autoScale options for seriesMarkers (#1937)

### LINKED ISSUE #1937: Option to disable autoScale behavior for seriesMarkers
When I create seriesMarkers asynchronously, the candlestick chart (which is already rendered) briefly jumps when the markers are drawn. 

After some investigation, I found that this behavior is caused by the autoscaleInfo of SeriesMarkersPrimitive.

Would it be possible to add an option to control whether seriesMarkers should use the autoScale feature? 

This would help avoid unwanted visual shifts when updating markers dynamically.

Image

zbinlin:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #1937 
- [x] Includes tests
- [ ] Documentation update

### DIFF
--- a/src/plugins/series-markers/options.ts
+++ b/src/plugins/series-markers/options.ts
@@ -13,7 +13,15 @@ export type SeriesMarkerZOrder = 'top' | 'aboveSeries' | 'normal';
  */
 export interface SeriesMarkersOptions {
 	/**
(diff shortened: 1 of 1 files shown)


SlicedSilver:
@tpunt Merged to master.
It will be included in the next version of the library. I'm not sure when that will be, but if you need to use this earlier than that then you could use a local build of master.

PR #1945, opened 2025-08-07
POC data conflation

illetid:
**Type of PR:** 

**PR checklist:**

- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**

### DIFF
--- a/src/model/data-conflater.ts
+++ b/src/model/data-conflater.ts
@@ -1,0 +1,187 @@
+import { TimePoint } from './horz-scale-behavior-time/types';
+import { PlotRowValueIndex } from './plot-data';
+import { SeriesPlotRow } from './series-data';
+import { SeriesType } from './series-options';
+import { TimePointIndex } from './time-data';
+
+/**
+ * Defines a conflation rule that determines how many bars to merge based on bar spacing.
+ */
+interface ConflationRule {
+	barsToMerge: number;
+	forBarSpacingLargerThan: number;
+}
+
+/**
+ * Represents a conflated chunk of data points.
+ */
+export interface ConflatedChunk {
+	startIndex: TimePointIndex;
+	endIndex: TimePointIndex;
+	startTime: TimePoint;
+	endTime: TimePoint;
+	open: number;
+	high: number;
+	low: number;
+	close: number;
+	originalDataCount: number;
+}
+
+const CONFLATION_RULES: ConflationRule[] = [
+	{
+		barsToMerge: 2,
+		forBarSpacingLargerThan: 0.5,
+	},
+	{
+		barsToMerge: 5,
+		forBarSpacingLargerThan: 0.2,
+	},
+	{
+		barsToMerge: 10,
+		forBarSpacingLargerThan: 0.1,
+	},
+	{
+		barsToMerge: 25,
+		forBarSpacingLargerThan: 0.05,
+	},
+	{
+		barsToMerge: 50,
+		forBarSpacingLargerThan: 0.02,
+	},
+	{
+		barsToMerge: 100,
+		forBarSpacingLargerThan: 0,
+	},
+];
+
+export class DataConflater<T extends SeriesType> {
+	private _conflatedChunks: Map<number, ConflatedChunk[]> = new Map();
+	private _lastBarSpacing: number = -1;
+	private _lastDataVersion: number = -1;
+
+	/**
+	 * Conflates data based on adaptive bar spacing rules instead of a fixed factor.
+	 * Uses multiple levels of detail for optimal performance at different zoom levels.
+	 *
+	 * @param data - The original series data
+	 * @param barSpacing - Current bar spacing in pixels
+	 * @returns Conflated data array
+	 */
+	public conflateDataAdaptive(data: readonly SeriesPlotRow<T>[], barSpacing: number): SeriesPlotRow<T>[] {
+		if (data.length === 0) {
+			return [];
+		}
+
+		const rule = this._findConflationRule(barSpacing);
+		if (rule.barsToMerge <= 1) {
+			return data.slice();
+		}
+
+		const dataVersion = this._getDataVersion(data);
+		if (this._lastBarSpacing !== barSpacing || this._lastDataVersion !== dataVersion) {
+			this._rebuildChunks(data, rule);
+			this._lastBarSpacing = barSpacing;
+			this._lastDataVersion = dataVersion;
+		}
+
+		return this._chunksToSeriesPlotRows(rule.barsToMerge);
+	}
+
+	private _findConflationRule(barSpacing: number): ConflationRule {
+		for (const rule of CONFLATION_RULES) {
+			if (barSpacing >= rule.forBarSpacingLargerThan) {
+				return rule;
+			}
+		}
+		// This should theoretically not be reached if the last rule has forBarSpacingLargerThan: 0
+		return { barsToMerge: 1, forBarSpacingLargerThan: 0.5 };
+	}
+
+	/**
+	 * Generates a simple version hash for the data to detect changes.
+	 */
+	private _getDataVersion(data: readonly SeriesPlotRow<T>[]): number {
+		if (data.length === 0) {
+			return 0;
+		}
+		// Simple hash based on data length and first/last items
+		const first = data[0];
+		const last = data[data.length - 1];
+		return data.length * 31 + first.index * 17 + last.index * 13;
+	}
+
+	/**
+	 * Rebuilds conflated chunks for the given rule.
+	 */
+	private _rebuildChunks(data: readonly SeriesPlotRow<T>[], rule: ConflationRule): void {
+		const chunks: ConflatedChunk[] = [];
+		let currentChunk: ConflatedChunk | null = null;
+		let dataPointsInChunk = 0;
+
+		for (let i = 0; i < data.length; i++) {
+			const row = data[i];
+			const price = this._extractPrice(row);
+
+			if (currentChunk === null || dataPointsInChunk >= rule.barsToMerge) {
+				if (currentChunk !== null) {
+					chunks.push(currentChunk);
+				}
+
+				currentChunk = {
+					startIndex: row.index,
+					endIndex: row.index,
+					startTime: row.time as unknown as TimePoint,
+					endTime: row.time as unknown as TimePoint,
+					open: price,
+					high: price,
+					low: price,
+					close: price,
+					originalDataCount: 1,
+				};
+				dataPointsInChunk = 1;
+			} else {
+				currentChunk.endIndex = row.index;
+				currentChunk.endTime = row.time as unknown as TimePoint;
+				currentChunk.high = Math.max(currentChunk.high, price);
+				currentChunk.low = Math.min(currentChunk.low, price);
+				currentChunk.close = price;
+				currentChunk.originalDataCount++;
+				dataPointsInChunk++;
+			}
+		}
+
+		if (currentChunk !== null) {
+			chunks.push(currentChunk);
+		}
+
+		this._conflatedChunks.set(rule.barsToMerge, chunks);
+	}
+
+	/**
+	 * Converts conflated chunks back to SeriesPlotRow format.
+	 */
+	private _chunksToSeriesPlotRows(barsToMerge: number): SeriesPlotRow<T>[] {
+		const chunks = this._conflatedChunks.get(barsToMerge) || [];
+		const result: SeriesPlotRow<T>[] = [];
+
+		for (const chunk of chunks) {
+			const conflatedRow = {
+				index: chunk.startIndex,
+				time: chunk.endTime,
+				originalTime: chunk.endTime, // originalTime might need to be startTime for some use cases
+				value: [chunk.open, chunk.high, chunk.low, chunk.close],
+			};
+
+			result.push(conflatedRow as unknown as SeriesPlotRow<T>);
+		}
+
+		return result;
+	}
+
+	/**
+	 * Extracts price from a series plot row (close value for simplicity).
+	 */
+	private _extractPrice(row: SeriesPlotRow<T>): number {
+		return row.value[PlotRowValueIndex.Close];
+	}
+}
--- a/src/model/series.ts
+++ b/src/model/series.ts
@@ -22,6 +22,7 @@ import { BarPrice } from './bar';
 import { IChartModelBase } from './chart-model';
 import { Coordinate } from './coordinate';
 import { CustomPriceLine } from './custom-price-line';
+import { DataConflater } from './data-conflater';
 import { isDefaultPriceScale } from './default-price-scale';
 import { CustomData, CustomSeriesWhitespaceData, ICustomSeriesPaneView, WhitespaceCheck } from './icustom-series';
 import { PrimitiveHoveredItem, PrimitivePaneViewZOrder } from './ipane-primitive';
@@ -104,6 +105,10 @@ export class Series<T extends SeriesType> extends PriceDataSource implements IDe
 	private readonly _options: SeriesOptionsInternal<T>;
 	private _animationTimeoutId: TimerId | null = null;
 	private _primitives: SeriesPrimitiveWrapper[] = [];
+
+	private readonly _dataConflater: DataConflater<T> = new DataConflater<T>();
+	private _conflatedDataCache: SeriesPlotList<T> | null = null;
+	private _lastConflationFactor: number = 1;
 
 	public constructor(
 		model: IChartModelBase,
@@ -241,6 +246,9 @@ export class Series<T extends SeriesType> extends PriceDataSource implements IDe
 	public setData(data: readonly SeriesPlotRow<T>[], updateInfo?: SeriesUpdateInfo): void {
 		this._data.setData(data);
 
+		// Invalidate conflated data cache when new data is set
+		this._conflatedDataCache = null;
+
 		this._paneView.update('data');
 
 		if (this._lastPriceAnimationPaneView !== null) {
@@ -305,6 +313,26 @@ export class Series<T extends SeriesType> extends PriceDataSource implements IDe
 
 	public bars(): SeriesPlotList<T> {
 		return this._data;
+	}
+
+	public conflatedBars(): SeriesPlotList<T> {
+		const timeScale = this.model().timeScale();
+		const barSpacing = timeScale.barSpacing();
+		const conflationEnabled = timeScale.options().enableConflation;
+
+		// If conflation is disabled or bar spacing is large enough, return original data
+		if (!conflationEnabled || barSpacing >= 0.5) {
+			return this._data;
+		}
+
+		// Use bar spacing as the cache key instead of conflation factor
+		const cacheKey = Math.round(barSpacing * 1000);
+		if (this._conflatedDataCache === null || this._lastConflationFactor !== cacheKey) {
+			this._regenerateConflatedDataAdaptive(barSpacing);
+			this._lastConflationFactor = cacheKey;
+		}
+
+		return this._conflatedDataCache !== null ? this._conflatedDataCache : this._data;
 	}
 
 	public dataAt(time: TimePointIndex): SeriesDataAtTypeMap[SeriesType] | null {
@@ -650,5 +678,14 @@ export class Series<T extends SeriesType> extends PriceDataSource implements IDe
 		const res: IPaneView[] = [];
 		extractPrimitivePaneViews(this._primitives, extractor, zOrder, res);
 		return res;
+	}
+
+	private _regenerateConflatedDataAdaptive(barSpacing: number): void {
+		const originalRows = this._data.rows();
+		const conflatedRows = this._dataConflater.conflateDataAdaptive(originalRows, barSpacing);
+
+		// Create a new plot list with conflated data
+		this._conflatedDataCache = createSeriesPlotList<T>();
+		this._conflatedDataCache.setData(conflatedRows);
 	}
 }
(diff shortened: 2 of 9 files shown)


src/model/conflation/types.ts
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Conflation cache entry for performance optimization.
+ */
+export interface ConflationCacheEntry<T extends SeriesType> {
+	version: number;
+	results: Map<number, ConflationResult<T>>;
+	inflight: Map<number, Promise<void>>;
+	lastAccessed: number;
+}
```
SlicedSilver:
```suggestion
```

These types aren't used anywhere.

src/model/series/pane-view.ts
```diff
@@ -1,12 +1,16 @@
...
-export interface ISeriesCustomPaneView extends IUpdatablePaneView {
-	priceValueBuilder(plotRow: CustomData<unknown> | CustomSeriesWhitespaceData<unknown>): CustomSeriesPricePlotValues;
-	isWhitespace(data: CustomData<unknown> | CustomSeriesWhitespaceData<unknown>): data is CustomSeriesWhitespaceData<unknown>;
+export interface ISeriesCustomPaneView<
+	HorzScaleItem = unknown,
+	TData extends CustomData<HorzScaleItem> = CustomData<HorzScaleItem>
+> extends IUpdatablePaneView {
+	priceValueBuilder(plotRow: TData | CustomSeriesWhitespaceData<HorzScaleItem>): CustomSeriesPricePlotValues;
+	isWhitespace(data: TData | CustomSeriesWhitespaceData<HorzScaleItem>): data is CustomSeriesWhitespaceData<HorzScaleItem>;
+	conflationReducer?(items: readonly CustomConflationContext<HorzScaleItem, TData>[]): TData;
```
SlicedSilver:
```suggestion
	conflationReducer: undefined | ((items: readonly CustomConflationContext<HorzScaleItem, TData>[]) => TData);
```
Make the method exist but possibly return undefined. This simplifies other code.

src/model/data-conflater.ts
```diff
@@ -0,0 +1,499 @@
...
+interface ConflationCacheEntry<T extends SeriesType> {
+	version: number;
+	results: Map<number, SeriesPlotRow<T>[]>;
+}
+
+export { DEFAULT_CONFLATION_FACTORS } from './conflation/constants';
+
+export class DataConflater<T extends SeriesType> {
+	private _dataCache: WeakMap<readonly SeriesPlotRow<T>[], ConflationCacheEntry<T>> = new WeakMap();
+	private _customRules: ConflationRule[] | null = null;
```
SlicedSilver:
```suggestion
```
Custom rules can rather be handled in time-scale only.

src/model/data-conflater.ts
```diff
@@ -0,0 +1,499 @@
...
+					result.push(conflatedRow as unknown as SeriesPlotRow<T>);
+				}
+			} else {
+				// built-in series use the standard format
+				const conflatedRow = {
+					index: chunk.startIndex,
+					time: chunk.endTime,
+					originalTime: chunk.endTime,
+					value: [chunk.open, chunk.high, chunk.low, chunk.close],
+				};
```
SlicedSilver: Discussion:
In this function we use the `endTime` for the conflated rows. I'm not sure if that is what we would like. Typically we use the time to represent the start of bar, so if we are conflating candlesticks then I think it should be the startTime for the combined data. Imagine we had 60 1 minute bars which we conflate into a single 1 hour bar then usually this would have the start time not the end time (08:00 instead of 08:59).

I didn't change this in my patch because we should discuss it first.

src/model/series.ts
```diff
@@ -307,6 +341,52 @@ export class Series<T extends SeriesType> extends PriceDataSource implements IDe
 		return this._data;
 	}
 
+	public setCustomConflationReducer<TCustomData>(reducer: (items: readonly TCustomData[]) => TCustomData): void {
+		this._customConflationReducer = reducer as unknown as (items: readonly unknown[]) => unknown;
```
SlicedSilver:
```suggestion
	public setCustomConflationReducer(reducer: CustomConflationReducer<unknown>): void {
		this._customConflationReducer = reducer;
```
Refactored so we don't need type assertions

src/model/series/custom-pane-view.ts
```diff
@@ -78,6 +78,15 @@ export class SeriesCustomPaneView extends SeriesPaneViewBase<
...
 	}
 
+	public getConflationReducer(): ((items: readonly unknown[]) => unknown) | undefined {
+		const reducer = this._paneView.conflationReducer; // eslint-disable-line @typescript-eslint/unbound-method
+		if (!reducer) {
+			return undefined;
+		}
+
+		return reducer as (items: readonly unknown[]) => unknown;
+	}
```
SlicedSilver:
```suggestion
	public get conflationReducer(): ((items: readonly CustomConflationContext<unknown, CustomData<unknown>>[]) => CustomData<unknown>) | undefined {
		// eslint-disable-next-line @typescript-eslint/unbound-method
		return this._paneView.conflationReducer;
	}
```
Using a getter simplifies the code using this, and also simplifies this code and it's types.

src/model/time-scale.ts
```diff
@@ -252,6 +314,7 @@ export interface ITimeScale {
 	options(): Readonly<HorzScaleOptions>;
 
 	recalculateIndicesWithData(): void;
+	conflationFactor(): number;
```
SlicedSilver:
```suggestion
	conflationFactor(): number;
	possibleConflationFactors(): number[];
```

SlicedSilver (changes requested):
Here is a patch with all the changes I've suggested. Please try it and test it. It reduces the bundle size and would hopefully be faster as well.

I have another discussion / suggestion about how we could further simplify the feature but it requires refactoring a few things so it's best to discuss separately. I'll post that shortly once I've written it.

[conflate-bars-review-suggestions.patch](https://github.com/user-attachments/files/23337019/conflate-bars-review-suggestions.patch)

SlicedSilver:
## Excellent work on the conflation feature! 🎉 

This is a substantial and well-implemented addition that will significantly improve performance for large datasets. However, I'd like to suggest a refinement that could address a key UX issue and simplify the implementation considerably.

## The Core UX Problem

The current rule-based conflation system creates **visual whitespace gaps** that confuse users, particularly for **non-continuous series types** like candlesticks, bars, and histogram columns (line series are less affected since they connect data points continuously). Here's why:

- **Happy path example**: At `barSpacing: 0.5` on a DPR=2 display, we have exactly 1 bar per physical pixel - this displays perfectly with no conflation needed
- **Problem case**: At `barSpacing: 0.45` (example), the current system conflates 5 bars into 1 chunk
    - This renders 1 bar in the space originally allocated for 5 bars (`0.45 × 5 = 2.25` pixels)  
    - Users see gaps between candles where data should be
    - Crosshair can hover over these "empty" spaces, creating confusion about data presence
    - The visual discontinuity makes it appear as if data is missing when zooming out

## When Should We Conflate?

The fundamental question is: **when do we have more data than we can physically display?**

For any given display:
- **Physical pixel constraint**: We can't meaningfully display more than 1 bar per physical pixel
- **Data density calculation**: `barsPerPhysicalPixel = 1 / (barSpacing × devicePixelRatio)`
- **Conflation trigger**: When `barsPerPhysicalPixel > 2`, we should conflate by the next power of 2

This ensures we only conflate when the display literally cannot show the additional detail, making the process invisible to users since no visual information is actually lost.

## Proposed Solution: Power-of-2 Conflation Tied to Display Resolution

Instead of arbitrary conflation rules, let's use a **DPR-aware, power-of-2 approach** that only conflates when the display physically cannot show the detail:

### Key Benefits:
1. **No visual gaps** - conflation levels align perfectly with display capabilities
2. **Simpler API** - eliminates need for custom conflation rules  
3. **Faster performance** - recursive building and simpler chunking (always merge 2 items)
4. **Better DX** - easier for plugin authors (always handle exactly 2 items)

### Technical Approach:

**Conflation Levels:** Powers of 2 only (2, 4, 8, 16, 32, 64, 128, 256, 512)

**Threshold Calculation:**
```typescript
// Conflate when we can't physically display the detail
const conflationThreshold = 1.0 / devicePixelRatio; // 0.5 for DPR=2, 1.0 for DPR=1
const conflationLevel = barSpacing < conflationThreshold 
  ? Math.pow(2, Math.floor(Math.log2(conflationThreshold / barSpacing)))
  : 1;
```

**Recursive Building:**
Since we're using consistent power-of-2 reduction, each conflation level can build directly on the previous level's results rather than recalculating from the original dataset. This dramatically reduces computational work:

```
Original: 1000 bars
Level 2:  500 bars (build from original: 1000 → 500)
Level 4:  250 bars (build from level 2: 500 → 250) ✅ Reuse previous work
Level 8:  125 bars (build from level 4: 250 → 125) ✅ Reuse previous work  
Level 16: 62 bars + 1 remainder (build from level 8: 125 → 62 + 1)
Level 32: 31 bars + 0 remainder (build from level 16: 62 → 31) + 1 remainder 
```

**Detailed Remainder Handling:**
Remainders are accumulated and carried forward through conflation levels:

1. **Single remainders**: When we have an odd number of items (e.g., 125 bars), the extra item becomes a remainder and is preserved as-is
2. **Remainder accumulation**: Remainders from previous levels are carried forward and only conflated when we have enough to form a complete chunk at the current level. For example, if we eventually get more remainders than the conflation level (e.g. 20 remainders at level 16) then we can conflate the remainder as well. However I'm not sure this can ever actually happen. An alternative would be that we conflate remainders whenever they are even.
3. **Example walkthrough**:
   - Level 8: 125 bars → 62 conflated + 1 remainder  
   - Level 16: 62 bars → 31 conflated + 0 new remainders + 1 carried remainder = 31 conflated + 1 total remainder
   - Level 32: 31 bars → 15 conflated + 1 new remainder + 1 carried remainder = 15 conflated + 2 total remainders
   - Level 64: 15 bars + 2 remainders = 15 bars + 2 remainders (not enough remainders to conflate at level 64)

This approach ensures no data is lost and maintains temporal accuracy by keeping remainders at their original resolution until they can be properly grouped.

Discussion point: do we conflate that 2 remainders into a single bar and thus we then have 16 bars? or wait until a full set (maybe not possible), or just always leave as a remainder?

**Simplified Chunking:** Always merge exactly 2 items, handle remainders by accumulation.

### API Simplifications:

**Remove these options:**
```typescript
// ❌ Remove - no longer needed
customConflationRules?: CustomConflationRules;
```

**Simplify custom series interface:**
```typescript
// ✅ Always exactly 2 items instead of variable array
conflationReducer?(item1: CustomConflationContext, item2: CustomConflationContext): TData;
```

**Smarter precomputation:**
```typescript
const effectiveBarSpacingLimit = Math.max(
  minBarSpacing, 
  Math.min(currentBarSpacing, width / dataLength)
);
// Calculate only the conflation levels we'll actually need
```
The effective bar spacing can be used to determine the largest conflation level we can expect (or should consider for precompute). We consider minBarSpacing so we don't calculate more than could ever be used, and the other part is best guess typical bar spacing that might be used (`fitContent` is common use-case).

## Implementation Benefits:

1. **Performance**: Recursive building reduces computational work exponentially
2. **Memory**: More predictable memory usage patterns with consistent chunk sizes (preallocate chunks, arrays, and simplify chunks data possibily)
3. **Code Complexity**: Eliminate rules engine, simpler chunking logic (always 2-item merge)
4. **Bundle Size**: Remove rules/configuration overhead
5. **User Experience**: Seamless zooming without visual artifacts
6. **Plugin Development**: Simpler conflation reducer interface - always handle exactly 2 items

## Questions for Discussion:

This approach maintains all the performance benefits (well technically it is less aggresive at the early stages of zooming out but gets close, see example levels below) while creating a more intuitive and visually consistent experience. The power-of-2 approach is mathematically elegant, computationally efficient, and most importantly, aligned with how displays actually work. *What do you think?*


## Appendix
### Effective equivalent for current rules system
So instead of conflation rules like:
```ts
const DEFAULT_CONFLATION_RULES_VALUES: InternalConflationArray[] = [
  [2, 0.5],
  [5, 0.2],
  [10, 0.1],
  [25, 0.05],
  [50, 0.02],
  [100, 0.01],
  [200, 0.005],
  [300, 0],
];
```

we are essentially doing this (these values might not be perfect, pixel perfect examination and testing required):
```ts
const DEFAULT_CONFLATION_RULES_VALUES: InternalConflationArray[] = [
  [2, 0.25],
  [4, 0.125],
  [8, 0.0625],
  [16, 0.03125],
  [32, 0.015625],
  [64, 0.0078125],
  [128, 0.00390625],
  [256, 0.001953125],
  [512, 0],
];
```

SlicedSilver:
## Additional Benefit: Conflation as Smoothing for Sparklines 📈

I have a follow-up idea that showcases another powerful application of the conflation feature, particularly with the power-of-2 approach.

### The Sparkline Smoothing Use Case

When creating **sparkline-style plots** (small charts with no visible time scale or price scale), we often plot high-frequency financial data (e.g., 1-minute intervals for an entire trading day) in a constrained space. This typically results in very **spiky, rough-looking charts** that are hard to read and visually unappealing.

**The interesting observation**: Conflation doesn't just solve performance problems - it can also provide a **natural smoothing effect** that makes small charts much more readable.

### Why Sparklines Are Different

For sparkline use cases, the whitespace gap issue we discussed earlier **doesn't apply**:

1. **No visible time scale** - users can't see the temporal spacing inconsistencies
2. **Line series typically used** - continuous lines don't create visual gaps like candlesticks  
3. **Crosshair interaction less critical** - sparklines are often read-only visualizations
4. **Readability prioritized** - smooth, clean appearance is more important than showing every data point

This creates an opportunity to use conflation **intentionally for smoothing**, even when the display could technically show more detail.

### Proposed Enhancement: Adjustable Conflation Factor

We can extend the power-of-2 approach with a simple **adjustment factor** that allows controlled over-conflation:

```typescript
// Current approach - conflate only when display can't show detail
export const CONFLATION_THRESHOLD_PIXELS = 0.5; // For DPR=2

// Enhanced approach - allow intentional smoothing
interface TimeScaleOptions {
  /**
   * Smoothing factor for conflation thresholds. 
   * - 1.0 = conflate only when display can't show detail (default)
   * - 2.0 = conflate at 2x the display threshold (moderate smoothing)
   * - 4.0 = conflate at 4x the display threshold (strong smoothing)
   * 
   * Useful for sparklines and small charts where smoothing improves readability.
   *
   * Note: should only be adjusted for continuous series types (like line, area, and baseline series).
   */
  conflationSmoothingFactor?: number; // probably a better name for this still to be uncovered
}

// Implementation
const smoothingFactor = options.conflationSmoothingFactor ?? 1.0;
const adjustedThreshold = (1.0 / devicePixelRatio) * smoothingFactor;

const conflationLevel = barSpacing < adjustedThreshold 
  ? Math.pow(2, Math.floor(Math.log2(adjustedThreshold / barSpacing)))
  : 1;
```

### Example Configuration:

```typescript
// Sparkline with moderate smoothing
const sparklineChart = createChart(container, {
  timeScale: {
    visible: false, // Typical sparkline setup
    enableConflation: true,
    conflationSmoothingFactor: 4.0, // 4x more aggressive conflation
  },
  rightPriceScale: {
    visible: false,
  },
  grid: {
    horzLines: {
      visible: false,
    },
    vertLines: {
      visible: false,
    },
  },
});
```

### Benefits of This Approach:

1. **Unified system** - same power-of-2 conflation handles both performance and aesthetics
2. **Simple API** - single parameter controls smoothing intensity  
3. **Predictable behavior** - still uses consistent power-of-2 levels
4. **Performance benefits maintained** - recursive building still applies
5. **Flexible application** - can be used for any chart size/type where smoothing is desired

### Implementation Notes:

The beauty of this approach is that it **builds naturally on the power-of-2 system** we're already proposing. The recursive building, remainder handling, and simplified chunking all remain exactly the same - **we're just adjusting when conflation kicks in**.

For the custom series interface, this remains equally simple since we're still always merging exactly 2 items, regardless of why we're conflating.

### API Design Question:

Should this be:
1. A time scale option (`smoothingFactor`)
2. A per-series option (different series could have different smoothing)  
3. Both (series-level overrides global setting)

I lean toward **time scale option** for simplicity, since conflation is fundamentally about the display resolution and zoom level.

## Screenshots:
### No conflation
Screenshot 2025-11-05 at 21 39 02

### Normal conflation (smoothing factor of 1)
Screenshot 2025-11-05 at 21 40 42


### Extra smooth (factor of 4)
Screenshot 2025-11-05 at 21 42 06

Note: there is some loss of detail (heights of spikes, etc). We can address some smarter why to handle this in the future if needed.

illetid:
@SlicedSilver Thanks for the super detailed review and suggestions. 
The idea with recursive rebuilding super clever, I tweaked how we are working with remainder, now sure if it's better, so would be great to have a feedback on that. Also implemented the `conflationSmoothingFactor`, it is not working when we already have a maximum level of the confaltion, and it should be avoided for the bars series, since it creates gaps between bars, but it works well for the line series

src/api/options/series-options-defaults.ts
```diff
@@ -27,4 +27,5 @@ export const seriesOptionsDefaults: SeriesOptionsCommon = {
 		precision: 2,
 		minMove: 0.01,
 	},
+	conflationSmoothingFactor: undefined,
```
SlicedSilver: If it is `undefined` then can't we omit it here?

src/model/data-conflater.ts
```diff
@@ -0,0 +1,530 @@
...
+			} as unknown as SeriesPlotRow<T>;
+		} else {
+			return {
+				index: chunk.startIndex,
+				time: chunk.startTime,
+				originalTime: chunk.startTime,
+				value: [chunk.open, chunk.high, chunk.low, chunk.close],
+				originalDataCount: chunk.originalDataCount,
+			} as unknown as SeriesPlotRow<T>;
+		}
```
SlicedSilver: Can we shorten this? Perhaps like this:
```
const openOrClose = isCustomSeries && chunk.conflatedData ? chunk.close : chunk.open;

    const base = {
        index: chunk.startIndex,
        time: chunk.startTime,
        originalTime: chunk.startTime,
        value: [openOrClose, chunk.high, chunk.low, chunk.close],
        originalDataCount: chunk.originalDataCount,
    } as any;

    if (isCustomSeries && chunk.conflatedData) {
        base.data = chunk.conflatedData;
    }

    return base as SeriesPlotRow<T>;
```

src/model/series.ts
```diff
@@ -307,6 +346,162 @@ export class Series<T extends SeriesType> extends PriceDataSource implements IDe
...
+		const devicePixelRatio = window.devicePixelRatio || 1;
+		const globalSmoothing = timeScale.options().conflationSmoothingFactor;
+		const seriesSmoothing = this._options.conflationSmoothingFactor;
+		const effectiveSmoothing = seriesSmoothing ?? globalSmoothing ?? 1.0;
+
+		const conflationFactor = this._dataConflater.calculateConflationLevelWithSmoothing(
+			barSpacing,
+			devicePixelRatio,
+			effectiveSmoothing
+		);
```
SlicedSilver: This logic is repeated in a few places. Lets try reduce the amount of code used in the file by extract common code like this.

src/model/series.ts
```diff
@@ -307,6 +346,162 @@ export class Series<T extends SeriesType> extends PriceDataSource implements IDe
...
+		const globalSmoothing = timeScale.options().conflationSmoothingFactor;
+
+		// Get series-specific smoothing factor from series options
+		const seriesSmoothing = this._options.conflationSmoothingFactor;
+
+		// Series > Global > Default
+		const effectiveSmoothing = seriesSmoothing ?? globalSmoothing ?? 1;
+
+		// When using default conflation settings, check precomputed cache first
+		if (effectiveSmoothing === 1.0) {
```
SlicedSilver: Why the special treatment if effectiveSmoothing is `1.0`? It is possible that the smoothing factor is defined via the options before the data is even provided therefore the precompute step should be able to precompute with the smoothing factor.

src/model/data-conflater.ts
```diff
@@ -0,0 +1,530 @@
...
+			throw new Error(`Expected at least 2 items to merge, got ${items.length}`);
+		}
+
+		const firstItem = items[0];
+		const lastItem = items[items.length - 1];
+
+		if (!isCustomSeries || !customReducer || !priceValueBuilder) {
+			const high = Math.max(...items.map((item: SeriesPlotRow<T>) => item.value[PlotRowValueIndex.High]));
+			const low = Math.min(...items.map((item: SeriesPlotRow<T>) => item.value[PlotRowValueIndex.Low]));
+
```
SlicedSilver: My understanding is that this `_mergeItems` method will almost always be called with 2 items, and in a rare case be called with 3 items (1 remainder).
I think we should optimise this function to only be callable with 2 items, passed by reference in the parameters.

Reasons:
- Code like this with .max, spreading, mapping, etc... is more work for the CPU and memory since we are creating arrays. If it is 2 items then we can do direct comparison checks which will be faster.
- To pass data into this method requires slicing data to make new arrays. Again, this is extra work
- Since the case for 3 items is so rare, it is easier to handle it like `mergeItems(mergeItems(d1, d2), d3)` to get the same result of merging 3 items.

In general, I think there are quite a few places in this file where we can try take advantage of the 2 items assumption and prevent creating arrays for passing data around, or doing simple max/min calculations.

src/model/data-conflater.ts
```diff
@@ -0,0 +1,523 @@
...
+	public conflateByFactor(
+		data: readonly SeriesPlotRow<T>[],
+		barsToMerge: number,
+		customReducer?: CustomConflationReducer<HorzScaleItem>,
+		isCustomSeries: boolean = false,
+		priceValueBuilder?: (item: unknown) => number[]
+	): SeriesPlotRow<T>[] {
+		if (data.length === 0 || barsToMerge <= 1) {
+			return data.slice();
```
SlicedSilver: In a few places there are early returns like
```
return data.slice();
```
or
```
return originalData.slice();
```

Do we really need the `.slice()`? If not, rather return the data / originalData directly.

PR #1965, opened 2025-08-29
Fix price scale visible range log mode

### LINKED ISSUE #1963: Incorrect priceScale().getVisibleRange() values when in log scale.

**Steps/code to reproduce:**

1. Create a new plugin
2. In `IPrimitivePaneView.update()`, place `console.log(this._source.series.priceScale().getVisibleRange());`
3. Toggle log scale off/on in the chart options

**Screenshots:**

The prices returned from the visible price scale range are way off. Here are screenshots of the issue:

Regular mode:
Image

Log scale mode:
Image

Assuming I'm not misunderstanding the API usage, I have a fix for this in https://github.com/tradingview/lightweight-charts/compare/master...tpunt:lightweight-charts:fix-price-scale-visible-range-log-mode (need to add a test still)

tpunt:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #1963
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Fixes visible price range when in log scale mode.

I was not too sure how to best test this. Mocking `IChartWidgetBase` was a pain. The test in its current form still fails because of precision issues (100 != 99.99999999999994). Do you have any suggestions on how to best approach this? Thanks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,7 +88,7 @@ commands:
             echo 'export BRANCH_SPECIFIC_TEST=<< parameters.branchSpecificTest >>' >> $BASH_ENV
       - run:
           command: scripts/run-graphics-tests.sh
-          no_output_timeout: 20m
+          no_output_timeout: 5m
 
       - store_test_results:
           path: test-results/
--- a/indicator-examples/compile.mjs
+++ b/indicator-examples/compile.mjs
@@ -145,7 +145,7 @@ const compile = async () => {
 			}
 			const esModuleTyping = generateDtsBundle([
 				{
-					filePath: `./typings/indicator-examples/src/indicators/${stripCalculation(
+					filePath: `./typings/indicators/${stripCalculation(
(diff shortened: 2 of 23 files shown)


tpunt:
I used the price formatter to resolve the precision issues when applying the log formula. This seems to work well.

tpunt:
Hmm, the solution above doesn't work when `priceFormatter` is overridden in `ChartOptions`. This is because it could format the number with things like commas, causing NaN values . Example:
```
localization: {
    priceFormatter: (value: number): string => {
        return Intl.NumberFormat('en-GB', { maximumSignificantDigits: 15 }).format(value);
    },
},
```

Really, the solution needed here is to just align the number to `minMove`. I see that `PriceScale` does not expose this directly, but uses it internally. So I added a public method for it that can then be used for rounding when fetching the visible range.

SlicedSilver (approved):

PR #1977, opened 2025-09-16
add ability to include top layer to screenshot

illetid:
**Type of PR:** 
added two arguments to `takeScreenshot` function so we can export top layer primitives and crosshair


illetid:
Thanks for the suggestion with extracting duplicated logic, I'll create a separate PR for it

PR #2055, opened 2026-01-21
Fix resize jitter when using autoSize option

### LINKED ISSUE #1724: autoSize is not smooth
Please check this video 

https://youtu.be/q4TgSq-eZ-M

When resized from right to left its smooth.. but when resized left to right its shaking and not giving smooth effect.

I tried 

	new ResizeObserver((entries) => {
		if (entries.length === 0 || entries[0].target !== container) {
			return;
		}
		const newRect = entries[0].contentRect;
		chart.applyOptions({ height: newRect.height, width: newRect.width });
	}).observe(container);

, also tried autoSize : true option , 

even debounced , but when widht is increased on left side its causing this unpleasent experience.

DarrenHaba:
When a chart with autoSize enabled is inside a resizable container (e.g., a draggable splitter), resizing causes visible "jitter" - the chart content appears to jump left/right before snapping back to the correct position.

**Video demonstrating the bug:** https://youtu.be/VoDxlQ--4vY

**Video demonstrating the fix:** https://youtu.be/IeytRHZekq0

Root cause: The ResizeObserver callback called resize() which scheduled the canvas repaint to the next animation frame via fullUpdate(). This created a timing gap where the container had already resized but the canvas hadn't repainted yet, causing visual misalignment.

Fix:
1. Call resize() with forceRepaint=true from the ResizeObserver callback to paint synchronously rather than scheduling to the next frame
2. When forceRepaint is true, cancel any pending animation frames and merge pending invalidations before drawing
3. When autoSize is active, keep the outer element at 100%/100% instead of setting explicit pixel sizes, avoiding jitter from pixel rounding

Fixes #1724

**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #1724
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:** Modified src/gui/chart-widget.ts to paint synchronously when resize is triggered by ResizeObserver, eliminating the frame delay that caused visual jitter.

### DIFF
--- a/src/gui/chart-widget.ts
+++ b/src/gui/chart-widget.ts
@@ -202,14 +202,32 @@ export class ChartWidget<HorzScaleItem> implements IDestroyable, IChartWidgetBas
 		const heightStr = this._height + 'px';
 		const widthStr = this._width + 'px';
 
-		ensureNotNull(this._element).style.height = heightStr;
-		ensureNotNull(this._element).style.width = widthStr;
+		// When autoSize is enabled, keep outer element at 100%/100% to avoid jitter
+		// from pixel rounding differences between container and chart size
+		if (!this.autoSizeActive()) {
+			ensureNotNull(this._element).style.height = heightStr;
+			ensureNotNull(this._element).style.width = widthStr;
+		}
 
 		this._tableElement.style.height = heightStr;
 		this._tableElement.style.width = widthStr;
 
 		if (forceRepaint) {
-			this._drawImpl(InvalidateMask.full(), performance.now());
+			// Cancel any pending animation frame since we're doing a synchronous paint
+			if (this._drawRafId !== 0) {
+				window.cancelAnimationFrame(this._drawRafId);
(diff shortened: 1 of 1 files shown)


illetid:
Hi @DarrenHaba, thanks for the PR.
Switching to a 100%/100% element size to avoid jitter does indeed work, but it introduces a new behavior: the container element (`<div class="tv-lightweight-charts">`) can become slightly larger than the inner `table` and `canvas` elements. This happens because we floor the values to integers to achieve pixel-perfect rendering.

After some testing, we didn’t find any issues caused by this, and we think this trade-off is acceptable for the case shown in your video. I'll go ahead and merge the PR and have added this note as context so we can find it later if needed

illetid (approved):
LGTM

SlicedSilver:
Preview release available via:
```
npm i https://pkg.pr.new/lightweight-charts@02eaf8b
```

PR #2075, opened 2026-04-04
feat(series): add connectGaps option to line, area and baseline series

### LINKED ISSUE #699: Drawing a "gap" on chart when value is null/nan instead of 0
While drawing a line chart (especially, but some others might be a thing as well) there is a serious difference between "no value" (like null or NaN) and 0 - at the moment chart is drawing "no values" as 0. The only solution is to "split" one data series to many and draw a set of series.

this is how i'd see it to look like:
image

leo01102:
**Type of PR:** enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #699
- [x] Includes tests
- [x] Documentation update (added TSDoc comments to the options interface)

**Overview of change:**
This PR introduces a new `connectGaps: boolean` option (defaulting to `true`) for `Line`, `Area`, and `Baseline` series. 
When set to `false`, the chart will no longer draw a line connecting data points separated by one or more empty bars (time index distance > 1).

Key changes:
- Updated `walkLine` renderer utility to detect gaps and break path segments.
- Added `connectGaps` to `LineStyleOptions`, `AreaStyleOptions`, and `BaselineStyleOptions`.
- Added a new graphics test case to verify the behavior for all three series types.

Verified that existing charts are unaffected (100+ graphics tests passing). The implementation uses the time index difference to detect gaps as suggested in the original issue discussion.

The image below shows the new connectGaps: false setting in action, correctly breaking the line and area at missing data points.
1 golden

### DIFF
--- a/src/model/series-options.ts
+++ b/src/model/series-options.ts
@@ -231,6 +231,13 @@ export interface LineStyleOptions {
 	 * @defaultValue {@link LastPriceAnimationMode.Disabled}
 	 */
 	lastPriceAnimation: LastPriceAnimationMode;
+
+	/**
+	 * Connect gaps in data.
+	 *
+	 * @defaultValue `true`
+	 */
+	connectGaps?: boolean;
 }
 
 /**
@@ -351,6 +358,13 @@ export interface AreaStyleOptions {
 	 * @defaultValue {@link LastPriceAnimationMode.Disabled}
 	 */
 	lastPriceAnimation: LastPriceAnimationMode;
+
+	/**
+	 * Connect gaps in data.
+	 *
+	 * @defaultValue `true`
+	 */
+	connectGaps?: boolean;
 }
 
 /**
@@ -504,6 +518,13 @@ export interface BaselineStyleOptions {
 	 * @defaultValue {@link LastPriceAnimationMode.Disabled}
 	 */
 	lastPriceAnimation: LastPriceAnimationMode;
+
+	/**
+	 * Connect gaps in data.
+	 *
+	 * @defaultValue `true`
+	 */
+	connectGaps?: boolean;
 }
 
 /**
--- a/src/model/series/line-pane-view.ts
+++ b/src/model/series/line-pane-view.ts
@@ -25,6 +25,7 @@ export class SeriesLinePaneView extends LinePaneViewBase<'Line', LineStrokeItem,
 			pointMarkersRadius: options.pointMarkersVisible ? (options.pointMarkersRadius || options.lineWidth / 2 + 2) : undefined,
 			visibleRange: this._itemsVisibleRange,
 			barWidth: this._model.timeScale().barSpacing(),
+			connectGaps: options.connectGaps as boolean,
 		};
 
 		this._renderer.setData(data);
--- a/src/renderers/walk-line.ts
+++ b/src/renderers/walk-line.ts
@@ -1,7 +1,7 @@
 import { BitmapCoordinatesRenderingScope } from 'fancy-canvas';
 
 import { Coordinate } from '../model/coordinate';
-import { SeriesItemsIndexesRange } from '../model/time-data';
+import { SeriesItemsIndexesRange, TimePointIndex } from '../model/time-data';
 
 import { LinePoint, LineType } from './draw-line';
 
@@ -10,7 +10,7 @@ function distanceByCoordinates(p1x: number, p1y: number, p2x: number, p2y: numbe
 }
 
 // eslint-disable-next-line max-params, complexity
-export function walkLine<TItem extends LinePoint, TStyle extends CanvasRenderingContext2D['fillStyle' | 'strokeStyle']>(
+export function walkLine<TItem extends LinePoint & { time?: TimePointIndex }, TStyle extends CanvasRenderingContext2D['fillStyle' | 'strokeStyle']>(
 	renderingScope: BitmapCoordinatesRenderingScope,
 	items: readonly TItem[],
 	lineType: LineType,
@@ -20,7 +20,8 @@ export function walkLine<TItem extends LinePoint, TStyle extends CanvasRendering
 	// so if styleGetter returns objects, then styleGetter should return the same object for equal styles
 	styleGetter: (renderingScope: BitmapCoordinatesRenderingScope, item: TItem) => TStyle,
 	finishStyledArea: (renderingScope: BitmapCoordinatesRenderingScope, style: TStyle, areaFirstItem: LinePoint, newAreaFirstItem: LinePoint) => void,
-	dashPatternLength: number = 0
+	dashPatternLength: number = 0,
+	connectGaps: boolean = true
 ): void {
 	if (items.length === 0 || visibleRange.from >= items.length || visibleRange.to <= 0) {
 		return;
@@ -74,11 +75,22 @@ export function walkLine<TItem extends LinePoint, TStyle extends CanvasRendering
 			const currentY = currentItem.y * verticalPixelRatio;
 			const itemStyle = styleGetter(renderingScope, currentItem);
 
+			const prevItem = items[i - 1];
+			const isGap = !connectGaps && currentItem.time !== undefined && prevItem.time !== undefined && currentItem.time - prevItem.time > 1;
+
+			if (isGap) {
+				finishStyledArea(renderingScope, currentStyle, currentStyleFirstItem, prevItem);
+				ctx.beginPath();
+				currentStyle = itemStyle;
+				currentStyleFirstItem = currentItem;
+				ctx.moveTo(currentX, currentY);
+				continue;
+			}
+
 			switch (lineType) {
 				case LineType.Simple: {
 					ctx.lineTo(currentX, currentY);
 					if (shouldTrackDashOffset) {
-						const prevItem = items[i - 1];
 						const prevX = prevItem.x * horizontalPixelRatio;
 						const prevY = prevItem.y * verticalPixelRatio;
 						accumulatedDistance += distanceByCoordinates(prevX, prevY, currentX, currentY);
@@ -86,7 +98,6 @@ export function walkLine<TItem extends LinePoint, TStyle extends CanvasRendering
 					break;
 				}
 				case LineType.WithSteps: {
-					const prevItem = items[i - 1];
 					const prevY = prevItem.y * verticalPixelRatio;
 					ctx.lineTo(currentX, prevY);
 					if (shouldTrackDashOffset) {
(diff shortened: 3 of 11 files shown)


leo01102:
CI checks that fail are expected consequences of this PR, not regressions.

### `dts-changes` - API surface changed (expected)

This PR intentionally adds a new optional property to three interfaces:
- `LineStyleOptions.connectGaps?: boolean`
- `AreaStyleOptions.connectGaps?: boolean`
- `BaselineStyleOptions.connectGaps?: boolean`

The property is optional and defaults to `true`, preserving the existing behavior for all existing code.

### `graphics-tests` - New test case, no golden baseline (expected)

The new `connect-gaps` test case exercises `connectGaps: false`, which produces a visually different result on `master` (feature doesn't exist, so line draws continuously) vs. this PR (line breaks at gaps).

### Local Verification Results

- [x] **Unit Tests**: 144/144 passed
- [x] **Linting**: Passed (no errors)
- [x] **Type Check**: `tsc-verify` passed
- [x] **Graphics**: `connect-gaps` rendering verified locally (DPR 1.0-2.0)

src/model/series-options.ts
```diff
@@ -231,6 +231,13 @@ export interface LineStyleOptions {
 	 * @defaultValue {@link LastPriceAnimationMode.Disabled}
 	 */
 	lastPriceAnimation: LastPriceAnimationMode;
+
+	/**
+	 * Connect gaps in data.
+	 *
+	 * @defaultValue `true`
+	 */
+	connectGaps?: boolean;
```
SlicedSilver, line 240: **Suggestion:** `connectGaps` should be required (not optional).

Every other property in this interface (e.g. `pointMarkersVisible`, `lastPriceAnimation`) is declared as required, and this property always has a default value provided in `lineStyleDefaults`. Declaring it as optional forces every consumer to cast with `as boolean` — there are currently 6 instances of `options.connectGaps as boolean` across the pane views.

```suggestion
	connectGaps: boolean;
```

The JSDoc `@defaultValue` is also misleading alongside `?:` — it implies the value is always present (via defaults), but the type says it could be `undefined`.

Same change needed in `AreaStyleOptions` (line 367) and `BaselineStyleOptions` (line 527).
Once this is done, all six `as boolean` casts in `line-pane-view.ts`, `area-pane-view.ts`, and `baseline-pane-view.ts` should be removed.

src/model/series-options.ts
```diff
@@ -231,6 +231,13 @@ export interface LineStyleOptions {
 	 * @defaultValue {@link LastPriceAnimationMode.Disabled}
 	 */
 	lastPriceAnimation: LastPriceAnimationMode;
+
+	/**
+	 * Connect gaps in data.
+	 *
+	 * @defaultValue `true`
+	 */
```
SlicedSilver, line 239: The documentation should explain more precisely what constitutes a "gap". Users may be confused about the interaction with multiple series on the same chart. I'd suggest something like:
```suggestion
	/**
	 * Whether to draw a connecting line across gaps (whitespace) in the data.
	 *
	 * When `true`, the line will be drawn continuously even if the data contains
	 * whitespace entries. When `false`, the line will break at each whitespace gap.
	 *
	 * Note: Gaps are only detected where this series has explicit whitespace data
	 * (entries with only a `time` property and no value). If another series on the
	 * same chart has data at different time points, this will not create gaps in
	 * this series; add whitespace entries to this series' data to create gaps.
	 *
	 * @defaultValue `true`
	 */
```
Same JSDoc update needed for the identical blocks in `AreaStyleOptions` and `BaselineStyleOptions`.

src/renderers/walk-line.ts
```diff
@@ -74,19 +75,29 @@ export function walkLine<TItem extends LinePoint, TStyle extends CanvasRendering
 			const currentY = currentItem.y * verticalPixelRatio;
 			const itemStyle = styleGetter(renderingScope, currentItem);
 
+			const prevItem = items[i - 1];
+			const isGap = !connectGaps && currentItem.time !== undefined && prevItem.time !== undefined && currentItem.time - prevItem.time > 1;
```
SlicedSilver, line 79: Two issues with this gap detection:

1. **`undefined` checks are unnecessary**. As noted in the comment on the generic constraint, `time` is always present on these items. The `currentItem.time !== undefined && prevItem.time !== undefined` guards are dead code.
2. **`time - prevItem.time > 1` is too broad — it catches gaps from other series, not just whitespace** `TimePointIndex` is a global index shared across all series on a chart. When Series A has data at times `[Jan 1, Jan 3]` and Series B has data at `[Jan 1, Jan 2, Jan 3]`, the merged time scale gives indices `[0, 1, 2]`. Series A's items will be `[{time: 0}, {time: 2}]` — a difference of 2, which this check treats as a gap. But Series A never provided whitespace; the index gap comes from Series B contributing a time point that Series A doesn't have.

The intended behaviour (per discussion) is: only break the line where this series has **explicit whitespace data** between two points. If the user wants gaps for timestamps that only exist in another series, they should add whitespace entries to this series.

This means the gap detection can't be done purely at the renderer level by comparing `TimePointIndex` values. The information about whether whitespace existed between two points is lost by the time we reach `walkLine` — whitespace rows are filtered out in `DataLayer._setRowsToSeries()`.

One approach: during item construction in the pane view (e.g. in `LinePaneViewBase`), compute a boolean flag like `gapBefore` on each item by checking whether the series' original data contained whitespace rows between consecutive plot rows. This information is available at that layer and could be passed through to the renderer without adding complexity to `walkLine`'s parameters.

src/renderers/walk-line.ts
```diff
@@ -74,19 +75,29 @@ export function walkLine<TItem extends LinePoint, TStyle extends CanvasRendering
...
+
+			if (isGap) {
+				finishStyledArea(renderingScope, currentStyle, currentStyleFirstItem, prevItem);
+				ctx.beginPath();
+				currentStyle = itemStyle;
+				currentStyleFirstItem = currentItem;
+				ctx.moveTo(currentX, currentY);
+				continue;
+			}
```
SlicedSilver, line 88: **Bug: `accumulatedDistance` is not reset when breaking at a gap.**

When the line breaks at a gap, `finishStyledArea` is called and a new path begins, but `accumulatedDistance` (used for dash pattern offset tracking) is not reset. Compare with the `changeStyle` helper above (lines 52-65), which correctly handles this:

```typescript
if (shouldTrackDashOffset) {
    const offset = accumulatedDistance % dashPatternLength;
    ctx.lineDashOffset = offset;
    accumulatedDistance = offset;
}
```

This means that when `connectGaps: false` is combined with a dashed or dotted line style, the dash pattern after a gap will start at the wrong offset rather than resetting cleanly.

Suggested fix:  add after `ctx.beginPath();`:
```typescript
if (shouldTrackDashOffset) {
    accumulatedDistance = 0;
    ctx.lineDashOffset = 0;
}
```

src/model/series/line-pane-view.ts
```diff
@@ -25,6 +25,7 @@ export class SeriesLinePaneView extends LinePaneViewBase<'Line', LineStrokeItem,
 			pointMarkersRadius: options.pointMarkersVisible ? (options.pointMarkersRadius || options.lineWidth / 2 + 2) : undefined,
 			visibleRange: this._itemsVisibleRange,
 			barWidth: this._model.timeScale().barSpacing(),
+			connectGaps: options.connectGaps as boolean,
```
SlicedSilver, line 28: The `as boolean` cast here is a workaround for `connectGaps` being declared optional in the type interface. If `connectGaps` is made required in `LineStyleOptions` (see comment on `series-options.ts`), this cast should be removed.

Same applies to the casts in `area-pane-view.ts` (lines 56 and 67) and `baseline-pane-view.ts` (lines 71 and 85).

tests/e2e/graphics/test-cases/series/connect-gaps.js
SlicedSilver, line 1: The test covers the `connectGaps: false` case well for all three series types, but some additional coverage would be valuable:

- **`connectGaps: true` (default) with whitespace data** — verifying the line connects across gaps (regression test for the default behaviour).
- **Dashed/dotted line styles with `connectGaps: false`** — this exercises the `accumulatedDistance` / `lineDashOffset` code path which currently has a bug (see comment on `walk-line.ts`).
- **`LineType.Curved` with `connectGaps: false`** — the bezier control point calculation in `getControlPoints()` uses adjacent items. After a gap break + `moveTo`, the first curved segment of the new sub-path might produce unexpected control points since `getControlPoints` still sees the pre-gap item in the array.
- **Multiple series with different time ranges and `connectGaps: false`** — verifying that a series doesn't get spurious gaps from another series' timestamps (this relates to the gap detection logic concern in `walk-line.ts`).

src/renderers/walk-line.ts
```diff
@@ -10,7 +10,7 @@ function distanceByCoordinates(p1x: number, p1y: number, p2x: number, p2y: numbe
 }
 
 // eslint-disable-next-line max-params, complexity
-export function walkLine<TItem extends LinePoint, TStyle extends CanvasRenderingContext2D['fillStyle' | 'strokeStyle']>(
+export function walkLine<TItem extends LinePoint & { time?: TimePointIndex }, TStyle extends CanvasRenderingContext2D['fillStyle' | 'strokeStyle']>(
```
SlicedSilver, line 13: The `{ time?: TimePointIndex }` intersection is unnecessary and over-permissive. Every caller passes items that extend `TimedValue` (via `LineItemBase` or `AreaFillItemBase`), and `TimedValue` already has `time: TimePointIndex` as a **required** property. This optional annotation is what forces the defensive `!== undefined` checks on line 79.

```suggestion
export function walkLine<TItem extends LinePoint & TimedValue, TStyle extends CanvasRenderingContext2D['fillStyle' | 'strokeStyle']>(
```

With `TimedValue` (already exported from `../model/time-data`), the import can stay the same, `TimedValue` already brings in `TimePointIndex`. And the gap check on line 79 can be simplified to remove the `undefined` guards (see comment on that line).

SlicedSilver (changes requested):
Thanks for the PR @leo01102 
I've left some suggestions which should hopefully be clear enough.

One key behaviour difference I'm recommending is within the gap detection logic so that it should only consider gaps when a series has explicit whitespace added (instead of just a jump in the time index). If you would also like the gap detection to cover those other cases (gaps created due to other series with different time points) then I think we should have an option to enable that so that the developer has more control over the feature. Requiring explicit whitespace to be added to the series is the most robust and flexible approach for now; however it may not be the most convenient depending on your use-case.

PR #2095, opened 2026-06-09
add a11y plugin example

illetid:
## Summary

Adds a drop-in Accessibility plugin under `plugin-examples/`
 
It attaches one labelled, focusable primitive per pane, so multi-pane / multi-series charts work out of the box. detach() restores the DOM to exactly how it was found.

### What it does

- Semantic layer — each pane gets a focusable overlay (role="application", aria-label, aria-roledescription, aria-describedby); the canvases are hidden from assistive tech (aria-hidden), the layout table is marked presentational, and focusable chart internals (e.g. the attribution link) are taken out of the tab order.
- Keyboard navigation — ←/→ move between data points (the viewport pages so the whole series is reachable), ↑/↓ switch series, Page Up/Down jump, Home/End first/last, Enter/Space read a summary, H lists the controls.
- ARIA-live announcements — a per-pane assertive region for navigation, and a single shared polite region for background data updates so multi-pane updates never talk over each other (default announces only the active/last-focused pane; configurable to combine all panes).
- Visible focus indicator — an optional ring that tracks the active point and stays aligned with the canvas while scrolling/zooming (WCAG 2.4.7).
- Localization — every announced string is overridable via a messages bundle; dates/numbers follow the chart's localization.locale (and its priceFormatter/timeFormatter if set); a lang attribute is set on the regions so screen readers use the right voice.
- Lightweight & event-driven — "active-point-only" strategy (never mirrors data points into the DOM); data sync is driven by subscribeDataChanged, so scrolling/zooming does no data work.

### Notes & tradeoffs

- `role="application"` is used so arrow keys drive data navigation under screen readers; this intentionally suspends the SR virtual cursor while focused on the chart (press Tab to move on — it is not a focus trap). Documented in the README.

### DIFF
--- a/plugin-examples/src/plugins/accessibility/accessibility.ts
+++ b/plugin-examples/src/plugins/accessibility/accessibility.ts
@@ -133,7 +133,11 @@ export interface AccessibilityOptions {
 	 * `aria-live` region.
 	 */
 	announceDataUpdates: boolean;
-	/** Number of points to jump when using `PageUp` / `PageDown`. */
+	/**
+	 * Number of points to jump when using `PageUp` / `PageDown`. `PageUp` moves
+	 * forward in time and `PageDown` back, following the ARIA slider convention
+	 * that `PageUp` increases the value.
+	 */
 	pageStep: number;
 	/**
 	 * Controls what the on-demand summary (`Enter` / `Space`) and the data-update
@@ -228,8 +232,14 @@ function defaultTimeFormatter(time: Time, locale?: string): string {
 		// BusinessDay months are 1-12, whereas Date expects 0-11.
 		date = new Date(Date.UTC(time.year, time.month - 1, time.day));
 	} else {
-		// Business-day string, e.g. '2019-05-15'.
-		return time;
+		// Business-day string, e.g. '2019-05-15' – parse it so it is localised
+		// like the other time formats (with a verbatim fallback if it does not
+		// match the expected YYYY-MM-DD shape).
+		const [year, month, day] = time.split('-').map(Number);
+		if (!year || !month || !day) {
+			return time;
+		}
+		date = new Date(Date.UTC(year, month - 1, day));
 	}
 	// `locale || undefined` guards against the empty-string locale used server-side,
 	// which is not a valid Intl locale.
@@ -362,9 +372,11 @@ export const defaultMessages: AccessibilityMessages = {
 			? 'Use the left and right arrow keys to move between data points, and the up and down arrows to switch series.'
 			: 'Use the left and right arrow keys to move between data points.',
 	help: ({ multiSeries }): string =>
-		`Keyboard controls. Left and right arrows move between data points. ${multiSeries ? 'Up and down arrows switch between series. ' : ''}Page Up and Page Down jump ten points. Home and End jump to the first and last points. Enter or Space reads a summary of the series.`,
+		`Keyboard controls. Left and right arrows move between data points. ${multiSeries ? 'Up and down arrows switch between series. ' : ''}Page Up jumps ten points forward, Page Down ten points back. Home and End jump to the first and last points. Enter or Space reads a summary of the series.`,
+	// Most important information first: the value, then the date; the position
+	// counter is context, so it comes last.
 	point: ({ position, total, time, label, values }): string =>
-		`Point ${position} of ${total}. ${time}, ${label} ${values}.`,
+		`${label} ${values}, ${time}. Point ${position} of ${total}.`,
 	seriesPosition: ({ label, position, total, point }): string =>
 		`${label}, series ${position} of ${total}.${point}`,
 	ohlcValues: ({ labels, open, high, low, close }): string => {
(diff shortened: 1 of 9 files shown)


website/tutorials/a11y/conclusion.mdx
```diff
@@ -46,6 +46,30 @@ discussed in this tutorial.
...
+large data sets. Use `dataScope: 'all'` when you want those summaries to describe
+the full history.
+
+It is also fully translatable: numbers and dates follow the chart's
+`localization.locale`, and every announced string can be overridden through a
+`messages` bundle (with a `lang` attribute so screen readers use the right
+voice). See the plugin's README and its Spanish example.
+
+:::
+
```
SlicedSilver, line 72: I would suggest that we also mention the plugin on the first page of the A11y tutorial. Most people would be happy to see that and skip the tutorial by just using a plugin instead.

plugin-examples/src/plugins/accessibility/example/example.ts
```diff
@@ -0,0 +1,105 @@
...
+	AreaSeries,
+	HistogramSeries,
+	LineSeries,
+	Time,
+	createChart,
+} from 'lightweight-charts';
+import { generateLineData } from '../../../sample-data';
+import { addAccessibilityPlugin } from '../accessibility';
+
+const chart = ((window as unknown as any).chart = createChart('chart', {
```
SlicedSilver: I think we can remove `= ((window as unknown as any).chart =` from the examples

plugin-examples/src/plugins/accessibility/accessibility.ts
```diff
@@ -0,0 +1,1731 @@
...
+	let date: Date;
+	if (isUTCTimestamp(time)) {
+		date = new Date(time * 1000);
+	} else if (isBusinessDay(time)) {
+		// BusinessDay months are 1-12, whereas Date expects 0-11.
+		date = new Date(Date.UTC(time.year, time.month - 1, time.day));
+	} else {
+		// Business-day string, e.g. '2019-05-15'.
+		return time;
+	}
```
SlicedSilver, line 317: `defaultTimeFormatter` returns business-day string times verbatim (unlocalised) while timestamp/BusinessDay-object times get `toLocaleDateString`. Edge case, probably acceptable, but slightly inconsistent localisation.

plugin-examples/src/plugins/accessibility/accessibility.ts
```diff
@@ -0,0 +1,1731 @@
...
+				this._moveSeries(1);
+				break;
+			case 'PageUp':
+				event.preventDefault();
+				this._movePoint(this._options.pageStep);
+				break;
+			case 'PageDown':
+				event.preventDefault();
+				this._movePoint(-this._options.pageStep);
+				break;
```
SlicedSilver, line 1340: `PageUp` moves `+pageStep` (forward/right in time), `PageDown` moves backward. This is internally consistent with ARIA slider semantics (PageUp increases the value), so I wouldn't call it a bug — but on a left-to-right time axis some users expect PageDown to advance. The README just says "jump ten points" without direction. A one-line note on direction in the keyboard table would remove the ambiguity.

plugin-examples/src/plugins/accessibility/accessibility.ts
```diff
@@ -0,0 +1,1731 @@
...
+				this._movePoint(-1);
+				break;
+			case 'ArrowUp':
+				event.preventDefault();
+				this._moveSeries(-1);
+				break;
+			case 'ArrowDown':
+				event.preventDefault();
+				this._moveSeries(1);
+				break;
```
SlicedSilver, line 1332: If you have multiple series (and they don't have all the same timestamps, such as a moving average line not having the first 10 points) then the up / down keys won't select the same points (time) on the different series.

This also means that we don't always scroll the selected point into view when switching series because we assumed that they would have the same index on the time scale but this is incorrect.

plugin-examples/src/plugins/accessibility/accessibility.ts
```diff
@@ -0,0 +1,1731 @@
...
+		return `${title}. ${seriesPart}Press H for keyboard help.`;
+	},
+	description: ({ multiSeries }): string =>
+		multiSeries
+			? 'Use the left and right arrow keys to move between data points, and the up and down arrows to switch series.'
+			: 'Use the left and right arrow keys to move between data points.',
+	help: ({ multiSeries }): string =>
+		`Keyboard controls. Left and right arrows move between data points. ${multiSeries ? 'Up and down arrows switch between series. ' : ''}Page Up and Page Down jump ten points. Home and End jump to the first and last points. Enter or Space reads a summary of the series.`,
+	point: ({ position, total, time, label, values }): string =>
+		`Point ${position} of ${total}. ${time}, ${label} ${values}.`,
```
SlicedSilver: When moving to different points (up and down arrows), I think we should mention the price first, then the date, and then the 'point 447 of 500' at the end. Mentioning the point 447... at the start doesn't make sense to me because it is the least important piece of information.

SlicedSilver (commented):
Thanks, this looks really good and extends beyond what was previously mentioned in the tutorial.

plugin-examples/src/plugins/accessibility/example/example.ts
illetid: removed

website/tutorials/a11y/conclusion.mdx
illetid: added a section at the first page

plugin-examples/src/plugins/accessibility/accessibility.ts
illetid: nice catch, fixed that one and changed moving average to be more realistic in the example

SlicedSilver (approved):

PR #2098, opened 2026-06-14
fix: restore autoscale when autoscaleInfoProvider is reset to undefined

### LINKED ISSUE #2080: Autoscale doesn't work after enabling and disabling fixed autoscaleInfoProvider
Lightweight Charts™ Version: last

When creating a chart with
`autoscaleInfoProvider: undefined`, autoscale works as expected.
Then, if I setup 
```
autoscaleInfoProvider:function () {
return {
   priceRange: {
      minValue: 0,
      maxValue: 100
   },
}}

```
and then set `autoscaleInfoProvider: undefined` back, autoscale will no longer work.

Look at it here [https://jsfiddle.net/sas2006/u7s2nvL3/4/](https://jsfiddle.net/sas2006/u7s2nvL3/4/)
1. Run
2. Move the chart - autoscale works
3. Check "Use fixed autoscaleInfoProvider 0..100" checkbox - autoscale do not works
4. uncheck "Use fixed autoscaleInfoProvider 0..100" checkbox - **autoscale still doesn't work!** **And it is no longer possible to turn it on.**

0007aadil:
**Type of PR:** bugfix

**PR checklist:**

- [x] Addresses an existing issue: fixes #2080
- [ ] Includes tests
- [ ] Documentation update

**Overview of change:**

When `autoscaleInfoProvider` is set to a custom function and then reset to `undefined`, autoscale permanently stops working.

Root cause: the internal `merge()` helper skips `undefined` values (by design), so the old provider stays in `this._options` even after the user explicitly resets it.

Fix: after `merge()` runs, use `Object.prototype.hasOwnProperty.call()` to detect if `autoscaleInfoProvider` was explicitly passed as `undefined`, and if so, reset it directly on `this._options`.

I encountered this while building a real-time stock trading dashboard using lightweight-charts and confirmed the fix restores correct autoscale behavior.

Whether `Object.prototype.hasOwnProperty.call()` is the preferred pattern here vs any other approach to detect explicit `undefined` in partial options.


SlicedSilver (dismissed):
Approving, however I will add a test case as well

SlicedSilver:
@0007aadil Apologies for the confusing nature of our test cases. It is expected that a new visual test case fails, this means that the visual test is correctly testing the behaviour (and that it has been fixed). Can you please re-add the visual test case (we can keep the new interaction test as well)?

0007aadil:
Thanks for clarifying 👍🏻....I misunderstood how the graphics tests work here  makes sense that a failing visual test confirms the fix. I'll re-add the graphics test case alongside the interactions test.

0007aadil:
@SlicedSilver  Re-added the graphics test in 0e572e6 alongside the interactions test ready for review whenever you get a chance

SlicedSilver (approved):

SlicedSilver:
Thanks @0007aadil

PR #2108, opened 2026-07-28
fix: better curved line rendering

### LINKED ISSUE #1680: Optimize the algorithm for curved line
The display effect of the curved line is not good, it is recommended to optimize the algorithm. ↓
image

It will cause strange sharp corners to appear.
image

Expectations are like this(echarts).

image

edew:
Replace the Catmull-Rom logic used for curved lines with Fritsch-Carlson monotone, the algorithm used for d3-shape's curveMonotoneX and Chart.js's 'monotone' mode.

| Golden | Test |
|--------|--------|
| 1 golden | 2 test  |

**Type of PR:** bugfix, enhancement

**PR checklist:**

- [x] Addresses an existing issue: fixes #1680
- [x] Includes tests
- [ ] Documentation update

**Overview of change:**

Note: a segment is a line between two points.

With the monotone algorithm the curve never leaves the y-range of each segment's end points, so plateaus stay flat without spikes. At the beginning (end) of the series the missing neighbour segment is replaced with a copy of the first (last) segment, so the curve enters and leaves the line pointing along its first (last) segment, and a two point series is a straight line.

The signature of getControlPoints has been simplified to take the index of the segment's end point instead of two point arguments. `getControlPoints(points, endPointIndex)` calculates control points for the segment from `points[endPointIndex - 1]` to `points[endPointIndex]`. The old signature allowed arbitrary point pairs, but the neighbour lookups only make sense for adjacent indices, so non-adjacent input would produce rubbish.

The existing curved-line graphics tests will have diffs as expected. The new curved-line-plateau case covers the specific GitHub issue data, a spiky plateau, and a zigzag.

The BaselineSeries size limit increased slightly from 4.5KB to ~4.56 KB.

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -95,7 +95,7 @@ export default [
 		path: 'dist/lightweight-charts.production.mjs',
 		import: '{ BaselineSeries }',
 		ignore: ['fancy-canvas'],
-		limit: '4.5 KB',
+		limit: '4.56 KB',
 		brotli: true,
 	},
 	{
--- a/src/renderers/line-hit-test.ts
+++ b/src/renderers/line-hit-test.ts
@@ -52,7 +52,7 @@ function lineSegmentHorizontalBounds(
 ): [number, number] {
 	switch (lineType) {
 		case LineType.Curved: {
-			const [firstControlPoint, secondControlPoint] = getControlPoints(items, toItemIndex - 1, toItemIndex);
+			const [firstControlPoint, secondControlPoint] = getControlPoints(items, toItemIndex);
 			const minX = Math.min(firstItem.x, secondItem.x, firstControlPoint.x, secondControlPoint.x);
 			const maxX = Math.max(firstItem.x, secondItem.x, firstControlPoint.x, secondControlPoint.x);
 			return [minX, maxX];
@@ -87,7 +87,7 @@ function hitTestLineSegment(
 				return minDistance <= radius ? minDistance : null;
 			}
 		case LineType.Curved: {
-			const [firstControlPoint, secondControlPoint] = getControlPoints(items, toItemIndex - 1, toItemIndex);
+			const [firstControlPoint, secondControlPoint] = getControlPoints(items, toItemIndex);
 			const distance = distanceToBezierCurve(x, y, [firstItem, firstControlPoint, secondControlPoint, secondItem]);
 			return distance <= radius ? distance : null;
 		}
--- a/src/renderers/walk-line.ts
+++ b/src/renderers/walk-line.ts
@@ -105,7 +105,7 @@ export function walkLine<TItem extends LinePoint, TStyle extends CanvasRendering
 					break;
 				}
 				case LineType.Curved: {
-					const [cp1, cp2] = getControlPoints(items, i - 1, i);
+					const [cp1, cp2] = getControlPoints(items, i);
 					const cp1x = cp1.x * horizontalPixelRatio;
 					const cp1y = cp1.y * verticalPixelRatio;
 					const cp2x = cp2.x * horizontalPixelRatio;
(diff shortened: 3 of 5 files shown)


src/renderers/line-hit-test.ts
SlicedSilver, line 161:
```suggestion
		// Items are sorted by x, and every line type (including curved - its control points always
		// lie within the segment) stays within its segment horizontally, so the item xs bound the segment.
		if (!isWithinHorizontalSweep(x, previousItem.x, currentItem.x, radius)) {
			continue;
		}
```

SlicedSilver, line 68: We can delete this entire function now. With the new algorithm, control-point x's **always** lie inside the segment (`start.x + w/3`, `end.x − w/3`), so the `Curved` case in `lineSegmentHorizontalBounds` is now redundant.

This saves up to `60B` for the individual series

.size-limit.js
```diff
@@ -95,7 +95,7 @@ export default [
 		path: 'dist/lightweight-charts.production.mjs',
 		import: '{ BaselineSeries }',
 		ignore: ['fancy-canvas'],
-		limit: '4.5 KB',
+		limit: '4.56 KB',
```
SlicedSilver, line 98: We can go back to `limit: '4.5 KB',` since we improved the bundle size in other places

src/renderers/walk-line.ts
```diff
@@ -149,28 +149,83 @@ export function walkLine<TItem extends LinePoint, TStyle extends CanvasRendering
...
+	return width !== 0 ? (p2.y - p1.y) / width : 0;
 }
 
-function add(p1: LinePoint, p2: LinePoint): LinePoint {
-	return { x: p1.x + p2.x as Coordinate, y: p1.y + p2.y as Coordinate };
-}
+/**
+ * This is the Fritsch-Carlson formula for monotone interpolation.
+ * https://en.wikipedia.org/wiki/Monotone_cubic_interpolation
+ */
```
SlicedSilver, line 160: Apparently this is actually the "Steffen's method".
```
/**
 * This is the slope limiter from Steffen's method for monotone interpolation.
 * Steffen, M. (1990), "A simple method for monotonic interpolation in one dimension", Astron. Astrophys. 239, 443.
 */
```
Steffens doesn't overshoot while the Fritsch-Carlson formula can still have some overshooting

src/renderers/walk-line.ts
SlicedSilver, line 1: With some code-golf, the curved path calculation can be simplified further to reduce the bundle size a bit more.

```ts
import { BitmapCoordinatesRenderingScope } from 'fancy-canvas';

import { Coordinate } from '../model/coordinate';
import { SeriesItemsIndexesRange } from '../model/time-data';

import { LinePoint, LineType } from './draw-line';

function distanceByCoordinates(p1x: number, p1y: number, p2x: number, p2y: number): number {
	return Math.hypot(p2x - p1x, p2y - p1y);
}

// eslint-disable-next-line max-params, complexity
export function walkLine<TItem extends LinePoint, TStyle extends CanvasRenderingContext2D['fillStyle' | 'strokeStyle']>(
	renderingScope: BitmapCoordinatesRenderingScope,
	items: readonly TItem[],
	lineType: LineType,
	visibleRange: SeriesItemsIndexesRange,
	barWidth: number,
	// the values returned by styleGetter are compared using the operator !==,
	// so if styleGetter returns objects, then styleGetter should return the same object for equal styles
	styleGetter: (renderingScope: BitmapCoordinatesRenderingScope, item: TItem) => TStyle,
	finishStyledArea: (renderingScope: BitmapCoordinatesRenderingScope, style: TStyle, areaFirstItem: LinePoint, newAreaFirstItem: LinePoint) => void,
	dashPatternLength: number = 0
): void {
	if (items.length === 0 || visibleRange.from >= items.length || visibleRange.to <= 0) {
		return;
	}

	const { context: ctx, horizontalPixelRatio, verticalPixelRatio } = renderingScope;

	const firstItem = items[visibleRange.from];
	let currentStyle = styleGetter(renderingScope, firstItem);
	let currentStyleFirstItem = firstItem;

	if (visibleRange.to - visibleRange.from < 2) {
		const halfBarWidth = barWidth / 2;

		ctx.beginPath();

		const item1: LinePoint = { x: firstItem.x - halfBarWidth as Coordinate, y: firstItem.y };
		const item2: LinePoint = { x: firstItem.x + halfBarWidth as Coordinate, y: firstItem.y };

		ctx.moveTo(item1.x * horizontalPixelRatio, item1.y * verticalPixelRatio);
		ctx.lineTo(item2.x * horizontalPixelRatio, item2.y * verticalPixelRatio);

		finishStyledArea(renderingScope, currentStyle, item1, item2);
	} else {
		const shouldTrackDashOffset = dashPatternLength > 0;
		let accumulatedDistance = 0;

		const changeStyle = (newStyle: TStyle, currentItem: TItem) => {
			finishStyledArea(renderingScope, currentStyle, currentStyleFirstItem, currentItem);

			ctx.beginPath();
			currentStyle = newStyle;
			currentStyleFirstItem = currentItem;

			if (shouldTrackDashOffset) {
				const offset = accumulatedDistance % dashPatternLength;
				ctx.lineDashOffset = offset;
				// reset to the remainder to avoid floating-point precision drift over very long series.
				accumulatedDistance = offset;
			}
		};

		let currentItem = currentStyleFirstItem;

		ctx.beginPath();
		ctx.moveTo(firstItem.x * horizontalPixelRatio, firstItem.y * verticalPixelRatio);

		for (let i = visibleRange.from + 1; i < visibleRange.to; ++i) {
			currentItem = items[i];
			const currentX = currentItem.x * horizontalPixelRatio;
			const currentY = currentItem.y * verticalPixelRatio;
			const itemStyle = styleGetter(renderingScope, currentItem);

			switch (lineType) {
				case LineType.Simple: {
					ctx.lineTo(currentX, currentY);
					if (shouldTrackDashOffset) {
						const prevItem = items[i - 1];
						const prevX = prevItem.x * horizontalPixelRatio;
						const prevY = prevItem.y * verticalPixelRatio;
						accumulatedDistance += distanceByCoordinates(prevX, prevY, currentX, currentY);
					}
					break;
				}
				case LineType.WithSteps: {
					const prevItem = items[i - 1];
					const prevY = prevItem.y * verticalPixelRatio;
					ctx.lineTo(currentX, prevY);
					if (shouldTrackDashOffset) {
						accumulatedDistance += Math.abs(currentItem.x - prevItem.x) * horizontalPixelRatio;
					}

					if (itemStyle !== currentStyle) {
						changeStyle(itemStyle, currentItem);
						ctx.lineTo(currentX, prevY);
					}

					ctx.lineTo(currentX, currentY);
					if (shouldTrackDashOffset) {
						accumulatedDistance += Math.abs(currentItem.y - prevItem.y) * verticalPixelRatio;
					}
					break;
				}
				case LineType.Curved: {
					const [cp1, cp2] = getControlPoints(items, i);
					const cp1x = cp1.x * horizontalPixelRatio;
					const cp1y = cp1.y * verticalPixelRatio;
					const cp2x = cp2.x * horizontalPixelRatio;
					const cp2y = cp2.y * verticalPixelRatio;
					ctx.bezierCurveTo(
						cp1x,
						cp1y,
						cp2x,
						cp2y,
						currentX,
						currentY
					);

					if (shouldTrackDashOffset) {
						const prevItem = items[i - 1];
						const prevX = prevItem.x * horizontalPixelRatio;
						const prevY = prevItem.y * verticalPixelRatio;
						const chord = distanceByCoordinates(prevX, prevY, currentX, currentY);
						const controlPolygon = distanceByCoordinates(prevX, prevY, cp1x, cp1y) +
							distanceByCoordinates(cp1x, cp1y, cp2x, cp2y) +
							distanceByCoordinates(cp2x, cp2y, currentX, currentY);
						accumulatedDistance += (chord + controlPolygon) / 2;
					}
					break;
				}
			}

			if (lineType !== LineType.WithSteps && itemStyle !== currentStyle) {
				changeStyle(itemStyle, currentItem);
				ctx.moveTo(currentX, currentY);
			}
		}

		if (currentStyleFirstItem !== currentItem || currentStyleFirstItem === currentItem && lineType === LineType.WithSteps) {
			finishStyledArea(renderingScope, currentStyle, currentStyleFirstItem, currentItem);
		}

		if (shouldTrackDashOffset) {
			ctx.lineDashOffset = 0;
		}
	}
}

function segmentSlope(p1: LinePoint, p2: LinePoint): number {
	const width = p2.x - p1.x;
	return width !== 0 ? (p2.y - p1.y) / width : 0;
}

/**
 * This is the slope limiter from Steffen's method for monotone interpolation.
 * Steffen, M. (1990), "A simple method for monotonic interpolation in one dimension", Astron. Astrophys. 239, 443.
 */
function monotoneSlope(
	leftWidth: number,
	leftSlope: number,
	rightWidth: number,
	rightSlope: number
): number {
	// Points on a line always have increasing x coordinates, so segments always have a width that is greater than zero.
	// If the input does not obey this rule a slope calculation is not possible so we make the curve horizontal at the shared point.
	if (leftWidth <= 0 || rightWidth <= 0) {
		return 0;
	}

	// The slope of a segment is a good estimate of the slope of the curve at the center of that
	// segment. Thus the two segment slopes give the slope of the curve at two known positions.
	// A linear interpolation between the two known slopes gives the slope at the shared point.
	// The center of the narrow segment is nearer to the shared point than the center of the wide
	// segment. Thus the slope of the narrow segment must get the larger weight. This is the
	// reason that each slope is multiplied by the width of the other segment.
	const weightedSlope =
		(leftSlope * rightWidth + rightSlope * leftWidth) /
		(leftWidth + rightWidth);
	// `Math.sign(left) + Math.sign(right)` is +2 or -2 when the slopes are in the same up/down direction, and 0 when they don't.
	return (
		(Math.sign(leftSlope) + Math.sign(rightSlope)) *
		Math.min(Math.abs(leftSlope), Math.abs(rightSlope), 0.5 * Math.abs(weightedSlope))
	);
}

/**
 * Calculates the slope the curve has as it passes through `point`, one end of the segment described
 * by `width` and `slope`. `neighbour` is the point on the far side of `point` from that segment, or
 * `null` where the line begins or ends. Without a neighbouring segment we pretend the line continues
 * with a copy of the drawn segment, so the curve enters/leaves the line pointing straight along its
 * first/last segment (and a two point series is a straight line). Note that `monotoneSlope` is
 * symmetric in its two (width, slope) pairs, so the same helper serves both ends of the segment.
 */
function slopeThroughPoint(point: LinePoint, neighbour: LinePoint | null, width: number, slope: number): number {
	return neighbour === null
		? slope
		: monotoneSlope(Math.abs(neighbour.x - point.x), segmentSlope(point, neighbour), width, slope);
}

/**
 * A control point `offset` into the segment horizontally, shifted vertically so that the curve
 * passes through `point` with slope `slope`.
 */
function controlPoint(point: LinePoint, offset: number, slope: number): LinePoint {
	return {
		x: point.x + offset as Coordinate,
		y: point.y + offset * slope as Coordinate,
	};
}

/**
 * Calculates the control points needed to draw the curve between the point at `endPointIndex - 1` and the point at `endPointIndex`.
 *
 * @returns Two control points that can be used as arguments to {@link CanvasRenderingContext2D.bezierCurveTo} to draw the curve segment.
 */
export function getControlPoints(
	points: readonly LinePoint[],
	endPointIndex: number
): [LinePoint, LinePoint] {
	// The segment (line between two points) being drawn...
	const startPoint = points[endPointIndex - 1];
	const endPoint = points[endPointIndex];
	const width = endPoint.x - startPoint.x;
	const slope = segmentSlope(startPoint, endPoint);

	// ...and its neighbouring points, absent where the line begins or ends.
	const pointBeforeStart = endPointIndex > 1 ? points[endPointIndex - 2] : null;
	const pointAfterEnd = endPointIndex < points.length - 1 ? points[endPointIndex + 1] : null;

	// Each control point sits a third of the way into the segment.
	const offset = width / 3;
	return [
		controlPoint(startPoint, offset, slopeThroughPoint(startPoint, pointBeforeStart, width, slope)),
		controlPoint(endPoint, -offset, slopeThroughPoint(endPoint, pointAfterEnd, width, slope)),
	];
}

```

This change saves `20 B` which is not a big change because the duplication in code that it reduces still compresses well.

SlicedSilver (commented):
Thanks Ed. 🏆
This is a much better approach for curving lines for financial charts because it avoids the overshoots. Looks really nice. 🎉

PR #2114, opened 2026-08-31
switch to pnpm workspace

illetid:


illetid:
> ci/circleci: install-deps-website

needs to be removed from the protected branches rules

.circleci/config.yml
```diff
@@ -102,25 +110,18 @@ jobs:
...
-      - checkout-with-deps
-      - run: cd website && npm install
-      - save_cache: *save-node-modules-cache-website
-
   build:
     executor: node-executor
     steps:
       - checkout-with-deps
-      - run: npm run build:prod
+      - run: pnpm run build:prod
```
SlicedSilver: with pnpm you don't normally need to write 'run'

```suggestion
      - run: pnpm build:prod
```

SlicedSilver: and in the other places as well...

.github/workflows/deploy.yml
```diff
@@ -20,38 +23,39 @@ jobs:
...
+      # its uniqueName (and with it every chunk's content hash) from it, so
+      # adding one would needlessly invalidate all cached site assets.
+      # It is addressed by directory instead of by --filter name.
       - name: Build documentation website
-        working-directory: ./website
-        run: npm run build
-      - name: Install plugin example dependencies
-        working-directory: ./plugin-examples
-        run: npm install
+        run: pnpm --dir website run build
```
SlicedSilver: Here we are moving into the `website` folder and running build; but in other parts we are rather using the `--filter` flag. Is this intentional? I assume it is but wanted to check.

We could put `"name": "lightweight-charts-website",` inside the website package.json and then rather use filter here.

scripts/check-dts-changes.sh
```diff
@@ -12,6 +17,28 @@ function print_step {
...
+    if [ -d node_modules/.pnpm ]; then
+      echo "Removing pnpm-created node_modules before npm install"
+      rm -rf node_modules
+    fi
+    # This script only builds typings, it never launches a browser,
+    # so skip the browser download in puppeteer's postinstall.
+    PUPPETEER_SKIP_DOWNLOAD=1 npm install
+  fi
+}
+
```
SlicedSilver, line 45: Can we mark code like this for deletion after merging into master?

.circleci/config.yml
illetid: removed the `run` only one tricky place is `pnpm run deploy` in the docs, it needs to be called with `run` because `pnpm deploy` is reserved in pnpm

.github/workflows/deploy.yml
illetid: name was omitted intentionally to have the byte identical output, but I think it's better to add it now (and I verified it before adding name), fixed

scripts/check-dts-changes.sh
illetid: added todo comments

SlicedSilver:
> > ci/circleci: install-deps-website
> 
> needs to be removed from the protected branches rules

Done

SlicedSilver (approved):

PR #2117, opened 2026-09-02
Create the private @tradingview/lwc-plugin-utils workspace package

illetid:


.github/workflows/deploy.yml
```diff
@@ -10,6 +10,7 @@ on:
       - 'website/**'
       - 'plugin-examples/src/**'
       - 'indicator-examples/src/**'
+      - 'packages/lwc-plugin-utils/src/**'
```
SlicedSilver: Probably not needed here but we can keep it to be safe.

SlicedSilver:
Closed because the changes are folded into the #2119 PR instead.

PR #2119, opened 2026-09-02
CLL-390: `create-lwc-plugin` for the plugin registry, plus `@tradingview/lwc-toolkit`

SlicedSilver:
**This supersedes the CLL-389 PR (#2117), which is closed in favour of this one.** CLL-389's
work is the first four commits here, rebased onto Andrii's latest `cll-389_workspace-package`
and attributed to him as co-author.

### `@tradingview/lwc-toolkit` (CLL-389)

The helpers that lived in `plugin-examples/src/helpers` — pixel-perfect dimension maths,
`PluginBase`, a `Delegate`, time conversion — become a publishable package. Previously every
plugin, and every scaffold, carried its own copy.

- **Renamed from `@tradingview/lwc-plugin-utils`.** The `lwc-plugin-` prefix is a validated
  contract: `create-lwc-plugin` rejects a plugin package name without it, and strips it to
  derive the entry filename. A package of shared helpers is not a plugin and should not sit
  inside that namespace.
- **No barrel; 13 sub-path `exports` entries**, one per module (`assertions`, `plugin-base`,
  `dimensions/positions`, …). Explicit entries rather than a `./*` wildcard, so the public
  surface stays a deliberate list.
- **Fixed the ESM emit.** `moduleResolution: "Node"` was producing extensionless relative
  specifiers inside a `"type": "module"` package, which real Node resolution rejects with
  `ERR_MODULE_NOT_FOUND`. It only worked because Vite and tsc tolerate it. Now `NodeNext`
  with explicit `.js` extensions, `target` pinned to `es2020` to match the library, and
  declaration and source maps emitted.
- Publishable metadata: `publishConfig.access`, keywords, `CHANGELOG.md`, `LICENSE`,
  `NOTICE`. Peer range widened to `^5.0.0` — nothing in it needs 5.2, and scaffolds derive
  their peer version from the live registry, so a narrower range only produced spurious
  warnings.

**Not published in this PR.** It merges unpublished; the only hard constraint is that the
toolkit is published before the next `create-lwc-plugin` release, since that release is what
starts telling community authors to install it.

### Scaffolding (CLL-390)

- Wizard gains a three-way category choice (`custom-series` / `series-primitive` /
  `pane-primitive`), plus description, author, licence, minimum-library-version and tag
  prompts, and a `--workspace` mode for official in-repo packages.
- ESM-only output: entry bundle, standalone bundle for CDN use, bundled `.d.ts`. No UMD, no
  generated `dist/package.json` — the authored `package.json` is the published manifest, and
  carries the `lwcPlugin` catalogue metadata.
- README template doubles as the registry detail-page body, with npm/CDN sections and
  community-only parts stripped in workspace mode.
- **Templates now depend on the toolkit instead of shipping copies of it.** Nine helper
  files deleted. Most were already dead weight: across all three starters the only helpers
  any template imported were `dimensions/positions`, `dimensions/common`, `assertions` and
  `plugin-base`, all from the series-primitive starter.
- **Every scaffold now gets a `.gitignore`.** Workspace mode had skipped it on the stated
  grounds that the repo already ignores package build output — it does not. The root
  `.gitignore` has `/dist/` with a leading slash (repository root only) and nothing for
  `typings/`, so a scaffolded package offered 11 build files to the next `git add`.

### Repo-wide

- **TypeScript 5.5.4 → 5.9.3**, which clears the `unbuild → typescript@^5.9.2` unmet-peer
  warning every install printed. Stayed on the 5.x line rather than the 7.x native port:
  `pnpm tsc` runs through `tspc`, which patches TypeScript's internals, and
  dts-bundle-generator and @typescript-eslint 7 would need coordinated upgrades too.
  `website` stays on 5.4.5for now because typedoc pins a narrow range.
- `lint:md` reduced to `markdownlint "**/*.md"`, with the exclusions moved into
  `.markdownlintignore`. Adds `*.local.md`, which is what made `pnpm lint:md` fail on any
  checkout holding a personal notes file, and `website/build/` for local Docusaurus builds.
- `create-lwc-plugin`'s own devDependencies exact-pinned, matching the rest of the repo.
- Two new CI jobs: `lwc-toolkit` (typecheck, build, per-sub-path Node ESM import, publint,
  arethetypeswrong) and `build-plugin-examples` (compiles every example plugin, including
  the bundled typings).

## Bugs fixed along the way

- **Scaffolded declarations were silently wrong.** With `followSymlinks: false`, the linked
  toolkit counts as an external library, and an external library in neither of
  dts-bundle-generator's library lists is emitted as *neither* a declaration nor an import.
  Confirmed by removing the fix again: the series-primitive declarations come out with
  `TS2304: Cannot find name 'PluginBase'` and a cascading `TS2420` that drops `chart` and
  `series` from the published class — so it degraded the public API, and `skipLibCheck` hid
  all of it from consumers. Fixed with `inlinedLibraries`.
- **The series-primitive starter leaked a subscription.** Its `plugin-base.ts` was an older
  copy whose `detached()` never called `unsubscribeDataChanged`, and whose `_fireDataUpdated`
  was a prototype method rather than a bound class property, so the unsubscribe could not
  have matched the subscribed reference in any case. Every scaffolded series-primitive
  inherited it. Depending on the toolkit fixes it.
- `compile.mjs` in `plugin-examples` now exits non-zero when typings generation fails,
  instead of logging and exiting 0.

## Verification

`107 files changed, 2104 insertions(+), 1003 deletions(-)`

- Every CI check green: `no-crlf`, `trailing-newlines`, `lint:eslint`, `lint:md`, `tspc`,
  `tsc-verify`, `type-tests`, unit tests — plus both new jobs, and the
  `indicator-examples` and `create-lwc-plugin` builds.
- **All seven published artefacts, `typings.d.ts` included, are byte-identical to `master`.**
  Built both branches and compared, so neither the TypeScript bump nor anything else here
  changes what ships.
- All 13 toolkit sub-paths import by specifier through real Node ESM. `publint --strict` and
  `arethetypeswrong` clean — the latter reports failures under `node10` and for a `require`
  from CJS, which is what an ESM-only package with an `exports` map is meant to do;
  `node16 (from ESM)` and `bundler` resolve every entry point.
- All three scaffold categories built in workspace mode, and one in community mode against
  a packed toolkit tarball: three artefacts each, no peer warnings, and **the generated
  `.d.ts` typechecks with `skipLibCheck` off** — the check that would have caught the
  declaration bug above.
- `plugin-examples` compiles 29 plugin bundles with no `lwc-toolkit` reference anywhere in
  `compiled/**`.

## Follow-ups, deliberately not here

- Publish `@tradingview/lwc-toolkit` before the next `create-lwc-plugin` release.
- `packages/*` and both examples workspaces are still excluded from ESLint, for a structural
  reason: the config's type-aware `parserOptions.project` points at the root `tsconfig.json`,
  which does not include them, so every file errors before a rule runs. Tracked separately
  alongside an oxlint/oxfmt evaluation.
- `tslint-eslint-rules` still lints through the deprecated `@typescript-eslint/tslint`
  bridge. `tslint-microsoft-contrib` was retired separately; the remaining rule (`ter-indent`)
  has no equivalent that agrees with the codebase, which the oxlint/oxfmt work addresses.


plugin-examples/src/helpers/time.ts
```diff
@@ -1,34 +0,0 @@
-import { Time, isUTCTimestamp, isBusinessDay } from 'lightweight-charts';
-
-export function convertTime(t: Time): number {
-	if (isUTCTimestamp(t)) return t * 1000;
-	if (isBusinessDay(t)) return new Date(t.year, t.month, t.day).valueOf();
```
illetid, line 5: I think we had error in this code. 
it should be `t.month -1`

packages/create-lwc-plugin/template-common/package.json
```diff
@@ -1,17 +1,59 @@
...
+	"author": "_AUTHOR_",
+	"license": "_LICENSE_",
+	"type": "module",
+	"types": "./dist/_ENTRYNAME_.d.ts",
+	"exports": {
+		".": {
+			"types": "./dist/_ENTRYNAME_.d.ts",
+			"default": "./dist/_ENTRYNAME_.js"
+		},
+		"./standalone": "./dist/_ENTRYNAME_.standalone.js"
```
illetid: do we need to add types here too? 

```
"./standalone": {
	"types": "./dist/_ENTRYNAME_.d.ts",
	"default": "./dist/_ENTRYNAME_.standalone.js"

}

```

illetid (approved):

PR #2121, opened 2026-09-04
Add the plugin CI checks: PR gates in CircleCI, dry-run, floor and sm…

illetid:


.github/workflows/plugins.yml
```diff
@@ -0,0 +1,158 @@
...
+        run: pnpm install --frozen-lockfile
+
+      # The published manifest is the cleaned one: the root package.json's
+      # scripts (a postinstall among them) and devDependencies must not end up
+      # in the tarball the consumer project installs. Same step as pkg-pr-new.yml.
+      - name: Pack the workspace library
+        if: env.LIBRARY_SOURCE == 'workspace'
+        run: |
+          pnpm build:prod
+          pnpm prepare-package-json-for-release
```
SlicedSilver, line 146: **`P1`**: Preserve the root manifest before the smoke test

`pnpm prepare-package-json-for-release` runs `clean-package-json.js`, which permanently removes the root `scripts` field. The following step therefore cannot run `pnpm plugins:smoke`; release-branch, tag, and manually dispatched workspace smoke runs will fail with a missing-script error. Even invoking the script directly would later fail when `ensureWorkspaceBuilt()` calls `pnpm build`. Pack from a staging copy or restore the original `package.json` immediately after packing.

SlicedSilver (changes requested):

.github/workflows/plugins.yml
illetid: fixed in https://github.com/tradingview/lightweight-charts/pull/2121/commits/ba4588fb461f673625a5b695c872403250a94a5f

SlicedSilver (approved):

PR #2124, opened 2026-09-07
cll-404 catalogue data

illetid:


SlicedSilver (commented):
**1. `validateAll` drops `repoRoot` — the `repository.directory` check is silently disabled for any `--repo-root` run** ([catalogue-data.mjs:145](scripts/plugins/catalogue-data.mjs:145))

`release.mjs` and `validate-metadata.mjs` were both updated to pass `repoRoot`; this call site wasn't. Because it falls back to `DEFAULT_REPO_ROOT` (the real repo), a package under a different root computes a `..`-prefixed relative path, `insideRepo` is false, and the check is skipped. Confirmed: I set `repository.directory` to `packages/totally-wrong` in a fixture and `catalogue.mjs --repo-root <fixture>` accepted it.

It's latent — the docs build never passes `--repo-root` — but it's a real inconsistency, and it means the catalogue tests can never exercise that validation. Fix is one line:

```js
validatePackageMetadata(plugin.dir, { isOfficial: true, repoRoot });
```

with `validateAll(plugins, repoRoot)`.

**2. `err.dataPath` fallback is dead code** ([utils.mjs:375](scripts/plugins/utils.mjs:375))

`dataPath` is Ajv 6; this repo pins `ajv@7.2.4`, which only sets `instancePath`. The `|| err.dataPath` never fires. Harmless, but it implies an Ajv-version ambiguity that doesn't exist — I'd drop it. (The real fix in that line — `|| '/'` for root-level errors — is right, and the new `/category` assertion covers it.)

**3. The retry tests sleep ~9 real seconds**

`RETRY_DELAY_MS` isn't injectable, so `fetchPackument`'s two retry tests burn 3s + 6s of wall clock — 95% of the 9.9s suite runtime. Threading a `delayMs` option through `fetchWithRetry` (defaulted, passed as ~0 in tests) would make it instant. Worth doing before the suite grows.

**4. The tarball is never integrity-checked** ([catalogue-data.mjs:81](scripts/plugins/catalogue-data.mjs:81))

The README is fetched from `dist.tarball` and rendered as Markdown into the public docs site, but `dist.integrity`/`dist.shasum` from the packument is ignored. Extraction itself is safe (strict `^package/readme(\.[^/]+)?$` regex, `-xzO` to stdout, nothing written). Still, verifying the hash you already have in hand is cheap defence-in-depth for content that ends up on the site.

**5. Minor notes** (take or leave)

- `assertUniqueSlugs` only covers published entries; an unpublished package colliding with a published one won't surface until its first publish breaks the docs build. A cheap fix is to check slugs across `plugins.map(p => p.name)` before splitting.
- `Promise.all` over all plugins downloads every tarball concurrently. Fine at wave-1 scale; worth a concurrency cap if the list gets long.
- `description` falling back to the workspace warns; `license` and `keywords` fall back silently. Probably deliberate, but slightly inconsistent.
- `fetchTarballReadme` shells out to `tar`. Node ≥22.3 is required anyway and Windows 10+ ships `tar.exe`, so this is fine — just noting it's now a build-time dependency of the website.

Nothing here is a blocker. #1 is the only one I'd call a genuine defect.

SlicedSilver (approved):

PR #2135, opened 2026-09-10
Update readmes

anachuprina:
**Overview of change:**
Cross-links between the docs, the repo READMEs, and the upcoming plugin catalog (/plugins), plus a general cleanup pass on the READMEs touched along the way.

The `/plugins` links use plain paths, so CI is expected to be red (onBrokenLinks + broken-link checks) until CLL-405 (catalog pages) merges. Do not merge this before CLL-405, ideally after cll-394-403 (plugin graduation) too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,14 +21,12 @@
 
 TradingView Lightweight Charts™ are one of the smallest and fastest financial HTML5 charts.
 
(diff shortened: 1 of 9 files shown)


SlicedSilver:
Look okay but we can't merge until we merge the other changes (because the site has broken links). We can't merge the other part without it publishing it automatically...

So I will change the target of this PR to be the other branch and we can merge into that.

